PR #3, opened 2021-02-21
Added TCP Port setting for SSH. 

richardschoen:


richardschoen:
Also change temp file name from /tmp/randomvalue to /tmp/vscodetemp-randomvalue so they can be cleaned up easily. Perhaps home directory makes sense at some point but at least they can clean their temp dirs easier with this naming.

src/api/IBMi.js
```diff
@@ -17,7 +18,7 @@ module.exports = class IBMi {
   }
 
   /**
-   * @param {{host: string, username: string, password: string, keepaliveInterval?: number}} connectionObject 
+   * @param {{host: string, port: string, username: string, password: string, keepaliveInterval?: number}} connectionObject
```
worksofliam, line 21: Why is port a string? Shouldn't it b a number?

worksofliam (approved):
LGTM. Need to do some testing but then I'll give it a whirl.

package.json
```diff
@@ -25,84 +27,100 @@
...
-									}
+						"type": "object",
+						"title": "Connection details",
+						"properties": {
+							"host": {
+								"type": "string",
+								"description": "Host name or IP address"
+							},
+							"port": {
+								"type": "string",
```
worksofliam, line 71: This should be a number.

worksofliam (changes requested):
After the user is done in the webview, the `port` needs to be converted to a number.

Please make those changes and I can review again.

src/api/IBMi.js
richardschoen: I guess, but it works. Must get auto-converted.

worksofliam: It may work, but I am trying my best to write the easier code to understand + eslint is moaning at me! 
image

richardschoen:
Updates pushed

worksofliam:
Accidently put it onto another branch.

PR #15, opened 2021-02-23
fix Instance class not found on linux

barrettotte:
This was quite a weird one because I didn't encounter any issues on Windows at all. Perhaps it's Windows or different versions of Node that makes file naming a little more relaxed, I'm still a bit confused. 

I encountered the same error as https://github.com/halcyon-tech/code-for-ibmi/issues/14 on Ubuntu 20.04.2 and VS Code 1.53.2.

```
Activating extension 'halcyontechltd.code-for-ibmi' failed: Cannot find module './src/instance'
Require stack:
- /home/barrett/Programming/code-for-ibmi/extension.js
- /usr/share/code/resources/app/out/vs/loader.js
- /usr/share/code/resources/app/out/bootstrap-amd.js
- /usr/share/code/resources/app/out/bootstrap-fork.js.
```

I debugged it locally and found that multiple files had a problem resolving the paths to 'Instance.js'
So, I changed the import paths from 'instance' to 'Instance'.
I also added the explicit class name Instance to 'Instance.js' although I don't think its completely necessary.

This appears to have fixed the problem on Ubuntu and I am able to get the connection prompt to appear.
However, I can only get as far as attempting the connection.
I guess I locked myself out of my PUB400 account so I cannot check further than this on Linux.

I also checked on Windows and this fix does not seem to break anything and works as it did before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@ Browse, edit and compile your source members right from Visual Studio Code.
 * [@worksofliam](https://github.com/worksofliam)
 * [@alanseiden](https://github.com/alanseiden)
 * [@richardschoen](https://github.com/richardschoen)
+* [@barrettotte](https://github.com/barrettotte)
 
 ## Guide
 
--- a/extension.js
+++ b/extension.js
@@ -5,7 +5,7 @@ const vscode = require('vscode');
 // this method is called when your extension is activated
 // your extension is activated the very first time the command is executed
 
-var instance = require('./src/instance');
+var instance = require('./src/Instance');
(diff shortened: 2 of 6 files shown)


worksofliam:
I think this will fix #15! Thank you @barrettotte. You are awesome.

Would you also add yourself to the contributors' list on the readme?

worksofliam:
Going to do some quick tests and will merge very soon.

worksofliam:
LGTM. Thank you @barrettotte!

PR #48, opened 2021-02-26
targeted member list refresh, fixes #31

### LINKED ISSUE #31: Source member list refresh at top refreshes all current source files. Could be slow.
Is there a way to only refresh the selected source file in case a developer has several source files in their list ?

Ex: Click on source file and then click the refresh icon at the top. 

If this change is made title it: "Refresh Selected Source File" instead of "Refresh Member List"

Another option would be to just re-add the Right-Click option at the source file level to just refresh the selected source file.

**Either way the idea is to NOT have to refresh all members across all my source files in the list all the time.**

barrettotte:
### Changes

Changes to address #31 

Add ability for targeted member list refresh via context menu on source file.
Also, update create, delete, copy, updateText, rename functionality on members to use targeted refresh.

### Checklist

* [x] have tested my change

**Note:** I think I still kind of suck at git branch workflows. So I was trying to get a workflow of
```upstream/master -> origin/master -> origin/feature/fixes -> upstream/feature/fixes``` going. 

Should I be doing something different in the future? I figured I shouldn't do a PR straight to master.

I can close this PR and resituate my commits if need be (with a little guidance)


worksofliam:
This is an awesome feature fix. The issue is that, yes, you branched off an old branch. 

It might be a bad idea for me to merge a branch that has 63 new commits in it. For some reason, git has includes those thinking they're all new commits.

In the future, you should only PR off of master and not another branch (unless I have explicitly mentioned it somewhere).

Would it be possible, whenever you can, to resubmit your PR based off of master?

barrettotte:
@worksofliam Yeah definitely a good idea. No problem, that works for me

Issue #26, opened 2021-02-24
Object browser similar to member browser


richardschoen:
This was Liams

worksofliam:
What would the object browser be used for? That's the real question

richardschoen:
Not sure. I was just writing down your idea :-)

I suppose just like WRKOBJPDM or WRKOBJ or Object list in RDI to access objects and attributes.

PR #56, opened 2021-02-28
Start of object browser (#26)

worksofliam:
### Changes

Covers issue #26. This adds an object browser to the IBM i view. I am using the same caching technique used in the member browser by @barrettotte.

Right-click contexts include:

* Run Action

### Checklist

* [X] have tested my change

### DIFF
--- a/src/views/objectBrowser.js
+++ b/src/views/objectBrowser.js
@@ -124,7 +124,7 @@ class Object extends vscode.TreeItem {
    */
   constructor({library, name, type, text}) {
     if (type.startsWith('*')) type = type.substring(1);
-    
+
     const icon = objectIcons[type] || objectIcons[''];
 
     super(`${name.toLowerCase()}.${type.toLowerCase()}`);
@@ -133,7 +133,8 @@ class Object extends vscode.TreeItem {
     this.path = `${library}/${name}`;
     this.type = type;
     this.description = text;
-    this.iconPath = new vscode.ThemeIcon(icon)
+    this.iconPath = new vscode.ThemeIcon(icon);
+    this.resourceUri = vscode.Uri.parse(`${library}/${name}.${type}`).with({scheme: 'object'})
   }
 }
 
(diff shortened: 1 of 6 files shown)


worksofliam:
@barrettotte I have given you 'read' access on a repo that says you can now review PRs. Any chance, whenever you have the time AND if you wanted, you could review this PR for me?

If not, let me know. 

Thanks :)

worksofliam:
Before this is merged, should I add more right-click options?

* Delete
* Change text?

src/views/objectBrowser.js
```diff
@@ -0,0 +1,148 @@
...
+    this.resourceUri = vscode.Uri.parse(`${library}/${name}.${type}`).with({scheme: 'object'})
+  }
+}
+
+//https://code.visualstudio.com/api/references/icons-in-labels
+const objectIcons = {
+  'FILE': 'database',
+  'CMD': 'terminal',
+  'MODULE': 'extensions',
+  'PGM': 'chevron-right',
```
barrettotte, line 146: Is there eventually going to be nested info in objects of type *PGM? If not, maybe setting the icon to something like ```'PGM': 'file-binary'``` would make sense. I really like the other icons though, it feels so natural

worksofliam: `file-binary` is much nicer. I will change that.

barrettotte (commented):
Code looks awesome and follows the same pattern as the other browsers. I also like the use of the targeted refresh by library. The icons feel very natural minus *PGM, let me know what you think of the comment I left on objectBrowser.js

One thing that I thought of while testing a couple small CL compiles:  Is there any easy way to refresh the object browser after running a compile? That way the newly created PGM would show up in the object list. If not, its not a big deal at all to just refresh

worksofliam:
@barrettotte 

I have fixed an issue where the right-click refresh was not showing up. I also changed the icon to `file-binary`. As well as those, if you have auto-refresh enabled, it will invoke`refreshObjectList` after an action has finished executing.

Please review again :)

barrettotte (approved):
Just tried compiling again and the appropriate library refreshed on compile, thanks!

worksofliam:
Will be merging today! Thanks @barrettotte :)

PR #82, opened 2021-03-06
Add Checkbox to CustomUI

priceaj:
### Changes

Added ability to add a checkbox to CustomUI webview panels

### Checklist

* [X] have tested my change

### DIFF
--- a/src/api/CustomUI.js
+++ b/src/api/CustomUI.js
@@ -52,7 +52,7 @@ class CustomUI {
     }
 
     const onDiskPath = vscode.Uri.file(
-      path.join(context.extensionPath, `src`, `webviews`, `extern`, `vscwe.js`)
+      path.join(context.extensionPath, `node_modules`, `@bendera`, `vscode-webview-elements`, `dist`, `bundled.js`)
     );
     // And get the special URI to use with the webview
     const onDiskSrc = panel.webview.asWebviewUri(onDiskPath);
@@ -94,7 +94,7 @@ class CustomUI {
     
                 var data = {};
     
-                for (const field of fields) data[field] = document.getElementById(field).value;
+                for (const field of fields) data[field] = document.getElementById(field).checked || document.getElementById(field).value;
     
                 vscode.postMessage(data)
             };
@@ -120,7 +120,7 @@ class CustomUI {
 
 class Field  {
   constructor(type, id, label) {
-    /** @type {"input"|"password"|"submit"} */
+    /** @type {"input"|"password"|"submit"|"checkbox"}} */
     this.type = type;
 
     /** @type {string} */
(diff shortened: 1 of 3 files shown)


worksofliam:
@priceaj Just going to do some testing!

src/api/CustomUI.js
```diff
@@ -140,7 +140,13 @@ class Field  {
     switch (this.type) {
     case `submit`:
       return `<vscode-button id="${this.id}">${this.label}</vscode-button>`;
-
+    case `checkbox`:
+      return `
+        <vscode-form-item>
+        <vscode-form-control>
+        <vscode-checkbox id="${this.id}" value="${this.default}">${this.label}</vscode-checkbox>
```
worksofliam, line 147: I think this line should be more like below, as setting the value doesn't let you set it as checked or not.

```
<vscode-checkbox id="${this.id}" ${this.default === `checked` ? `checked` : ``}>${this.label}</vscode-checkbox>
```

src/api/CustomUI.js
```diff
@@ -140,7 +140,13 @@ class Field  {
     switch (this.type) {
     case `submit`:
       return `<vscode-button id="${this.id}">${this.label}</vscode-button>`;
-
+    case `checkbox`:
+      return `
+        <vscode-form-item>
```
worksofliam, line 145: Below this line, I think we should also include the description like in the other components below this line.

```
${this.description ? `<vscode-form-description>${this.description}</vscode-form-description>` : ``}
```

worksofliam (changes requested):
@priceaj Please see my two comments and let me know what you think.

We really need a way to set if the default value is checked or not.

src/api/CustomUI.js
```diff
@@ -94,7 +94,7 @@ class CustomUI {
     
                 var data = {};
     
-                for (const field of fields) data[field] = document.getElementById(field).value;
+                for (const field of fields) data[field] = document.getElementById(field).checked || document.getElementById(field).value;
```
worksofliam, line 97: There is a bit of a bug in this expression.

`document.getElementById(field).checked` always returns `true`, `false` or `undefined`. If it's not true, then it's returning the `.value` prop, which if it's a checkbox.. it's always `checked`.

We need to turn this into a switch statement that checks what type of input it is, and then assign `.checked` (for checkbox) or `.value` (for text) accordingly.

priceaj: When I was testing I didn't get this behaviour, it either returned True (i.e. because it was checked) or Undefined (i.e. because it was not checked and the value attribute of the checkbox was set to be this.default which is undefined.

I figured that was good enough, because the checked attribute only appears on checkbox (and radio) elements

worksofliam: Let me have another go and I will share my results with you here.

priceaj: Having said that I hadn't actually worked out what the default field was being used for, because it couldn't be passed into a function which creates the element, so I was using it more of a placeholder, if your intention was to use it to set a default value on a field e.g. as you have done above, then yes I can see how you could get this sort of issue. And I'd agree we would have to check the field type, and pick the relevant field

worksofliam: Okay, I see it is coming back as `undefined` now - but I personally think we should make it as clean as possible and have it return `true`/`false`. What do you think?

image

src/api/CustomUI.js
priceaj: applied

src/api/CustomUI.js
priceaj: applied

worksofliam (approved):
Looking good and working good from my testing!

worksofliam:
Thank you @priceaj! This is one awesome change. I will merge now.

PR #94, opened 2021-03-10
Add File Picker for SSH Key

priceaj:
### Changes

I've expanded on the work done by @onewheelonly and added a file picker for the SSH Key.

### Checklist

* [X] have tested my change - with caveats (I have not tested the actual connection works as I don't have SSH Keys set up on PUB400.com)

### DIFF
--- a/src/api/CustomUI.js
+++ b/src/api/CustomUI.js
@@ -81,7 +81,8 @@ class CustomUI {
         (function () {
             const vscode = acquireVsCodeApi();
             const submitButton = document.getElementById('${submitButton.id}');
-            const fields = [${this.fields.filter(field => field.type !== `submit`).map(field => `'${field.id}'`).join(`,`)}];
+            const submitfields = [${this.fields.filter(field => field.type !== `submit`).map(field => `'${field.id}'`).join(`,`)}];
+            const filefields = [${this.fields.filter(field => field.type == `file`).map(field => `'${field.id}'`).join(`,`)}];
     
             const doDone = (event) => {
                 if (event)
@@ -89,7 +90,7 @@ class CustomUI {
     
                 var data = {};
     
-                for (const field of fields) {
+                for (const field of submitfields) {
                   var fieldType = document.getElementById(field).nodeName.toLowerCase();
                    switch (fieldType) {
                     case "vscode-checkbox"
@@ -105,7 +106,7 @@ class CustomUI {
             submitButton.onclick = doDone;
             submitButton.onKeyDown = doDone;
 
-            for (const field of fields) {
+            for (const field of submitfields) {
                 document.getElementById(field)
                     .addEventListener('keyup', function(event) {
                       event.preventDefault();
@@ -114,6 +115,23 @@ class CustomUI {
                     }
                 });
             }
+
+            for (const field of filefields) {
+              document.getElementById(field)
+                  .addEventListener('vsc-change', (e) => {
+                      const VirtualField = document.getElementById(e.target.id)
+                      let input = VirtualField.shadowRoot.querySelector("input");
+                      for (let file of Array.from(input.files)) {
+                          let reader = new FileReader();
+                          reader.addEventListener("load", () => {
+                            console.log(file.path);
+                            document.getElementById(e.target.id).setAttribute("value", file.path)   
+                          });
+                          reader.readAsText(file);
+                      }
+                  }
+                  )}
+
         }())
     </script>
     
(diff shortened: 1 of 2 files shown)


onewheelonly:
Yup, I can give this a try on Windows (at some point). Thanks for this, Aaron!

worksofliam:
This is awesome. I will give the mac change a go.

worksofliam:
Looking good on a mac!

@onewheelonly let me know how it is for Windows and we will merge.

src/api/CustomUI.js
```diff
@@ -114,6 +115,23 @@ class CustomUI {
...
+                  .addEventListener('vsc-change', (e) => {
+                      const VirtualField = document.getElementById(e.target.id)
+                      let input = VirtualField.shadowRoot.querySelector("input");
+                      for (let file of Array.from(input.files)) {
+                          let reader = new FileReader();
+                          reader.addEventListener("load", () => {
+                            console.log(file.path);
+                            document.getElementById(e.target.id).setAttribute("value", file.path)   
+                          });
+                          reader.readAsText(file);
```
priceaj, line 130: I realise I am actually reading the file here, and I could have done this more efficiently, but I wanted to leave scope to enable me to read the key into the VSCode secret storage and use it from there

worksofliam: @priceaj You're reading the file but not assigning it anywhere. Are you planning to remove this piece of code? Or something else with it?

worksofliam: Wait. I now see it's triggering an event when it is read - except it's still passing the path back in the data.

src/api/CustomUI.js
```diff
@@ -114,6 +115,23 @@ class CustomUI {
                     }
                 });
             }
+
+            for (const field of filefields) {
+              document.getElementById(field)
+                  .addEventListener('vsc-change', (e) => {
+                      const VirtualField = document.getElementById(e.target.id)
+                      let input = VirtualField.shadowRoot.querySelector("input");
```
priceaj, line 123: this isn't great, but the framework doesn't expose the "files" attribute of the underlying input box

worksofliam:
@priceaj 

```js
                          let reader = new FileReader();
                          reader.addEventListener("load", () => {
                            console.log(file.path);
                            document.getElementById(e.target.id).setAttribute("value", file.path)   
                          });
                          reader.readAsText(file);
```

What's the point of reading the file? Perhaps you should just pass the selected path back (like you are already doing)

src/api/CustomUI.js
priceaj: yeah, It's there so that the feature can be expanded to read the content of the Private Key file, and store it in the VSCode secret storage (along with any passwords)... and I've just spotted a console.log I didn't remove!

worksofliam: @priceaj I will wait for another push from you before I do any more :)

src/api/CustomUI.js
```diff
@@ -114,6 +115,23 @@ class CustomUI {
...
+                      for (let file of Array.from(input.files)) {
+                          let reader = new FileReader();
+                          reader.addEventListener("load", () => {
+                            console.log(file.path);
```
worksofliam, line 127: Naughty

worksofliam (changes requested):
* Are you planning on removing the `readAsText` and just passing back the path?
* Please remove your console.log

src/api/CustomUI.js
priceaj: I have removed the console.log statement. 

I wasn't intending to remove the reading of the file, mainly because:
1. I've left it there for a future (hopefully imminent) change to read the key and pass through to node-ssh directly as a raw string (as per the commented example https://github.com/steelbrain/node-ssh#example)
2. The change gives me scope to read the key and store the ssh-key in the VSCode secret storage which should be safer than leaving an unprotected private key file sat on the file system with a plain text pointer to it in the VSCode settings
3. It leaves an obvious jumping off point if anyone else wants to add another file picker into the UI and return the content rather than the path 
4. The reason I am reading the path into the value attribute of the virtual DOM field is that by default its gets a value of C:\fakepath\filename.ppk (for security) and I am overriding that security to align with current functionality, but adding the convenience of being able to choose a file rather than having to paste/type a file path. I'd like to read the content of the file so that this workaround isn't necessary. 
5. As far as I can tell it doesn't create any bugs or overhead, even if the file is deleted after it's selected

worksofliam: Thanks for all this detailed info!

worksofliam:
Everything is in order. I just would like @onewheelonly to confirm it works for them as well and we shall go forward with the merge.

onewheelonly:
Everything looks fine on the Windows side 👍

worksofliam (approved):

PR #147, opened 2021-05-29
Member and IFS search functionality

worksofliam:
### Changes

This PR implements right-click menus on source members and IFS directories that allow you to search through the content of those files/directories.

### Things to consider

* [ ] Does `FNDSTRPDM` return different output for different languages? Obviously.. I coded towards English.

### Checklist

* [X] have tested my change

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
 		"onCommand:code-for-ibmi.createStreamfile",
 		"onCommand:code-for-ibmi.moveIFS",
 		"onCommand:code-for-ibmi.deleteIFS",
+		"onCommand:code-for-ibmi.searchIFS",
 		"onView:objectBrowser",
 		"onCommand:code-for-ibmi.refreshObjectList",
 		"onCommand:code-for-ibmi.createSourceFile",
@@ -631,6 +632,11 @@
 				"icon": "$(files)"
 			},
 			{
+				"command": "code-for-ibmi.searchIFS",
+				"title": "Search Directory",
+				"category": "IBM i"
+			},
+			{
 				"command": "code-for-ibmi.createDirectory",
 				"title": "Create directory",
 				"category": "IBM i",
@@ -904,6 +910,11 @@
 					"command": "code-for-ibmi.createStreamfile",
 					"when": "view == ifsBrowser && viewItem == directory",
 					"group": "2_ifsStuff@1"
+				},
+				{
+					"command": "code-for-ibmi.searchIFS",
+					"when": "view == ifsBrowser && viewItem == directory",
+					"group": "3_ifsStuff@1"
 				},
 				{
 					"command": "code-for-ibmi.refreshObjectList",
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
 		"onCommand:code-for-ibmi.deleteMember",
 		"onCommand:code-for-ibmi.updateMemberText",
 		"onCommand:code-for-ibmi.renameMember",
+		"onCommand:code-for-ibmi.searchSourceFile",
 		"onView:ifsBrowser",
 		"onCommand:code-for-ibmi.refreshIFSBrowser",
 		"onCommand:code-for-ibmi.changeHomeDirectory",
@@ -555,6 +556,11 @@
 				"icon": "$(refresh)"
 			},
 			{
+				"command": "code-for-ibmi.searchSourceFile",
+				"title": "Search Source File",
+				"category": "IBM i"
+			},
+			{
 				"command": "code-for-ibmi.addSourceFile",
 				"title": "Add Source File",
 				"category": "IBM i",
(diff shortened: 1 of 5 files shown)


worksofliam:
image
image

worksofliam:
Looks like there is an issue with the implementation since different locales print different text to the spool files. Right now this PR only works for English. Not good.

> It does. For 2928 they are 'Membre', 'Type' and 'Texte'. You should be able to see it by using a secondary language too.

https://twitter.com/dferrand/status/1398705688650657798

dferrand:
Why don't you use grep?

worksofliam:
@dferrand I completely forgot about grep. We are using grep from yum to search IFS directories... but I forgot about grep in QSH! 

That should allow us to use it against source members:

Note, that because members in QSYS do not have extensions, we will have to fetch that list before doing the grep so we can give them the correct extension in the result.

https://www.itjungle.com/2014/04/02/fhg040214-story02/

I will work on this today.

worksofliam:
Okay, so now the member search uses grep which is awesome. We had to do a call to get a list of members first because we need the extension when listing the results.

Excited for this feature to be merged in!

priceaj:
Code looks good and grep search seems to work well (on PUB400 ymmv), I have some comments about the functionality:

* IFS cant search current directory, only subdirectories, would having it in the navigation group rather than a right click menu on a folder (or both) be better?
* Search is missing from the object browser, and can only be initiated from the member browser, could/should this be added to source files?
* Result file is an editable TextDocument, would an OutputChannel (or webview) be better?
* not sure if naming of "Search directory" implies a filename search (which it does not do) rather than a content search

also docs need updating as I couldn't work out where to go to search without looking at the code!

worksofliam:
@priceaj awesome review. Thanks! See my comments

> IFS cant search current directory, only subdirectories, would having it in the navigation group rather than a right click menu on a folder (or both) be better?

I add it to both places. Now you can right-click on a directory, or there is an option to search the working directory:

image

You can see that in the latest commits.

> Search is missing from the object browser, and can only be initiated from the member browser, could/should this be added to source files?

Yes. I didn't plan on adding it here. Do you think it's worth it?

> Result file is an editable TextDocument, would an OutputChannel (or webview) be better?

I don't think an OutputChannel would be a good idea. I am not a fan of an editable text document, but the textDocumentProvider is a ton of work. A webview might be nice but I have no idea what it should look like. I was mostly trying to match the style of the vscode search result file.

>not sure if naming of "Search directory" implies a filename search (which it does not do) rather than a content search

If you know a better label to use I'd love to hear it!

priceaj:
Looks good, definitely being able to search from the current directory is better! 

The other points were just minor things, maybe worth shipping and seeing if people using it run into issues/have any suggestions!



Just noticed one more thing, in search.js you are using a hardcoded path for grep (/usr/bin/grep) but the RemoteFeatures path is set to 

grep:'/QOpenSys/pkgs/bin/grep'

On PUB400 I can see 3 locations for grep:

```
$ whereis grep
grep: /usr/bin/grep /QOpenSys/pkgs/bin/grep /QOpenSys/usr/bin/grep (I believe the locations are for QSHELL, YUM and PASE respectively)
```

the default one in my path is:

```
$ which grep
/QOpenSys/pkgs/bin/grep
```

did you hardcode the executable path deliberately e.g. for EBCDIC > ASCII/QSHELL compatibility reasons or should this use the one in RemoteFeatures?


Conversely, is it worth using something like `which` (included with PASE) to pick up the location of commands before directly checking in /QOpenSys/pkgs/bin for enabling of the RemoteFeatures? (I think which will only check in the PATH, so still worth checking directly in the yum packages directory)

In this case I don't believe the IFS search feature will work if you don't have the YUM packages installed, yet grep is included with PASE and all 3 executables seem to work to search IFS:

$ whereis grep
grep: /usr/bin/grep /QOpenSys/pkgs/bin/grep /QOpenSys/usr/bin/grep
$ /usr/bin/grep oomo test.txt
oomo
$ /QOpenSys/pkgs/bin/grep oomo test.txt
oomo
$ /QOpenSys/usr/bin/grep oomo test.txt
oomo

worksofliam:
@priceaj Some great points.

> did you hardcode the executable path deliberately e.g. for EBCDIC > ASCII/QSHELL compatibility reasons or should this use the one in RemoteFeatures?

This is exactly right.

> Conversely, is it worth using something like which (included with PASE) to pick up the location of commands before directly checking in /QOpenSys/pkgs/bin for enabling of the RemoteFeatures? (I think which will only check in the PATH, so still worth checking directly in the yum packages directory)

> In this case I don't believe the IFS search feature will work if you don't have the YUM packages installed, yet grep is included with PASE and all 3 executables seem to work to search IFS:

The reason we only check `/QOpenSys/pkgs/bin` is so that we can be sure that _that_ directory has the latest tools in it. While it would be nice for it to work with just the pase version, I always want our users to be on the latest stuff.

Thanks for all your feedback!

Issue #146, opened 2021-05-29
Improved implementation of the library list


szsascha:
Today I've tried code for ibmi the first time. Thank you for creating this great project!

At the beginning I was a bit confused by the implementation of the library list. I think with a few improvements the implementation of the libraries can be made a bit more intuitive and easier.

- Loading of the entire library list after login

Currently only the user part of the library list is loaded after login. In my opinion it would be better if the curlib and maybe the system part are also loaded at the beginning. In my current case (PUB400 default setup) all objects and sources are in my curlib.

- Use the curlib as the default build library

- Consolidate member browser and object browser to a library browser (libraries from library list should be added by default)

- Initialize files of member browser automatically by library

worksofliam:
@szsascha 

These are some awesome suggestions. I am working on the search functionality today, but I will look into some of this for you.

> Loading of the entire library list after login

> Consolidate member browser and object browser to a library browser (libraries from library list should be added by default)

We actually had this way back when we started to develop this extension. The issue here is when there are thousands of objects and also thousands of source members. It wasn't pretty, hard to navigate, and even had some performance issues at times. It will have to stay like it is for now. You are able to right-click on source files from the object browser and add them to the member browser, though.

> Use the curlib as the default build library

This one is easy to do! Not sure why we didn't do this before.

szsascha:
@worksofliam 

Thanks for your quick reply!

> We actually had this way back when we started to develop this extension. The issue here is when there are thousands of objects and also thousands of source members. It wasn't pretty, hard to navigate, and even had some performance issues at times. It will have to stay like it is for now. You are able to right-click on source files from the object browser and add them to the member browser, though.

Are there any other possibilities? Maybe you could install some kind of agent on the IBM i. I don't know if that solves this problem, but it might open up new possibilities elsewhere. Just an idea that came to me randomly.

> Are you suggesting we search through all objects your library list and add all source files to the member browser?

Yes. exactly. But I guess this will also lead to performance issues?

PR #148, opened 2021-05-29
Curlib optimizations

szsascha:
#146 

### Checklist

* [x] have tested my change

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ node_modules
 .vscode-test/
 *.vsix
 dist
+package-lock.json
+.DS_Store
--- a/src/api/IBMi.js
+++ b/src/api/IBMi.js
@@ -70,6 +70,7 @@ module.exports = class IBMi {
             break;
               
           case `CUR`:
+            this.defaultUserLibraries.push(lib);
             currentLibrary = lib;
             break;
           }
--- a/src/api/IBMi.js
+++ b/src/api/IBMi.js
@@ -90,7 +90,7 @@ module.exports = class IBMi {
           undefined,
         );
       } catch (e) {
-        let [errorcode] = e.split(`:`)[0];
+        let [errorcode, errortext] = e.split(`:`);
 
         switch (errorcode) {
         case `CPF2111`: //Already exists, hopefully ok :)
--- a/src/api/IBMi.js
+++ b/src/api/IBMi.js
@@ -80,6 +80,7 @@ module.exports = class IBMi {
         if (this.config.libraryList.length === 0) await this.config.set(`libraryList`, this.defaultUserLibraries);
         if (this.config.objectBrowserList.length === 0) await this.config.set(`objectBrowserList`, this.defaultUserLibraries);
         if (this.config.databaseBrowserList.length === 0) await this.config.set(`databaseBrowserList`, this.defaultUserLibraries);
+        if (!this.config.buildLibrary || this.config.buildLibrary.length === 0 || this.config.buildLibrary == `QTEMP`) await this.config.set(`buildLibrary`, currentLibrary);
       }
 
       //Next, we need to check the temp lib (where temp outfile data lives) exists
--- a/src/api/IBMi.js
+++ b/src/api/IBMi.js
@@ -90,7 +90,7 @@ module.exports = class IBMi {
           undefined,
         );
       } catch (e) {
-        let [errorcode, errortext] = e.split(`:`);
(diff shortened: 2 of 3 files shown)


worksofliam:
@szsascha Thanks for this PR.

I updated the original comment to include the required checklist. After those items are complete and I've had a little test, it can be merged.

Thanks!

.gitignore
```diff
@@ -1,4 +1,6 @@
 node_modules
 .vscode-test/
 *.vsix
-dist
\ No newline at end of file
+dist
+package-lock.json
```
worksofliam, line 5: `package-lock.json` must not be in the .gitignore.

src/api/IBMi.js
```diff
@@ -90,7 +90,7 @@ module.exports = class IBMi {
           undefined,
         );
       } catch (e) {
-        let [errorcode, errortext] = e.split(`:`);
+        let [errorcode] = e.split(`:`)[0];
```
worksofliam, line 93: What was the reason for this change? I suppose ESLint was just saying it was undefined?

For the moment, please revert this change as I am about to start using the `errorText` in another PR I'm in.

szsascha: I just didn't want any open warnings :)

But I've reverted the commit.

worksofliam:
@szsascha I am going to test this now.

worksofliam:
Okay, so I remember why we didn't add `CUR` or `SYS` libraries to the `defaultUserLibraries` prop - that's because we can't remove them from the library list at compile time.

We do this at compile time:

```js
              commandResult = await connection.qshCommand([
                `liblist -d ` + connection.defaultUserLibraries.join(` `),
                `liblist -a ` + libl.join(` `),
                command,
              ], undefined, 1);
```

We remove the default library list (since each compile runs in its own job) and then set it to the user's library list in the extension.

Maybe, instead of this change, we actually need to display the `CUR` and `SYSTEM` libraries in the library list view, but not let the user move/remove them.

What do you think @szsascha?

worksofliam:
Next big crazy idea:

We completely replace `&BUILDLIB` with `&CURLIB` and run `liblist -c ${lib}` at compile time to set the current library.

szsascha:
@worksofliam 

Sorry, but I can't really form an opinion on this at the moment. For this I lack a bit of background on how the whole plugin is working.

I would assume that it makes the most sense to have the library list the same like you would have it when you compile it on i. This includes the CUR and SYS libraries. You need this for some cases. For example when you are using an object from QSYS without specifying a library.

I also had a look on the qsh liblist command documentation (https://www.ibm.com/docs/en/i/7.4?topic=jobs-liblist)
It says: "Remove library from the user portion of the library list or **unset the current library if the -c option is also specified.**"
So I guess it's possible to remove it?
The SYS part should be set automatically by the system.

How do we want to continue?

szsascha:
> We completely replace &BUILDLIB with &CURLIB and run liblist -c ${lib} at compile time to set the current library.

Sounds like a plan! Do we then still need &BUILDLIB for anything?

worksofliam:
@szsascha Let's go with adding `&CURLIB`. In fact, I can do this change for us.

I think we will keep `&BUILDLIB` around for backwards compatibility (since some Actions might use them), but it will be the same as `&CURLIB`.

szsascha:
> When I try and compile on a new connection with this change:
>
> Library list: QSYS2 QGPL QTEMP
> Command: CRTBNDRPG PGM(BARRY/BETH) SRCFILE(BARRY/QRPGLESRC) OPTION(*EVENTF) DBGVIEW(*SOURCE) TGTRLS(*CURRENT)
> liblist: 001-0083 Error CPF2104 found deleting library QSYS2 from library list.
> liblist: 001-0081 Error CPF2103 found adding library QSYS2 to library list.

I think we can completely ignore the SYS list at this point. The system will handle it by itself. You're also not able to change it with EDTLIBL.

szsascha:
> @szsascha Let's go with adding &CURLIB. In fact, I can do this change for us.

@worksofliam Ok, thank you! I already have an idea for my next contribution :)

worksofliam:
@szsascha Awesome - I can't wait to see it! My PR will be ready within the next 30 minutes or so. I will ping you so you can see it.

I also want to mention that I am sorry about how confusing this initial PR was. The app is quite big now, and playing with the library list affects the core piece of the extension (e.g. the compliation step).

Hope to make it easier in the future.

PR #149, opened 2021-05-29
Use current library over build library

worksofliam:
### Changes

This fix will rename 'build library', to 'current library'. We're doing this for many reasons:

* It makes more sense to IBM i devs
* We can actually set the builds current library at compile time now

### Checklist

* [X] have tested my change

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -109,7 +109,8 @@ Notice the special identifiers in the command begining with `&`. These identifie
 
 | Variable | Usage                              |
 |----------|------------------------------------|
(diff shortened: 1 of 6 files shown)


worksofliam:
@szsascha This is the PR. You're more than welcome to take a look and test it.

I've tested it and seems to work as expected. I also updated the button labels as well as the docs.

szsascha:
@worksofliam also works for me

But was really a bit more complex than assumed at the beginning ;)

worksofliam:
Lovely. I am going to merge it in and it should be out sometime this week.

Issue #45, opened 2021-02-26
Allow me to download / upload files to my PC with the IFS browser. 


richardschoen:
It would be nice to quickly download a copy of an IFS file or regular source member for offline editing.

Not high priority, but would be cool

Download selected IFS file:
Right click on IFS file, select download, prompt where to save to locally.

Upload selected IFS file:
Right click IFS file and select upload, prompt for local file to upload. Will replace contents of selected IFS file. 

Upload File by Selecting Upload option on dir itself
Right click dir name, select upload, prompt for file name, upload the IFS file into the selected directory. Prompt for replace if IFS file already exists.

worksofliam:
I'd class this as implemented.

PR #150, opened 2021-05-29
Basic IFS up- and download functionality

szsascha:
### Changes

Basic IFS upload and download functionality implemented as requested in #45.
Very simple but a start.

You're able to download files from the IFS browser with right-click and "Download".

There are two way's to upload files:
1. Use the upload icon in the IFS browser header
2. Use right-click and "Upload" on a directory

During development, I made a note of the following points to consider for the future:
1. Does it work correctly on Windows? I've tested it on macOS. Maybe there are some path specialties on another OS?
2. I've not tested the up- and download of files with a large size. The IBMiContent class is very basic at the moment and currently not supporting progress tracking.
3. Up- and downloading of directories didn't work yet
4. Only utf8 files are supported at the moment because of the configurations in IBMiContent. Maybe we've to create a new module or we've to extend the IBMiContent class.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/views/ifsBrowser.js
+++ b/src/views/ifsBrowser.js
@@ -243,7 +243,7 @@ module.exports = class ifsBrowserProvider {
 
           } else {
             //Get filename from path on server
-            const filename = node.path.replace(/^.*[\\\/]/, '');
+            const filename = node.path.replace(/^.*[\\\/]/, ``);
 
             const remoteFilepath = path.join(os.homedir(), filename);
 
--- a/src/views/ifsBrowser.js
+++ b/src/views/ifsBrowser.js
@@ -3,6 +3,7 @@ const { throws } = require(`assert`);
 const vscode = require(`vscode`);
 const os = require(`os`);
 const path = require(`path`);
+const fs = require(`fs`);
 
 let instance = require(`../Instance`);
 const Configuration = require(`../api/Configuration`);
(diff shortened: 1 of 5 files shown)


worksofliam:
This is awesome! I am going to do a review of this tomorrow.

Thanks @szsascha!

worksofliam:
About to review this today!

src/views/ifsBrowser.js
```diff
@@ -108,6 +111,59 @@ module.exports = class ifsBrowserProvider {
...
+        if (!destPathSuggestion.endsWith(`/`)) destPathSuggestion += `/`;
+        destPathSuggestion += filname;
+
+        const isStillOpen = vscode.workspace.textDocuments.find(document => document.uri.path === originPath[0].fsPath);
+
+        if (isStillOpen) {
+          //Be sure it's correctly saved
+          vscode.window.showInformationMessage(`Cannot upload streamfile while it is open.`);
+        } else {
+          const destinationPath = await vscode.window.showInputBox({
```
worksofliam, line 139: This is working great, unless the user presses cancel.

After this is done executing, you need to check if has a value, otherwise the code is erroring out:

```js
const destinationPath = await vscode.window.showInputBox({...});
if (destinationPath) {
  //...
} else {
  //...
}

**Comment by worksofliam** [comment]
@szsascha Other than the one issue I noted about, this is an awesome PR! Great work and a great feature.

If you're happy with it after you've made my recommended fix, then let me know and we can merge. If you'd like to add more to it, let me know. Maybe the following stuff can be done in another PR?

As a reference to all my notes, the library we are using is node-ssh: https://github.com/steelbrain/node-ssh#readme

> I've not tested the up- and download of files with a large size. The IBMiContent class is very basic at the moment and currently not supporting progress tracking.

Tracking progress isn't something we've needed before, because we're really just working with small (<5m) source files most of the time. Maybe you could look into the library we are using to see if they have support for tracking progress. 

> Up- and downloading of directories didn't work yet

The node-ssh library does support this. You should see about extending IBMiContent to support it also.

> Only utf8 files are supported at the moment because of the configurations in IBMiContent. Maybe we've to create a new module or we've to extend the IBMiContent class.

I believe extending the class would be nice. Maybe [downloadStreamfile](https://github.com/halcyon-tech/code-for-ibmi/blob/master/src/api/IBMiContent.js#L24) and [uploadStreamfile](https://github.com/halcyon-tech/code-for-ibmi/blob/master/src/api/IBMiContent.js#L32) could just have an extra parameter so we could pass [the encoding we want](https://stackoverflow.com/a/14551669/4763757), with a default of `utf8`?

```
async downloadStreamfile(remotePath, encoding = `utf8`) {
    const client = this.ibmi.client;
    let tmpobj = await tmpFile();
    await client.getFile(tmpobj, remotePath);
    return readFileAsync(tmpobj, encoding);
}
```

src/views/ifsBrowser.js
szsascha: Thank you for your feedback! I've just corrected it.

Now the whole upload process will be stopped if no destination path was set. Hope you're fine with it :)

szsascha:
@worksofliam 

Thanks a lot for your review, feedback and answers.

Sounds like there's a lot of potential there. Maybe in the future there will be some contributions from me :)

From my point of view, the PR can now be merged.

worksofliam:
@szsascha awesome! I hope we will see more contributors too!

I will do another quick test with your change and we'll get it merged in today!

You rock!

worksofliam:
I also added some additional commits to:

* Use the client to upload/download. This means it won't even touch the encoding and will keep to whatever the file is.
* Change the command name to follow the standard of the other IFS commands.

PR #173, opened 2021-06-10
Feature/member by path

### LINKED ISSUE #170: Is there a way to open a Source Member directly without navigating through the Member Browser?
Hi guys,

I'm looking for a way to directly open a source member without needing to scroll through the member browser and click on one. 

In RDi this can be achieved with the hotkey ctrl+alt+q which pulls up a window like this
image

Does there currently exist a way to do this with CODE for IBM i? If not could it be considered for a feature addition? Sometimes a Source File can contain a ton of members and it can take some time to scroll through and pinpoint the one I need to open.

Thanks for your help, this extension is awesome!

connorholyday:
### Changes

Adds a new command to open a source member by path.

### Checklist

* [x] have tested my change

### Next Steps
- Extend this to also work for the IFS.
- Add a section to the README for custom keyboard shortcuts.

Closes #170

### DIFF
--- a/package.json
+++ b/package.json
@@ -567,7 +567,7 @@
 			},
 			{
 				"command": "code-for-ibmi.openMemberByPath",
-				"title": "Open Source File",
+				"title": "Open Source Member",
 				"category": "IBM i",
 				"icon": "$(files)"
 			},
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
 		"onCommand:code-for-ibmi.updateMemberText",
 		"onCommand:code-for-ibmi.renameMember",
 		"onCommand:code-for-ibmi.searchSourceFile",
+		"onCommand:code-for-ibmi.openMemberByPath",
 		"onView:ifsBrowser",
 		"onCommand:code-for-ibmi.refreshIFSBrowser",
 		"onCommand:code-for-ibmi.changeHomeDirectory",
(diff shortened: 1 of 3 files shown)


src/views/memberBrowser.js
```diff
@@ -340,6 +340,22 @@ module.exports = class memberBrowserProvider {
...
         }
+      }),
+
+      vscode.commands.registerCommand(`code-for-ibmi.openMemberByPath`, async () => {
+        const searchFor = await vscode.window.showInputBox({
+          prompt: `Enter source path (Format: LIB/SPF/NAME.ext)`
+        });
+
+        if (searchFor) {
+          const [path] = searchFor.match(/\w+\/\w+\/\w+\.\w+/);
```
worksofliam, line 351: Looks like if this format is not correct, `match` actually throws an error which causes an error message to appear. 

That is something we should handle before we merge.

package.json
```diff
@@ -564,6 +565,12 @@
 				"title": "Search Source File",
 				"category": "IBM i"
 			},
+			{
+				"command": "code-for-ibmi.openMemberByPath",
+				"title": "Open Source File",
```
worksofliam, line 570: I think we should change the title to say `Open source member`, since a source file is just a folder of source members.

worksofliam (changes requested):

worksofliam:
I am gonna get this one does so I can work on the IFS open part too.

worksofliam (approved):

PR #176, opened 2021-06-10
Allow files with poundsigns to be opened (#169)

### LINKED ISSUE #169: Unable to open files with a '#' in the name
**Describe the bug**
The extension appears unable to open files that contain a '#' in the name. From some experimenting, this seems to include:

- A '#' in the containing library, file or member name
- A '#' in the path of a stream file, (both file and directory names)

Trying to open such a file doesn't result in any error messages. The editor just doesn't load.

**To Reproduce**
Find (or create if necessary) a source file with a '#' in the name and attempt to open it.

**Expected behavior**
The file is loaded into the editor.

**Environment (please complete the following information):**
 - Extension version 0.5.3
 - IBM i OS version 7.4

**Additional context**
I've had a little look at the code but couldn't see any obvious reason. This looked like it might be the relevant bit?

https://github.com/halcyon-tech/vscode-ibmi/blob/67cb13add499c5d1cdbea9f74df827b4ad43375f/src/Instance.js#L241

My expectation is it is probably not being escaped somewhere and being treated as a comment in a shell command. Hopefully that's not in the VS Code internals!

Our change management solution uses '#' in the library names, so a solution would very useful 🙂

worksofliam:
### Changes

In IBM i land, it is not uncommon for files to include weird characters - in this case, a pound sign, or hash tag, depending on where you're from.

This PR fixes the issue where [VS Code's Uri class](https://code.visualstudio.com/api/references/vscode-api#Uri) was parsing a path such as `BARRY/QRPGLESRC/PGM#01.rpgle` as `BARRY/QRPGLESRC/PGM` and treating `01.rpgle` as the Uri fragment. Same goes for IFS paths.

This will close #169.

### Checklist

* [X] have tested my change

### DIFF
--- a/src/Instance.js
+++ b/src/Instance.js
@@ -243,9 +243,9 @@ module.exports = class Instance {
             let uri;
             if (path.startsWith(`/`)) {
               //IFS
-              uri = vscode.Uri.parse(path).with({scheme: `streamfile`});
+              uri = vscode.Uri.parse(path).with({scheme: `streamfile`, path});
             } else {
-              uri = vscode.Uri.parse(path).with({scheme: `member`});
+              uri = vscode.Uri.parse(path).with({scheme: `member`, path: `/${path}`});
             }
   
             try {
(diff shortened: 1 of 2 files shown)


onewheelonly:
Seems to work great! 🎉 

Thanks for such a speedy fix - I'm sure there are quite a few of my colleagues who will be keen to use this now.

Issue #194, opened 2021-06-16
Ability to create library from object view


worksofliam:

szsascha:
@worksofliam You're assigned to this issue. Are you working currently on this?
I've a bit of free time at the moment and planning to work on vscode for ibmi again. This seems a good issue to me. Maybe I want to implement this. But I'm not sure.

worksofliam:
@szsascha feel free to take it on! I'd say this is a nice easy thing to work on.

PR #228, opened 2021-08-02
Library creation via object browser added

szsascha:
### Changes

Libraries can now be created via the object browser. You'll find a new + icon on the right of "OBJECT BROWSER". This feature was requested in #194.

Please check the following things before merging:
1. Is it working properly? Unfortunately it's not allowed to create new libraries on PUB400. So I couldn't test it properly.
2. Is the "+" icon good for this use case? In my opinion the icon with the directory and a "+" would fit better but it was already used.

### Checklist

* [x] have tested my change


worksofliam:
Working good. Thanks for your work here @szsascha!

Issue #297, opened 2021-09-17
Generate DDL from object browser


worksofliam:
Right-clicking on a database object and clicking Generate SQL should be a thing

szsascha:
Do you mean the generation of the CREATE TABLE SQL?

Do you know if there are APIs / tools that can be used for this?

worksofliam:
@szsascha yes, there is!

https://www.ibm.com/docs/en/i/7.4?topic=services-generate-sql-procedure

szsascha:
I've just looked more into it.

My current plan is to open a new, unsafed editor tab with the CREATE TABLE statements after rightclicking a file in the object browser.

Is there already a piece of code in the extension where a new, unsafed editor tab is opened? I already tried the following but it leaves without exception on Uri.parse. https://stackoverflow.com/questions/41068197/vscode-create-unsaved-file-and-add-content

worksofliam:
@szsascha Yes! The search results open up in a brand new tab - check that out!

szsascha:
@worksofliam I just implemented this feature. Would be really nice if you could have a look on it :)

worksofliam:
I'll check it out!

worksofliam:
Part of the Db2 for i extension. It will not be added to the base.

PR #301, opened 2021-09-18
Add functionality to generate CREATE TABLE SQL's from object browser

szsascha:
### Changes

It's now possible to generate CREATE TABLE SQL's from the object browser. This feature was requested in #297.

Things to consider for the future:
1. Any type of object shows the "Generate CREATE TABLE SQL" entry on right clicking. But it will only work for files. Other objects will show an error message.
2. The new tab containing the CREATE TABLE statement cannot be saved by STRG / CMD + S. Don't know the exact reason yet.
3. Maybe it will make sense to add this feature also to the database browser.

### Checklist

* [x] have tested my change

### DIFF
--- a/package.json
+++ b/package.json
@@ -565,6 +565,11 @@
 				"category": "IBM i"
 			},
 			{
+				"command": "code-for-ibmi.generateCreateTableSQL",
+				"title": "Generate CREATE TABLE SQL",
+				"category": "IBM i"
+			},
+			{
 				"command": "code-for-ibmi.addToLibraryList",
 				"title": "Add to library list",
 				"category": "IBM i",
@@ -1114,6 +1119,11 @@
 					"command": "code-for-ibmi.runAction",
 					"when": "view == objectBrowser && viewItem == object",
 					"group": "1_workspace@1"
+				},
+				{
+					"command": "code-for-ibmi.generateCreateTableSQL",
+					"when": "view == objectBrowser && viewItem == object",
+					"group": "2_workspace@1"
 				},
 				{
 					"command": "code-for-ibmi.removeSchemaFromDatabaseBrowser",
--- a/src/Instance.js
+++ b/src/Instance.js
@@ -388,6 +388,37 @@ module.exports = class Instance {
           })
         );
 
+        context.subscriptions.push(
+          vscode.commands.registerCommand(`code-for-ibmi.generateCreateTableSQL`, async (node) => {
+            if(!node) return;
+
+            if (node.type != `FILE`) {
+              vscode.window.showErrorMessage(`CREATE TABLE can only be generated for file's.`);
+              return;
+            }
+
+            let lib = node.path.split(`/`)[0];
+            let file = node.path.split(`/`)[1];
+            
+            vscode.window.showInformationMessage(`Generating CREATE TABLE statement...`);
+            let result = await this.getContent().runSQL(`CALL QSYS2.GENERATE_SQL ('${file}', '${lib}', 'TABLE')`);
+
+            let resultString = ``;
+            result.forEach(element => resultString += element.SRCDTA + `\r\n`);
+
+            try {
+              const textDoc = await vscode.workspace.openTextDocument(vscode.Uri.parse(`untitled:` + `Generated SQL`));
+              const editor = await vscode.window.showTextDocument(textDoc);
+              editor.edit(edit => {
+                edit.insert(new vscode.Position(0, 0), resultString);
+              })
+
+            } catch (e) {
+              vscode.window.showErrorMessage(`Error generating CREATE TABLE statement.`);
+            }
+          })
+        );
+
         
 
         initialisedBefore = true;


worksofliam:
@szsascha While this is awesome, I think it can be extended further!

When we use the database browser, we actually know the type of object in the database (e.g. `TABLE`). I think right clicking on objects in the database browser should also have this object and then `TABLE` would not have to be hardcoded.

szsascha:
@worksofliam 

I am just wondering if it makes sense to have the CREATE TABLE option in the object browser AND the database browser. Wouldn't this function be a perfect fit for the database browser? 
I don't like my own implementation in the object browser at the moment and what you mentioned is one reason for this. The other reason is that the option to generate a CREATE TABLE also shows up on other objects. Both can be avoided by implementing it only into the database browser. And there is where it should be, in my opinion.

worksofliam:
> I am just wondering if it makes sense to have the CREATE TABLE option in the object browser AND the database browser. Wouldn't this function be a perfect fit for the database browser? 

I think having this functionality in both browsers is an excellent idea!

> Both can be avoided by implementing it only into the database browser. And there is where it should be, in my opinion.

I think that's a good idea! Please take your time with this PR and if you need some help from me, just let me know!

szsascha:
@worksofliam 

CREATE TABLE SQL generation is now also available in the database browser. Please check and let me know if it's working fine. :)

worksofliam:
I will be reviewing this today!

package.json
```diff
@@ -564,6 +564,16 @@
 				"title": "Run Action",
 				"category": "IBM i"
 			},
+			{
+				"command": "code-for-ibmi.generateCreateTableSQLFromDatabaseBrowser",
+				"title": "Generate CREATE TABLE SQL",
```
worksofliam, line 569: This isn't specific to `create table` since it generates the code for many objects.

package.json
```diff
@@ -1115,9 +1125,20 @@
...
 					"group": "1_workspace@1"
 				},
+				{
+					"command": "code-for-ibmi.generateCreateTableSQLFromObjectBrowser",
+					"when": "view == objectBrowser && viewItem == object",
+					"group": "2_workspace@1"
+				},
+				{
+					"command": "code-for-ibmi.generateCreateTableSQLFromDatabaseBrowser",
+					"when": "view == databaseBrowser && viewItem == table",
```
worksofliam, line 1135: Is`viewItem` correct, since it applies to more than just tables?

src/views/databaseBrowser.js
```diff
@@ -55,6 +55,57 @@ module.exports = class databaseBrowserProvider {
...
+        case `T`:
+          type = `TABLE`
+          break;   
+        case `V`:
+          type = `VIEW`
+          break;      
+        }
+
+        if (type == null) {
+          vscode.window.showErrorMessage(`CREATE TABLE cannot be generated for ${node.table.type}`);
```
worksofliam, line 84: Since this is not specific to a table, the message should instead read `Cannot generate SQL for ${type}`

src/views/databaseBrowser.js
```diff
@@ -55,6 +55,57 @@ module.exports = class databaseBrowserProvider {
...
+
+        try {
+          const textDoc = await vscode.workspace.openTextDocument(vscode.Uri.parse(`untitled:` + `Generated SQL`));
+          const editor = await vscode.window.showTextDocument(textDoc);
+          editor.edit(edit => {
+            edit.insert(new vscode.Position(0, 0), resultString);
+          })
+
+        } catch (e) {
+          vscode.window.showErrorMessage(`Error generating CREATE TABLE statement.`);
```
worksofliam, line 105: Same with the messaging here.

src/views/objectBrowser.js
```diff
@@ -142,6 +142,35 @@ module.exports = class objectBrowserProvider {
           //Running from command
           console.log(this);
         }
+      }),
+
+      vscode.commands.registerCommand(`code-for-ibmi.generateCreateTableSQLFromObjectBrowser`, async (node) => {
```
worksofliam, line 147: Please see my final review message about this command.

worksofliam (changes requested):
Overall, this is a great feature and idea. As noted, some of the messaging needs cleaning up. 

My last request is about duplicate code and functionality. In this PR, we basically have two commands that do the same thing. I believe the functionality should exist in one place and that only one command should exist for this.

I think the `generateCreateTableSQL*` commands should be merged into one so there is less duplicate code.

Let me know what you think.

szsascha:
@worksofliam Thank you. Will have a look at it soon. But maybe later next week or next weekend.

package.json
szsascha: @worksofliam 
Unfortunately I couldn't find a list or something that describes this line. For my interpretation of the code, a "table" is anything in the treeview below the schema.
Do you know where I can get more information regarding this?

szsascha:
@worksofliam What do you think about using the term "generate creation SQL" in all places?

In my opinion "generate SQL" is too general for this.

Regarding duplicate code:
I wouldn't create a single function for both functions in this case. I think this place tends to get messy.
But maybe we can create a one or more utility functions out of it.

What do you think?

worksofliam:
> In my opinion "generate SQL" is too general for this.

It may be too general, but that's because it works for pretty much all SQL / IBM i database objects. Also, the Schemas functionality inside of ACS uses this naming too - which is part of the reason I think we should use the same.

Screenshot 2021-09-20 at 4 13 24 PM

> I wouldn't create a single function for both functions in this case. I think this place tends to get messy.
But maybe we can create a one or more utility functions out of it.

I think a generic function could be created which really just wraps `GENERATE_SQL` which would be called by commands:

```js
function generateSQL(schema, object, type) {
  const result = await content.runSQL(`CALL QSYS2.GENERATE_SQL ('${schema}', '${object}', '${type}')`);
  return result.map(row => row.SRCDAT).join(`\n`);
}
```

I personally think this function could live in `IBMiContent.js` too - but really anywhere is fine!

Also, I noticed you do this in your PR:

```js
        let resultString = ``;
        result.forEach(element => resultString += element.SRCDTA + `\r\n`);
```

I think that this could be replaced with what I do in the example function I provided:

```js
const fileData = result.map(row => row.SRCDAT).join(`\n`);
```

Let me know what you think.

package.json
szsascha: @worksofliam Sorry, I don't want to be annoying :D Maybe you didn't see my question?

PR #305, opened 2021-09-19
Variant Characters

SJLennon:
### Changes

1. Cleaned up unused .PNG files from previous update.
2. Added a section on Variant Characters.  **_This needs close review._** There seems to be little effective IBM documentation on PASE environment variables.  From what I can deduce, ILE variables get copied to PASE when the shell starts, so I'm not sure why you need to restart the IBM i, and perhaps this should not be mentioned.

### Checklist

* [x] have tested my change


worksofliam:
@benetti-smeup what do you think of this change to the documentation?

Perhaps we should also add links to IBM documentation about this topic.

benetti-smeup:
the readme seems clear to me.
the problem with IBM documentation is that there are different links and the solution is difficult to understand.

SJLennon:
I did a lot of searching for something definitive on IBM.COM but did not succeed.  I would really like to have included a link. Basically I wrote up the trial and error results from the issues.

Wonder if asking on Ryver would come up with anything?

worksofliam:
@SJLennon Feel free to ask.. otherwise I will likely give a merge a bit later!

SJLennon:
Still figuring out Ryver, but I see there is a ILE-PASE Interop forum, though it doesn't seem particularly active.
There is also PASE development forum that seem more alive.  I think I'll try a question there.

worksofliam:
@SJLennon 

Might be worth noting these simple steps:

> 1. end sshd, 2. WRKENVVAR *JOB set PASE_LANG to the correct local and QIBM_PASE_CCSID to 1208, 3. start sshd?

https://www.ibm.com/docs/en/i/7.2?topic=ssw_ibm_i_72/apis/pase_locales.htm

SJLennon:
I'll do some more research and revise the help. Not sure WRKENVVAR *JOB will work, as it applies to the current job.  You probably need a CL with either ADDENVVAR or a CHGENVVAR and then starts SSHD.

worksofliam:
@SJLennon please let me know when this is ready for review again.

SJLennon:
@worksofliam Here's a revised versions, based on Kevin Adler's comments. This has taken a lot of time, possibly for relatively few users. Let me know what you think.

SJLennon:
There is no mention of PASE CCSID variable, based on the last comment from Kevin.
"I did a bunch of digging and cannot come up with any reason changing the PASE CCSID should make any difference"

priceaj:
Whats a Variant Character?

SJLennon:
This page probably does a better job of explaining it than anything I can do, even though it is not IBM i specific:
https://www.ibm.com/docs/en/db2-for-zos/11?topic=ccsids-variant-characters

See also this page, the section "Code page layout":
https://en.wikipedia.org/wiki/EBCDIC

And here's another one that might help shed light on the issue:
https://www.ibm.com/support/pages/what-impact-changing-qccsid-shipped-65535-another-ccsid

priceaj:
I'd never heard that term before, but I have a list of the "Variant characters" between CP 37 and CP 285 on a wiki page at work entitled "CCSID Issues". 

Do you think telling people not to use Variant characters without stating what they are (or which characters are invariant and should be used) is [a big ask](https://dictionary.cambridge.org/dictionary/english/a-big-ask)?

For instance in my experience (of 37 vs 285) Hash (#) and At (@) are not variant characters, and I would therefore not know that they should not be used.

SJLennon:
I don't think I'm telling people not to use variant characters, just saying, as IBM is, that it is not a good practice. Clearly some people have done it, and it occasionally causes problems.  Would it be better/clearer if the heading was "Variant Characters/CCSID Issues"?

priceaj:
I think saying something is not good practice strongly implies that you should not do it, but in this case the bad practice is not clearly defined... or even a good practice well defined. 

I think it just needs a bit more explination, do we have a list of variant (or invariant) characters which could be supplied?

SJLennon:
@worksofliam I have made some changes to remove text about bad practice, based on comments from @priceaj . Also changes the heading and added some links to background info on Variant characters.

worksofliam:
@SJLennon I can see images have been deleted in this PR. Is that intentional?

SJLennon:
Yes.  I added new, replacement images in the prior PR.  This clears out the old, unused images.

worksofliam:
Thanks @SJLennon !

PR #308, opened 2021-09-20
Credential saving

priceaj:
### Changes

Implemented credential saving in secret credential storage, credentials will no longer be stored in settings.json for any extention to read. 

This adds a checkbox to the connection creation screen, which when checked will save your password/certificate in secure storage. 

In windows the saved credential can be seen (and edited) in "Control Panel > Credential Manager > Windows Credentials"

When deleting the connection the credential will also be removed.

Also fixed a bug when editing "Connecton Settings"

### Checklist

* [X] have tested my change
* * Tested using password only... I only have access to PUB400 via password login, therefore I cannot test a certificate based login

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -250,8 +250,7 @@ Here is a snippet of what the connection details look like:
     "name": "My IBMi Connection",
     "host": "DEV400",
     "port": 22,
-    "username": "OTTEB",
-    "privateKey": null
+    "username": "OTTEB"
   }
 ],
 ```
--- a/package.json
+++ b/package.json
@@ -524,6 +524,11 @@
 				"category": "IBM i"
 			},
 			{
+				"command": "code-for-ibmi.showLoginSettings",
+				"title": "Login settings",
+				"category": "IBM i"
+			},
+			{
 				"command": "code-for-ibmi.deleteConnection",
 				"title": "Delete connection",
 				"category": "IBM i"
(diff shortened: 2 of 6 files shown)


priceaj:
@onewheelonly if you get a chance, would you be able to verify this works with SSH Keys as well!

worksofliam:
@priceaj I haven't reviewed this yet, but we warned that this may clash after #307 gets merged.

Will let you know when I get around to look at it in a few hours.

priceaj:
> @priceaj I haven't reviewed this yet, but we warned that this may clash after #307 gets merged.
> Will let you know when I get around to look at it in a few hours.

Looks like it will as there is currenly no way to edit a password/credential without removing and recreating the connection... so this doesnt cater for the new form.

worksofliam:
@priceaj I am just getting to looking at this now and it is awesome!

It's a really simple change too. I learned about `secrets`.

I am about to merge the Login settings PR (#307) and I wonder if you could update your PR to follow suit and apply further changes to the Login settings panel to update the secrets too?

priceaj:
@worksofliam run across a slight issue, seems that the SSH Key stuff works differently than I thought, as it saves a Path to the Public Key File and does not read the actual file in.

I could read the file and save it into the secret storage, but I think this might actually be less secure, as if you are managing the Key files in a secure way, making a copy of it might just make it harder to keep track of your private keys!

I'm tempted to just revert all my changes and rewrite focusing just on password storage

priceaj:
Closing this, will rewrite and re-request

dennisnel:
@priceaj, I don't know if you want to look at [this project](https://github.com/SchoofsKelvin/vscode-sshfs)? I have found this to be very useful and works pretty well.

I like that you can mount the remote folders as local workspace folders and there is a form for the connection etc and an integrated remote terminal

worksofliam:
@priceaj i am sure not why that use has pasted the same comment 3 times on different issues/PRs.

PR #310, opened 2021-09-21
Save passwords

priceaj:
### Changes

Implemented secret password storage. 

There is now the option to save a password for a new connection. 

For existing connections one can be set/changed via the new login settings page. 

Deleting a connection will also delete the saved password

"Save passwords" uses the VS Code secrets API and from windows can be seen/removed/amended in "Control Panel > Credential Manager > Windows Credentials"

### Checklist

* [X] have tested my change

### DIFF
--- a/src/webviews/settings/index.js
+++ b/src/webviews/settings/index.js
@@ -216,6 +216,7 @@ module.exports = class SettingsUI {
               context.secrets.store(`${name}_password`, `${data.password}`)
             };
 
+            delete data.password;
 
             connection = {
               ...connection,
--- a/src/webviews/settings/index.js
+++ b/src/webviews/settings/index.js
@@ -189,6 +189,13 @@ module.exports = class SettingsUI {
           field.default = connection.username;
           ui.addField(field);
 
+          field = new Field(`paragraph`, `authText`, `Only provide either the password or a private key - not both.`);
+          ui.addField(field);
+
+          field = new Field(`password`, `password`, `Password`);
+          field.description = `Only provide a password if you want to update an existing one or set a new one.`
+          ui.addField(field);
+
           field = new Field(`file`, `privateKey`, `Private Key`);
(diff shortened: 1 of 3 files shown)


worksofliam:
About to give this a go now.

src/webviews/settings/index.js
```diff
@@ -204,6 +211,12 @@ module.exports = class SettingsUI {
...
 
+            if(data.password && !data.privateKey) {
+              context.secrets.delete(`${name}_password`);
+              context.secrets.store(`${name}_password`, `${data.password}`)
+            };
+
+
             connection = {
               ...connection,
               ...data
```
worksofliam, line 222: Slight issue here.

When the password is passed in, it is being stored in the general connection storage. 

Maybe when calling `store` on line 216, you could also `data.password = undefined`?

worksofliam (changes requested):
See above comment.

src/webviews/settings/index.js
priceaj: ah yes, good spot... the exact opposite of what we are trying to do!

It looks like the new connection function stores explicitly named fields, but this new forms overlays the data object over the existing connection object, which is adding the new field. 

something I need to watch out for I guess!

I'm amending the code to delete the password property out of the data object completely in all cases

priceaj:
See above comment.

priceaj:
@worksofliam if you are interested under the hood VS Code use https://github.com/atom/node-keytar to store the secrets.

"On macOS the passwords are managed by the Keychain, on Linux they are managed by the Secret Service API/libsecret, and on Windows they are managed by Credential Vault."

So I think on MacOS you can use this https://support.apple.com/en-gb/guide/keychain-access/kyca1083/mac to view and manage the created password

worksofliam (approved):
Looking and working great!

Issue #336, opened 2021-09-26
Config to change default wait time for connection 


worksofliam:
As suggested by @ChrisHird, we should add configuration to let the user change the wait time for the SSH job.

https://www.ibm.com/docs/en/i/7.2?topic=jobs-getjobid

```
# Store the job ID from this. Happens when connection is made
getjobid $$ 

# Use the JOB id here. Happens when connection is made
system "CHGJOB JOB(x) DFTWAIT(x)"

# Any future system commands will inherit it
system "CPYTOSTMF ..."
```

worksofliam:
There is a bit of an issue. I didn't know this, but it looks like `paseCommand` (node_ssh / ssh2 `exec` / `execCommand`) is actually forking the connection process when running a command.

I'm at a point here where I am unable to get the actual connection job ID because it's running the `echo` in a forked process? 

Happening here in this PR:

https://github.com/halcyon-tech/vscode-ibmi/pull/337/files#r717036957

worksofliam:
It may have something to do with channels being opened for every command. See `exec` here: https://github.com/mscdex/ssh2#client-methods

ChrisHird:
Not sure if I am reading it correctly but you are doing 2 await requests at the same time, so they would be different requests and probably fork to different processes???

If the request is not put in an await request the process should be the same as the one the requests run under? If a request comes in and the jobs is tied up waiting to respond to the previous request it probably spawns another job???

That would make sense with my limited knowledge.  😊 Not sure if that helps or not..

Chris…

Dman247:
Yes, I am seeing both `exec()` and `execCommand()`  create separate processes as well.

Next steps I would consider are to run the copy commands, that error out due to lock, through a shell.
There, you can definitively grab the jobid for the shell that was spawned and perform the CHGJOB on it.
All subsequent commands running through that shell should have the same job attributes the shell had. (I've done a primitive test of this and it seems to work)

Otherwise, I don't know how you would identify the right sshd job to change the DFTWAIT with, because we have to assume there could be numerous ssh connections on the system:
![image](https://user-images.githubusercontent.com/38368303/135002051-2077c4a9-cf4a-4d4f-a7f5-cb2123063ef2.png)

In the example above example here, one of those is my ssh connection from the extension and another is an ssh for a remote shell.

What is interesting, the IBMi job that supports the connection back to the extension has something unique logged in the job log.
[...]

PR #337, opened 2021-09-27
Custom wait time

worksofliam:
### Changes

New configuration which will run a `CHGJOB` with `DFTWAIT` time on the connection process.

Currently broken. See #336 for more information.

### Checklist

* [ ] have tested my change

### DIFF
--- a/src/api/IBMi.js
+++ b/src/api/IBMi.js
@@ -11,7 +11,7 @@ const remoteApps = [
   },
   {
     path: `/usr/bin/`,
-    names: [`setccsid`]
+    names: [`setccsid`, `getjobid`]
   }
 ];
 
@@ -21,8 +21,13 @@ module.exports = class IBMi {
     this.currentHost = ``;
     this.currentPort = 22;
     this.currentUser = ``;
+
+    /** @type {string|undefined} */
+    this.jobId = undefined;
     
     this.tempRemoteFiles = {};
+
+    /** @type {string[]} */
     this.defaultUserLibraries = [];
 
     /** @type {vscode.OutputChannel} */
@@ -43,7 +48,8 @@ module.exports = class IBMi {
       db2util: undefined,
       git: undefined,
       grep: undefined,
-      setccsid: undefined
+      setccsid: undefined,
+      getjobid: undefined
     };
   }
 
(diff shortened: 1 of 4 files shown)


src/api/IBMi.js
```diff
@@ -255,6 +261,49 @@ module.exports = class IBMi {
...
+          progress.report({
+            message: `getjobid found. Fetching connection job ID.`
+          });
+
+          const command = this.remoteFeatures.getjobid;
+
+          try {
+            let jobId;
+            jobId = await this.paseCommand(`echo $$`);
+            jobId = await this.paseCommand(`echo $$`);
```
worksofliam, line 274: This second call to `echo $$` returns a new process ID. I did not know it worked like this.

src/api/IBMi.js
```diff
@@ -294,7 +343,7 @@ module.exports = class IBMi {
    * @param {string} [directory] If not passed, will use current home directory
    */
   remoteCommand(command, directory) {
-    //execCommand does not crash..
+    command = command.replace(/\$/g, `\\$`);
     return this.paseCommand(`system "` + command + `"`, directory);
   }
```
Dman247, line 349:
```suggestion
  /**
   * 
   * @param {string} path 
   * @param {string} tempRmt
   * @param {string} ccsid
   * @param {number} dftwait
   */
  async cpyToStmF(path, tempRmt, ccsid, dftwait) {
    const shell = await this.client.requestShell();

    shell.stdout.on(`data`, (data) => {
      data = data.toString();
      console.log(`OUT:` + data);
      if (data.startsWith(`CPCA082`)) {
        // Success 
        // TODO: Emmit success event 
      } else if (data.startsWith(`CPF`)) {
        // Failed
        // TODO: Emmit error event
        console.log(`^^^^^^ERROR^^^^^^\n\n`);
      };
    });

    let command = `system "CHGJOB JOB($(getjobid -s $(echo $$))) DFTWAIT(${dftwait})";`;
    command += `system "CPYTOSTMF FROMMBR('${path}') TOSTMF('${tempRmt}') STMFOPT(*REPLACE) STMFCCSID(1208) DBFCCSID(${ccsid})";`;
    command += `exit\n` ;
    shell.write(command);
  }

```

Dman247: The suggestion I left is a method inside the IMBi.js file.
That method will utilize a shell to, first, set the DFTWAIT attribute of the shell, then the subsequent commands it runs ( this case only one) will inherit the shells job attributes.

What I don't know how you'd like to handle is the success/failure. Because, in the current implementation we handle the failure of the copy command through a catch block. Now, this won't work because we have to wait with a listener for the shell's response. ```shell.stdout.on(`data`, (data) => {``` 
I think we are going to need to emit something, then somewhere else listen to take the appropriate action for success/failure.

The success code is `CPCA082`
image

The failure code(s) are `CPFA09E` & `CPFA097`
image

But, the good news is from what I was testing I was able to get the failure response almost immediately, rather than the 30 seconds.

worksofliam: @Dman247 Seems like it might be worth me giving you access to the branch to have a go with your changes. How does that sound?

worksofliam: @Dman247 May I recommend you fork this repo and checkout this feature branch, and PR that?

worksofliam:
This isn't something I have a bunch of time to work on right now. It's not a high-priority issue. I will leave it open in hopes Dylan can provide the further changes as mentioned above, but I will mark it stale in the meantime.

Issue #319, opened 2021-09-23
Provide Object Filter in Object Browser


dlarsencbk:
**Is your feature request related to a problem? Please describe.**
In order to Edit Source files in this extension, each  library and file combination must be entered  individually in the member browser. Our Libraries contain a mix of production objects and source files.  Some libraries have over a hundred source files named by project. This process also requires time to individually identify source files for over 30 libraries and manually add all those libraries and source files to the member browser.  Not all files end in SRC and not all file names start with Q. The resulting work  for this extension would take too much time and is not practical.

**Describe the solution you'd like**
In RDI, Remote Systems View, there is an object filter that can automatically identify and produce a list of just source files. This is extremely helpful to have the system do this automatically
A filter in the Object Browser (or where ever it makes sense) would be an extremely valuable addition to this product.

**Describe alternatives you've considered**
Alternatives include manually adding each library/source file combination individually. Too time consuming. 
Another would be to move the source files to their own library. That involves creating hundreds of other libraries and redesigning the applications. Not practical.

**Additional context**
RDI Object Filter
![image](https://user-images.githubusercontent.com/84473787/134552633-fb5c737b-4e92-4ade-a505-472b4e5cda44.png)

worksofliam:
Hey @dlarsencbk 

Thanks for raising this!

While I really like the idea. I am struggling to think how we would implement it or how it would work from the VS Code side. I am going to add the 'help wanted' tag to see if anyone else would like to chime in.

p.s. I understand adding X source files can be slow. You can also manually edit the JSON for that view to add more.. which might be a tad faster.

Liam

SJLennon:
Bearing in mind that I don't understand much about the capabilities/limitations of VSCODE extensions, here's my take on this one request:

I envisage being able to right click on a library in the member browser, and being presented with a list of SRCPF files in the library. 

From this list you could pick one or more source files to be added to the member browser. 

The current setting could then be saved with an appropriate profile name and thus could be easily reloaded.

It wasn't clear whether the request also included a library level browser.  I suppose such an option could also be done if the above approach for SRCPF was feasible, and you would be able to pick only one library and would then be presented with the list of SRCPFs in it.

But I wonder how many other shops have a similar setup?

worksofliam:
I am still wondering how this would work. I'd like a concrete idea before just slapping something together. Any further input would be great.

worksofliam:
I think I understand this requirement a bit better now. Here's my take:

[...]

PR #366, opened 2021-11-04
POC: Job listener

worksofliam:
I am still debating how worthy this feature is. It is cool, but does it have much use? I am not sure.

Video example: https://www.youtube.com/watch?v=FtgVTZLD3Hc

### Changes

* Add a new feature to listen to a job's job log. This will allow functionality to allow programs to control listening Code for IBM i sessions. Requires db2util.

Todo list:

* [x] Program to open up members in VS Code. This would allow a `WRKMBRPDM` option to open a file up in VS Code.
* [ ] Program to add the source file to member browser. This would allow a `WRKOBJPDM` option to add a source file to the browser
* [ ] Feature to automatically create PDM options for valid options

### Checklist

* [X] have tested my change


SJLennon:
As I recall, you used to be able to open up a WDSC session, or maybe it was Code/400 session, from PDM.  (Wow, I'm dating myself.)
But that was because those products and connections were a bit slow. If you get  "Provide Object Filter in Object Browser" #319 working this probably isn't needed.

edmundreinhardt:
I think this is very interesting.  I like this JobListener approach as it is much better than the RDi approach.
We might be able to use something similar for setting up interactive debug for the VSCode debugger that is coming out soon.

edmundreinhardt:
I imagine we could use this to pop up the Merlin editor as well as the VSCode editor

Wright4i:
@worksofliam I'm on the fence. It's a neat POC and I can see some potential use cases so scrapping the idea entirely feels wrong, but does it meet a high enough bar to take priority over other features?

If you're a person who (in 2022) still lives in PDM but happen to have an active IBM i connection in vscode open then yes this is a time-savings. Personally, I'm more likely to `alt+tab` and `ctrl+shift+p` open the command window without even thinking instead of finding my source in pdm and doing a `V`.

For debugging I think you're the most qualified to judge if it will be useful or help overcome any challenges. 

Code Coverage integration might be a nice addition? Run your test and open code coverage in vscode at the same time?

Non-git Version Control software could also benefit. Anyone who has the ability to run custom commands could have vscode auto-open the source when you checkout a member in your vcs/scm. I could see this being a really nice way to streamline that process for some people.

PR #543, opened 2022-02-16
Make IFS temporary directory configurable solves #498

dferrand:
### Changes

Make IFS temporary directory configurable 

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/Configuration.js
+++ b/src/api/Configuration.js
@@ -39,6 +39,9 @@ module.exports = class Configuration {
     this.tempLibrary = base.tempLibrary || `ILEDITOR`;
 
     /** @type {string} */
+    this.tempDir = base.tempDir || `/tmp/`;
+
+    /** @type {string} */
     this.currentLibrary = base.currentLibrary || ``;
 
     /** @type {string|undefined} */
--- a/src/api/Configuration.js
+++ b/src/api/Configuration.js
@@ -39,7 +39,7 @@ module.exports = class Configuration {
     this.tempLibrary = base.tempLibrary || `ILEDITOR`;
 
     /** @type {string} */
-    this.tempDir = base.tempDir || `/tmp/`;
+    this.tempDir = base.tempDir || `/tmp`;
 
     /** @type {string} */
     this.currentLibrary = base.currentLibrary || ``;
--- a/src/api/IBMi.js
+++ b/src/api/IBMi.js
@@ -4,6 +4,7 @@ const vscode = require(`vscode`);
 const node_ssh = require(`node-ssh`);
 const Configuration = require(`./Configuration`);
 const Tools = require(`./Tools`);
+const path = require(`path`);
 
 let remoteApps = [
   {
@@ -275,9 +276,10 @@ module.exports = class IBMi {
         }
         
         if (tempDirSet) {
-          this.tempPath = this.config.tempDir + `/vscodetemp`;
+          this.tempDir = this.config.tempDir;
         } else {
-          this.tempPath = `/tmp/vscodetemp`;
+          this.tempDir = `/tmp`;
+          await this.config.set(`tempDir`,`/tmp`);
         }
 
         if (tempLibrarySet && this.config.autoClearTempData) {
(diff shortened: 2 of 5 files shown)


src/api/IBMi.js
```diff
@@ -566,7 +599,7 @@ module.exports = class IBMi {
       console.log(`Using existing temp: ` + this.tempRemoteFiles[key]);
       return this.tempRemoteFiles[key];
     } else {
-      let value = `${TEMP_PATH}-` + Tools.makeid();
+      let value = `${this.tempPath}-` + Tools.makeid();
```
worksofliam, line 602: I think that instead of 

```
let value = `${this.tempPath}-` + Tools.makeid();
```

We should use

```
let value = path.posix.join(this.tempPath, `vscodetemp-${Tools.makeid()}`);
```

src/api/IBMi.js
```diff
@@ -247,6 +245,41 @@ module.exports = class IBMi {
...
+            tempDirSet = true;
+          } else {
+            // Directory not created
+          }
+        }
+        
+        if (tempDirSet) {
+          this.tempPath = this.config.tempDir + `/vscodetemp`;
+        } else {
+          this.tempPath = `/tmp/vscodetemp`;
```
worksofliam, line 280: As well as this, we should update the user config with the default path when it failed.

```
await config.set(`tempDir`, `/tmp/vscodetemp`)
```

src/api/IBMi.js
```diff
@@ -270,14 +303,14 @@ module.exports = class IBMi {
             });
 
           this.paseCommand(
-            `rm -f ${TEMP_PATH}*`
+            `rm -f ${this.tempPath}*`
```
worksofliam, line 306: This could be bad since `tempPath` is not a directory. We don't want to delete everything except files we've created.

Could be

```
`rm -f ${path.posix.join(this.tempPath, `vscodetemp*`)}`
```

src/api/Configuration.js
```diff
@@ -38,6 +38,9 @@ module.exports = class Configuration {
     /** @type {string} */
     this.tempLibrary = base.tempLibrary || `ILEDITOR`;
 
+    /** @type {string} */
+    this.tempDir = base.tempDir || `/tmp/`;
```
worksofliam, line 42: Remove ending forward slash: `/tmp`

src/api/IBMi.js
```diff
@@ -247,6 +245,41 @@ module.exports = class IBMi {
...
+          if(typeof result === `object` && !result.code) {
+            // Directory created
+            tempDirSet = true;
+          } else {
+            // Directory not created
+          }
+        }
+        
+        if (tempDirSet) {
+          this.tempPath = this.config.tempDir + `/vscodetemp`;
```
worksofliam, line 278: Remove `/vscodetemp` from this string. Let tempPath be the path to the temporary directory.

worksofliam (changes requested):
A great PR, but requires some changes.

1. `this.tempPath` should be `this.tempDir` and then is the path to the temporary directory. Code for IBM i then creates files that start with `vscodetemp`
2. We should avoid create path strings and instead use `path.posix.join`. This is partly my fault because I was not doing this before.

Let me know what you think. Thanks!

src/api/IBMi.js
dferrand: done

src/api/IBMi.js
dferrand: done

src/api/IBMi.js
dferrand: done

worksofliam:
@dferrand Is this ready for me to review again after this latest commit/

src/api/Configuration.js
dferrand: Done

worksofliam:
@dferrand Just so you know - this is on my to do list. Got some things going on today, but tomorrow should be the day.

worksofliam:
@dferrand This is awesome. There are just some last things that need cleaning up.

### Option 1

Since you're not changing `this.tempDir` when the user updates the settings, maybe I recommend you add `tempDir` to the `restartFields` array at `src/webviews/settings/index.js:46`. This means when they change the `tempDir` value, it will ask the user to restart.

### Option 2 (prefered)

In `src/api/IBMi.js`, remove the `this.tempDir` definition and always refer to `this.config.tempDir` - this means when it's updated in the settings, it will reflect right away. Maybe we still want to add `tempDir` to `restartFields` anyway, but it's optional for sure.

Thanks!!

src/webviews/settings/index.js
```diff
@@ -54,6 +54,11 @@ module.exports = class SettingsUI {
         field.description = `Temporary library. Cannot be QTEMP.`;
         ui.addField(field);
     
+        field = new Field(`input`, `tempDir`, `Temporary IFS directory`);
+        field.default = config.tempDir;
+        field.description = `User must be authorized to create new files in this directory.`;
```
worksofliam, line 59: Change to

> Directory that will be used to write temporary files too. User must be authorized to create new files in this directory.

worksofliam (changes requested):
As noted previously.

dferrand:
I went for option 2.
It's ready for review.

worksofliam:
@dferrand Perfect - thanks! Looking good. Let's get this merged in.

Issue #492, opened 2022-02-01
Open member using STRPCCMD


ngmhun:
**Is your feature request related to a problem? Please describe.**
We have a green-screen source management tool, where we can edit the sources with a PDM-like option. It would be nice it we could define an option code that opens the source for editing in vscode-ibmi. (Could also work from PDM itself.)

**Describe the solution you'd like**
If it was possible from a windows command line to trigger somehow the `IBM i: Open File` command, then it would also be possible to do this from the IBM i server via STRPCCMD, and a PDM-like option could be defined for this.

**Describe alternatives you've considered**
I have tried to run `code -r member:/lib/file/member.ext`, but it did not work.

**Additional context**
Maybe a small js script could also work by connecting to the running vscode instance and triggering this `IBM i:Open File` command. I have tried the member: scheme, because I have seen in the source code, that this was registered with `vscode.workspace.registerfileSystemProvider`, but unfortunately, I am not a JS developer, so I could not figure out how it works.

worksofliam:
Hi @ngmhun Thanks for a cool idea!

Sounds very similar to an idea I had before.

Does your idea look something like this video? https://www.youtube.com/watch?v=FtgVTZLD3Hc&ab_channel=LiamAllan

ngmhun:
Yes, something like that, but maybe connecting to a specific session is not necessary, with the STRPCCMD method, opening a member from any session would work.

worksofliam:
@ngmhun Need some time to think about how this could work. Will get back to you.

pedromelim:
Ya, this would be very cool to be able to use. For companies that already have software management and do have sources spread on so many libraries, making a "shortcut command" to be able to use directly inside an emulator to open that source (library/file) in VS Code, would be great.

ngmhun:
Thank you! I don't know how this communication with an interactive session works. Did you consider listening to a data queue?

worksofliam:
@ngmhun Data queue has and will continue to be considered while I work on this feature.

This will work by the VS Code connection reading the job log and looking for specific commands that start with `c4i`.

dferrand:
I have it working using named pipe.

I can create a PR once #543 is merged.

PR #548, opened 2022-02-19
Add open from IBM i functionality

dferrand:
### Changes

Add open from IBM i functionnality.

Soves #492

### Checklist

* [x] have tested my change


src/api/Configuration.js
```diff
@@ -76,6 +76,9 @@ module.exports = class Configuration {
 
     /** @type {boolean} */
     this.autoSaveBeforeAction = (base.autoSaveBeforeAction === true);
+
+    /** @type {boolean} */
+    this.openFromIbmi = (base.openFromIbmi === true);
```
worksofliam, line 81: New config props also need to be added to the `package.json`.

src/api/IBMi.js
```diff
@@ -510,6 +510,25 @@ module.exports = class IBMi {
           vscode.window.showWarningMessage(`Code for IBM i may not function correctly until your user has a home directory. Please set a home directory using CHGUSRPRF USRPRF(${connectionObject.username.toUpperCase()}) HOMEDIR('/home/${connectionObject.username.toLowerCase()}')`);
         }
 
+
+        if (this.config.openFromIbmi) {
+          const fifo = path.posix.join(this.config.tempDir,`vscodetemp-O__${this.currentUser.trim().toUpperCase()}`);
+          const result = await this.paseCommand(
+            `rm -f ${fifo};mkfifo -m 600 ${fifo};/QOpenSys/usr/bin/setccsid 1208 ${fifo}`,
```
worksofliam, line 517: What is `mkfifo`? Does it exist in the system or does it have to be installed? I usually recommend absolute paths for tools like this.

Also, since this is a brand new server dependancy, you will also need to add `mkfifo` to the `remoteApps` at `src/api/IBMi.js:10` to check, then add it to remote features (`IBMi.js:56`) and then you can make this functionality conditional:

```
if (this.config.openFromIbmi && this.remoteFeatures.mkfifo) {
  // Work here
}
```

src/api/IBMi.js
```diff
@@ -510,6 +510,25 @@ module.exports = class IBMi {
...
+            1            
+          );
+          if (typeof result === `object` && !result.code) {
+            const readFifo = () => {
+              this.client.execCommand(`cat ${fifo}`).then(async result => {
+                await vscode.commands.executeCommand(`code-for-ibmi.openEditable`,result.stdout);
+                readFifo();
+              });
+            }
+            readFifo();
```
worksofliam, line 528: I am not sure how I feel about this loop.

Instead of calling `readFifo` within itself, might I suggest you use `setInterval` instead?

Also, it looks like there needs to be some handling if the file hasn't changed. For example, if `readFifo` is called over and over again and the `stdout` is the same path, won't it just keep opening it over and over again? Please correct me if I am wrong here!

docs/README.md
```diff
@@ -560,6 +560,19 @@ It is safe to have files created by Code for i automatically deleted during main
 If source files are located in a specific ASP, specify here.
 Otherwise, leave blank.
 
+#### Open from IBM i
+
+When enabled, it is possible to open a member or a stream file in Code for IBM i from an IBM i job. Only one instance of Code for IBM i per user profile can do the opening (the first one to connect to the IBM with each user profile).
+To open a member or a stream file, simply write its name into a named pipe called `vscodetemp-O__`*`usrprf`* in the temporary IFS directory (/tmp by default).
```
worksofliam, line 566: Perhaps there also need to be some notes that the file content CCSID needs to be ascii/utf8.

worksofliam (changes requested):
See changes required.

docs/README.md
dferrand: Very good point, I added something. Let me know if it's ok

src/api/Configuration.js
dferrand: Forgot to retrofit this one, done.

src/api/IBMi.js
dferrand: mkfifo is a standard unix/posix tool, it creates a named pipe.

It is part of PASE (it can be shadowed by the GNU coreutils one if you have open source tools installed and your path set but it doesn't matter, they both work). It is very similar to cat, rm, etc... I don't feel we should use an absolute path. I also don't see why we would check its existence, PASE would probably be very broken without it.

src/api/IBMi.js
dferrand: It is called from an async callback therefore it should not fill the stack.

You can see a named pipe as a blocking unstructured data queue without storage. The read is blocking so cat doesn't return until something has written to the named pipe (and closed it). Also the named pipe doesn't store anything so if a path has been read once, it won't be read again (unless it is written again).

It can be strange if you're not familiar with named pipes, let me know if you have other questions.

dferrand:
@worksofliam, it's ready for you to review my answers/code changes

src/api/IBMi.js
```diff
@@ -510,6 +510,25 @@ module.exports = class IBMi {
...
+          const fifo = path.posix.join(this.config.tempDir,`vscodetemp-O__${this.currentUser.trim().toUpperCase()}`);
+          const result = await this.paseCommand(
+            `rm -f ${fifo};mkfifo -m 600 ${fifo};/QOpenSys/usr/bin/setccsid 1208 ${fifo}`,
+            undefined,
+            1            
+          );
+          if (typeof result === `object` && !result.code) {
+            const readFifo = () => {
+              this.client.execCommand(`cat ${fifo}`).then(async result => {
+                await vscode.commands.executeCommand(`code-for-ibmi.openEditable`,result.stdout);
```
worksofliam, line 524: Please add a space between the last comma and `result.stdout`.

E.g. `,result.stdout` becomes `, result.stdout`

worksofliam (changes requested):
This is awesome @dferrand - I have learned something today.

Please see the last sematic change that is required and I can get this merged. After it's been merged, perhaps we could create a program and put it into the docs which could be called from PDM to write the member path to the file (I can do this).

worksofliam:
Much like CL prompting, perhaps we could automatically install this program and set up the PDM option.

From an old PR: https://github.com/halcyon-tech/vscode-ibmi/blob/29a824671b2660d71b1316227eb42f6d19f8e6a9/src/api/JobListener/sources/openmbr.js

src/api/IBMi.js
dferrand: Done

dferrand:
> Much like CL prompting, perhaps we could automatically install this program and set up the PDM option.
> 
> From an old PR: https://github.com/halcyon-tech/vscode-ibmi/blob/29a824671b2660d71b1316227eb42f6d19f8e6a9/src/api/JobListener/sources/openmbr.js

I added the create of the STRC4I command and program (same way as GENCMDXML).

The automatic creation of the PDM option is a bit problematic since the PDM user defined option file (QGPL/QAUOOPT) is customizable per user profile but there is no official way (AFAIK) to retrieve which file is used by a user (it is stored in the user profile object itself). I opted to explain in the documentation how to create the custom option.

worksofliam:
@dferrand - this is still on my to-do list. It's a big feature and want to make sure I give it the time it deserves when I review it.

dferrand:
No worry @worksofliam , take your time.

worksofliam:
Hi @dferrand - this is looking great! Before I merge this though, I will be looking into another VS Code API which may allow us to use a URL along with `STRPCCMD`. See this comment: https://github.com/halcyon-tech/vscode-ibmi/issues/492#issuecomment-1061111991

PR #575, opened 2022-03-04
New option to have source dates appear in gutter

worksofliam:
### Changes

* Removes old methods of viewing source date
* New config to show source dates in gutter

### Checklist

* [x] have tested my change

### DIFF
--- a/package.json
+++ b/package.json
@@ -217,15 +217,10 @@
 								"default": false,
 								"description": "When enabled, Code for IBM i will retain/update source dates of members when editing source members. Requires SQL to be enabled for this feature to work. If this config is changed, you must reconnect to the system."
 							},
-							"sourceDateLocation": {
-								"type": "string",
-								"default": "none",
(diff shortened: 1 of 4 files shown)


cwbuck:
@worksofliam can you tell me the settings I need to adjust in order to enable source dates to show in their normal position as they would in SEU? Just inadvertently pushed some code to production and didn't realize I wiped all the source dates for every line by editing in Code and now I need to fix. 

In the ```settings.json``` I see the ```enableSourceDates``` setting which I assume needs to be ```true```, but what about the ```sourceDateLocation```?

Thanks!

worksofliam:
As long as you have `enableSourceDates`, you should have them enabled. The only other config is `sourceDateGutter`, which is used to add the date gutter.

```
        "enableSourceDates": true,
        "sourceDateGutter": true,
```

worksofliam:
@cwbuck

PR #597, opened 2022-03-24
Allow library list to be delimited by commas and/or spaces(s)

p-behr:
Allow library list to be delimited by commas and/or spaces(s)
Remove duplicate libraries from the library list (keep first occurrence).
Add more specific check of stderr to determine library in error.

### Changes

Description of change here.

### Checklist

* [ ] have tested my change

### DIFF
--- a/src/views/libraryListView.js
+++ b/src/views/libraryListView.js
@@ -28,12 +28,12 @@ module.exports = class libraryListProvider {
       vscode.commands.registerCommand(`code-for-ibmi.changeCurrentLibrary`, async () => {
         const config = instance.getConfig();
         const currentLibrary = config.currentLibrary.toUpperCase();
-  
+
         const newLibrary = await vscode.window.showInputBox({
           prompt: `Changing current library`,
           value: currentLibrary
         });
-  
+
         if (newLibrary && newLibrary !== currentLibrary) {
           await config.set(`currentLibrary`, newLibrary);
         }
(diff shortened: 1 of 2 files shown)


src/views/libraryListView.js
```diff
@@ -257,7 +261,7 @@ module.exports = class libraryListProvider {
       const lines = result.stderr.split(`\n`);
 
       lines.forEach(line => {
-        const badLib = newLibl.find(lib => line.includes(lib));
+        const badLib = newLibl.find(lib => line.includes(`ibrary ${lib}`));
```
worksofliam, line 264: Looks like this needs to be changed back to `lib` and not this invalid string.

worksofliam:
@p-behr Nice PR!! Please see my comment above.

src/views/libraryListView.js
p-behr: Is there something invalid about that string?
Is interpolation not allowed or frowned upon?

There are only a handful of errors that can be generated by ADDLIBLE and they all have that pattern: "ibrary " + `LIB`.  Some error messages have "Library" (capital "L") and some have "library" (lower-case "l").  Since `includes` is case-senstive I left the "L" out of "Library" so it looks a bit goofy; could use regex or convert the stderr line to upper and then compare.  

Just trying to make a more specific test when determining which library an error message is referring to.  For example, if you enter a library list of "CPF, BAD1, BAD2, BAD3" the existing code will only remove CPF from the list since every error message contains that string.  Obviously those other libraries aren't actually getting added to the library list, but they still appear in the UI library list.

worksofliam: @p-behr I am with you now. That makes complete sense. I am going to give this a quick test tonight and then we'll get this merged.

worksofliam:
@p-behr Tested!! About to merge and get this out!

PR #675, opened 2022-05-18
Escape special characters in IFS directory name

chrjorgensen:
### Changes

This PR fixes the problem opening an IFS directory with a quote in the name.

The remote path will have the special characters quote, double-quote, dollar sign, backslash and the space escaped before opening the directory. The list of characters is not complete but can be extended - it only contains the characters that I personally experienced problems with. 

Since the name is now escaped, it is no longer necessary to put it in quotes on the `ls` command.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMiContent.js
+++ b/src/api/IBMiContent.js
@@ -387,8 +387,9 @@ module.exports = class IBMiContent {
    * @return {Promise<{type: "directory"|"streamfile", name: string, path: string}[]>} Resulting list
    */
   async getFileList(remotePath) {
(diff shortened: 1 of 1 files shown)


worksofliam:
@chrjorgensen Neat change! I am going to test this before I merge. I will do that this weekend.

chrjorgensen:
Added another fix to this PR - open symlinks to directories.
Let me know if you want it in a separate PR - but then it will have to wait for this PR to be merged, since it alter the same `ls` command.

chrjorgensen:
@worksofliam Hold your horses... just found an error in my code!

I will close this PR and create another containing only the symlink change.
When the error in the change for special characters in directory names is fixed, I will create a new PR again.

PR #689, opened 2022-06-01
Allow special chars in IFS browser

chrjorgensen:
### Changes

This PR allows handling of special characters in path names in the IFS browser.

The remote path will have the special characters quote, double-quote, dollar sign, backslash and the space escaped before using the path name. The list of characters is not complete but can be extended - it only contains the characters that I personally experienced problems with.

Now you can use all operations in the IFS browser even when the path contains the above mentioned special characters.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMiContent.js
+++ b/src/api/IBMiContent.js
@@ -388,7 +388,7 @@ module.exports = class IBMiContent {
    */
   async getFileList(remotePath) {
     const result = await this.ibmi.sendCommand({
-      command: `ls -a -p -L "${remotePath}"`
+      command: `ls -a -p -L ${Tools.escapePath(remotePath)}`
     });
 
     //@ts-ignore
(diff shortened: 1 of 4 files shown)


worksofliam:
@chrjorgensen Wow, neat!! I am actually going to have to really test this one. Great PR! Let me get back to you.

chrjorgensen:
I just did one test (teased by a colleague) with a directory name with one backslash, e.g. back\slash - and the directory was created as expected. I could also create a file in this directory. All operations on the directory and file works as expected - but...

The file could not be opened - because the unixifyPath() function called by the getFile() function checks for and replaces backslash `'\\'` with slash `'/'` in the remote path!
This is a problem with the node-ssh package, not my code... so I don't think it is a showstopper.

Why it does that is a mystery to me - since a backslash is allowed in a path name on Unix and Linux... :-(
Should be reported to the node-ssh package developers.

worksofliam:
> The file could not be opened - because the unixifyPath() function called by the getFile() function checks for and replaces backslash '\\' with slash '/' in the remote path!

I am guessing this is due to Windows.

PR #708, opened 2022-06-10
Feature/sort ifs shortcuts

chrjorgensen:
### Changes

This PR will add the ability to sort the IFS shortcuts in the IFS browser:

- The shortcuts have a new item type `shortcut` - previously they were `directory` like regular directories.
- There are new icons for the shortcuts to move the shortcut up or down in the list, to the top or bottom and to sort the list.
- There is a new icon in the view to sort the shortcuts alphabetically - case-insensitive.
- There is a new setting in the configuration to automatically sort the shortcuts when a shortcut is added or removed.

It was my ambition to have a drag-and-drop function implemented so one could just reorder by dragging the shortcut into he desired place - but this is currently way above my skills in VS Code extension programming! 😆 

I use the IFS browser almost all the time - with my sources in IFS - so that's why I have focused my contributions to this area of your excellent VS Code extension. My many IFS shortcuts had been added in a long time-frame and were in a chaotic order, and I added this functionality for my own convenience, but it could be useful to others as well. I know I could have manually ordered the shortcut list in the configuration, but this may not be obvious to all...

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/Configuration.js
+++ b/src/api/Configuration.js
@@ -26,7 +26,7 @@ module.exports = class Configuration {
     /** @type {string[]} */
     this.ifsShortcuts = base.ifsShortcuts || [];
 
-    /** @type {boolean} Undefined means not created, so default to on */
+    /** @type {boolean} Default auto sorting of shortcuts to off  */
     this.autoSortIFSShortcuts = (base.autoSortIFSShortcuts === true );
 
     /** @type {string} */
--- a/src/api/Configuration.js
+++ b/src/api/Configuration.js
@@ -26,6 +26,9 @@ module.exports = class Configuration {
     /** @type {string[]} */
     this.ifsShortcuts = base.ifsShortcuts || [];
 
+    /** @type {boolean} Undefined means not created, so default to on */
+    this.autoSortIFSShortcuts = (base.autoSortIFSShortcuts === true );
+
     /** @type {string} */
     this.homeDirectory = base.homeDirectory || `.`;
 
--- a/src/views/ifsBrowser.js
+++ b/src/views/ifsBrowser.js
@@ -193,6 +193,66 @@ module.exports = class ifsBrowserProvider {
         }
       }),
 
+      vscode.commands.registerCommand(`code-for-ibmi.moveIFSShortcutToTop`, async (node) => {
+        const config = instance.getConfig();
+
+        let moveDir;
+
+        let shortcuts = config.ifsShortcuts;
+
+        if (node) {
+          moveDir = node.path;
+        }
+
+        if (moveDir) {
+
+          try {
+            moveDir = moveDir.trim();
+
+            const inx = shortcuts.indexOf(moveDir);
+
+            if (inx >= 1 && inx < shortcuts.length) {
+              shortcuts.splice(inx, 1);
+              shortcuts.splice(0, 0, moveDir);
+              await config.set(`ifsShortcuts`, shortcuts);
+              if (Configuration.get(`autoRefresh`)) this.refresh();
+            }
+          } catch (e) {
+            console.log(e);
+          }
+        }
+      }),
+
+      vscode.commands.registerCommand(`code-for-ibmi.moveIFSShortcutToBottom`, async (node) => {
+        const config = instance.getConfig();
+
+        let moveDir;
+
+        let shortcuts = config.ifsShortcuts;
+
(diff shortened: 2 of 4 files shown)


worksofliam:
Wow, this is some excellent work. Expect a review over the weekend!

> It was my ambition to have a drag-and-drop function implemented so one could just reorder by dragging the shortcut into he desired place - but this is currently way above my skills in VS Code extension programming! 

I would love this also. This is actually not possible with the VS Code API at the moment... so it's not possible no matter how bad we want that!

src/views/ifsBrowser.js
```diff
@@ -110,14 +112,155 @@ module.exports = class ifsBrowserProvider {
...
+          if (Configuration.get(`autoRefresh`)) this.refresh();
+        } catch (e) {
+          console.log(e);
+        }
+      }),
+
+      vscode.commands.registerCommand(`code-for-ibmi.moveIFSShortcutDown`, async (node) => {
+        const config = instance.getConfig();
+
+        let moveDir;
```
worksofliam, line 139: I think this can be re-written as 

```
if (node) {
  const moveDir = node.path ? node.path.trim() || null;
  
  if (moveDir) {
    // logic
  }
}

File: src/api/Configuration.js [hunk 2]
```diff
@@ -26,6 +26,9 @@ module.exports = class Configuration {
     /** @type {string[]} */
     this.ifsShortcuts = base.ifsShortcuts || [];
 
+    /** @type {boolean} Undefined means not created, so default to on */
+    this.autoSortIFSShortcuts = (base.autoSortIFSShortcuts === true );
```
worksofliam, line 30: I see you've added the config here, which is correct. But, you've also got to add it to the `package.json`. 

It belongs in `configuration` -> `code-for-ibmi.connectionSettings`. Look for `autoConvertIFSccsid` as an example.

src/views/ifsBrowser.js
worksofliam: I see you've followed this format in a few commands, too.

worksofliam (changes requested):
It's very good!! Seriously, a great PR. Please have a review of my comments.

Also, for the right click options, I don't think 

* move up
* move down
* sort all

belong as right-click options on the `shortcut`. This is because they're already accessible from the UI and I feel as if it adds some clutter. I do like the move to top and bottom, though.

image

chrjorgensen:
Thanks for the feedback - and fair comments.
I'll have a look at it asap - hopefully this weekend...

src/views/ifsBrowser.js
chrjorgensen: Fixed in commit [54f0421](https://github.com/halcyon-tech/vscode-ibmi/pull/708/commits/54f0421c703896eba43012c234627af261ded7f9)

src/api/Configuration.js
chrjorgensen: Fixed in commit [78bd326](https://github.com/halcyon-tech/vscode-ibmi/pull/708/commits/78bd326a9dab8e768172cba5573475cf21254507)

chrjorgensen:
I agree the right-click options for moving a shortcut up and down and sorting the shortcuts may be redundant to the UI, and I have removed them now. I added them because the icons for moving a shortcut up and down stays on the display if the shortcut is clicked, and I don't know if I like that... 😄

chrjorgensen:
About the drag-and-drop - are you sure it is not possible with the current VS Code API?

It is my understanding, that it is possible (see this [example](https://github.com/microsoft/vscode-extension-samples/blob/799d4ab2499ff4d70361e77a1bfde4388114a79b/tree-view-sample/src/testViewDragAndDrop.ts)) but it requires the use of `createTreeView` and coding of a `dragAndDropController` instead of the simple `registerTreeDataProvider` - and that is a change way too big for me to code...!

worksofliam:
@chrjorgensen Interesting. As far as I'm aware it wasn't and I think it would require implementing our own provider. I will ask some of the other extension authors I know and see what they think.

worksofliam:
@chrjorgensen Your changes look excellent! Thanks. If you are happy with everything, let me know and we can get this merged.

worksofliam (approved):

chrjorgensen:
I'm happy - go ahead... 👍

PR #716, opened 2022-06-18
Data collection support

worksofliam:
### Changes

Support for collecting non-PPI data. I am also happy if this does not get merged, but it would be interesting to see what data it can provide.

#### Data collection

We do not collect any Personally identifiable information (PII). We do collect:

* When you connect to a system - but nothing about that system.
* The version of Visual Studio Code you're using.
* The version of the Code for IBM i you're using.
* The architecture of the machine you're using (e.g. Windows, Linux, Mac, Cloud, etc)

We do this for a few reasons:

1. Curious to see how many active users we have.
2. Allows us to see how often users are updating.
3. Let's us know what operating systems we should cater to (e.g. Windows vs Mac)

**You can disable this:** 

1. Open the VS Code settings (Control / Command + Comma)
2. Search `telemetry` and set `telemetry.telemetryLevel` to `none`.

### Checklist

* [x] have tested my change

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,3 +31,22 @@ Maintain your RPGLE, CL, COBOL, C/CPP on IBM i right from Visual Studio Code. Ed
 3. `npm i`
 4. 'Run Extension' from vscode debug.
 
+### Data collection
+
+We do not collect any Personally identifiable information (PII). We do collect:
+
+* When you connect to a system - but nothing about that system.
+* The version of Visual Studio Code you're using.
+* The version of the Code for IBM i you're using.
+* The architecture of the machine you're using (e.g. Windows, Linux, Mac, Cloud, etc)
+
+We do this for a few reasons:
+
+1. Curious to see how many active users we have.
+2. Allows us to see how often users are updating.
+3. Let's us know what operating systems we should cater to (e.g. Windows vs Mac)
+
+**You can disable this:** 
+
+1. Open the VS Code settings (Control / Command + Comma)
+2. Search `telemetry` and set `telemetry.telemetryLevel` to `none`.
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,15 +1,16 @@
 {
 	"name": "code-for-ibmi",
-	"version": "1.3.4",
+	"version": "1.3.7",
 	"lockfileVersion": 2,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "code-for-ibmi",
-			"version": "1.3.4",
+			"version": "1.3.7",
 			"license": "MIT",
 			"dependencies": {
 				"@bendera/vscode-webview-elements": "^0.6.3",
+				"@vscode/extension-telemetry": "^0.6.0",
 				"csv": "^6.0.5",
 				"escape-string-regexp": "^5.0.0",
 				"ignore": "^5.1.9",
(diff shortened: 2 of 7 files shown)


worksofliam:
@chrjorgensen Before I merge this, I want your opinion. I know you're an avid Codium user and I want to be respectful. Please check out the readme and let me know what you think.

https://github.com/halcyon-tech/vscode-ibmi/pull/716/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R34

I am also happy if this does not get merged, but it would be interesting to see what data it can provide.

chrjorgensen:
Hi @worksofliam 

Thank you for being respectful and asking for my opinion! 👍 

Oh boy - this is a dangerous topic and I could go on with a loooong rant about how privacy is eroding in the modern world.
Suffice it is to say that I value my privacy and do everything I can to protect it - the exact reason why I switched from VS Code (which sends data to Micro$oft) to Codium (which doesn't), from Chrome to Firefox, from Windows to Linux... well, you get it. Today it is getting more and more difficult to not being tracked, whether it is browsing the internet (Facebook is tracking all browsing, even for those not logged in or even being a Facebook member!) or using a piece of software (Windows 10/11 sends data to Micro$oft, unless you stop it - e.g. by using [O&O ShutUp10++](https://www.oo-software.com/en/shutup10)). The world is on a slide and everyone is obsessed by data - why is that?

- Because data is big business! The Big Tech companies all make billions of dollars on data generated by you and me - which should be our data to control. Heck, even the car manufacturers now collect data on our use of the cars that they built - that we bought and own which makes it our data. **We have become a product - do we really want that, just to have some "free" services?**
- Because data is about control! Knowing the actions and thoughts of people make it easier to influence / manipulate them in the wanted direction. This is done today by companies (to increase sales) and by governments (to control society). Look at Russia and China (and more countries) for bad examples of this. Do we want to support this? It can be a good thing - but it can also be bad, and **we can't stop it from being used in a bad way once the data is in the wild.**

I disagree with the current opinion that data collection is a right, that can be disabled for the few that protests - it should be the other way around, privacy should be a right by default. Think about how much data is generated by you and collected by others in your daily life - to what benefit for you? None - it's the data collectors that benefit. You are a product, which is sold to anyone interested in information about you - and this has become the norm. Everything and everyone is being monitored today - is this the society that we really want? Most don't bother to think about it - and society loses (if you believe monitoring / surveillance / control is a bad thing... 😃)

...and I better stop myself now - this is already getting way longer than it was meant to be... I'll just throw some questions up in the air:

- Why is it okay for software to send data about the use of a software product?
- Why is it up to the user to opt-out of any data collection - instead of the user actively agreeing by opt-in?
- How do you trust the software to not collecting more data after data collection has started?
- How do you protect your data after they have been sent (answer: you can't)?
- How do you control who accesses your data after they have been sent (answer: you can't)?
- How do you withdraw your data after they have been sent (answer: you can't)?

It's very difficult to make this short - it is such a big topic... so back to your question:

I'm against adding telemetry / data collection to your excellent piece of software. Don't get me wrong - I understand your desire to get to know your audience by having some data about them and their use of your software. And telemetry collection may seem innocent - but this is just another small step in the wrong direction that the world is headed right now. Could it be done differently, e.g. have a window at installation asking if the user wants to send these once? Could some error reporting be added, where the user is asked to send error data?

May I suggest a few things to make it nicer for the users who don't care as much about privacy as I do?

1. Make it an opt-in - let the user actively agree and select it - instead of opt-out.
2. Make it more obvious - put a window on screen before installation, that this is going to happen, so the user can back out.
3. Don't have it at the bottom of the instructions - no one reads the complete license or instructions before starting to use the software.

@worksofliam I have tremendous respect for your work and this great piece of software, and I want to help develop it and promote it. But I'm sad to see the (IMHO) bad trends of the modern world creep into it, and even if you would follow my suggestions above, I can't guarantee that I will continue using it: Trust is a difficult thing - It's hard to build and can be lost very quickly. And I have less trust in software that wants to collect data about me or my use of the software and more trust in software without it or software which asks **openly** for my contribution (error report, install info) and give me the option to decline - for the above reasons and thoughts...


Well, that was my opinion - or more of a rant, if you like. Hope you got to the end... 😆

worksofliam:
@chrjorgensen thanks for all the info. First off, I really don't want to make you or any of our users not trust the software - or me for that matter. 

You are right about opt-in by default, sadly VS Code by default is not like that, which sucks.

Window at the bottom is likely the best approach here, but you are right: when I see that I always decline tracking when I can. I wouldn't want to impose that on our users.

I really appreciate your reply and I hope you don't mind if I pass it around to others. This has been another pleasure of open-source for me. I tried something, got an opinion and made a new decision.

chrjorgensen:
@worksofliam Thank you for your kind response - and I'm glad you got something useful out of it. Feel free to pass it around - I stand by my opinion even though many probably disagree... 💣

PR #742, opened 2022-07-11
Local workspace improvements

worksofliam:
### Changes

* Better EVENTF support for local files that use the ibmi-bob naming format.
* Remove `deployFirst` option for interactive Action editor.
* Ability to download files as part of local action execution
* New messages show when a connection is made with a single workspace open.

### DIFF
--- a/global.d.ts
+++ b/global.d.ts
@@ -1,0 +1,10 @@
+
+interface Action {
+  name: string;
+  command: string;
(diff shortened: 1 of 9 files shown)


edmundreinhardt:
Liam,  I just wanted to make sure that you have it as PGM.RPGLE instead of RPGLE.PGM

worksofliam:
@edmundreinhardt that is correct!

PR #760, opened 2022-07-20
Contributing actions API

### LINKED ISSUE #755: Extensions can contribute their own Actions
It'd be cool if we provided a command so other extensions can provide Actions:

```
vscode.commands.executeCommand(`code-for-ibmi.contributeActions`, [
   {...} // Action model
])
```

worksofliam:
Signed-off-by: Liam Barry Allan <mrliamallan@live.co.uk>

### Changes

Allow other extensions to register actions. **Needs testing.**

Closes #755.

### Test list

* [ ] Local actions continue to work (validator changed)
* [ ] Command to add actions

### Checklist

* [ ] have tested my change


priceaj:
Speaking of actions, I noticed the other day that when you define a user action it copies the entire default set into the user config, rather than just adding the user defined one.  

Is it worth amending that behaviour so the user actions add to the defaults (just in case new default ones are added)?

worksofliam:
@priceaj There was a reason we update the global config. It was to do with when the user changes a default and how the config stores it. I personally think actions need an overhaul to be connection specific too.

Lots of room for improvement here.

PR #778, opened 2022-07-25
Action management improvements

chrjorgensen:
### Changes

This PR has the following improvements to action management:

- Duplicated actions are not saved before confirmation. The action selected for duplication will be shown in the action edit panel, where the user can change and save the action.
- The title for the action panel is changed according to the operation, e.g. `Change action "xxxx"` when changing an action. Previously the title was always `Work with Actions`.
- A button labeled `Cancel` is added to all action management panels. This is more intuitive for cancelling the operation and go back to the list of actions than having to close the panel using Ctrl-W or click on X.
- The action edit panel will stay on screen if the user clicks `Delete` and then selects No or cancel on the confirmation message. However, the changed values will be reset to the initial values, I have not yet figured out how to keep the changed values.
- Show Delete confirmation message as modal to bring full attention to the user.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/webviews/actions/index.js
+++ b/src/webviews/actions/index.js
@@ -292,7 +292,7 @@ module.exports = class SettingsUI {
       if (data) {
         switch (data.buttons) {
         case `deleteAction`:
-          const result = await vscode.window.showInformationMessage(`Are you sure you want to delete this action?`, `Yes`, `No`)
+          const result = await vscode.window.showInformationMessage(`Are you sure you want to delete this action?`, { modal:true }, `Yes`, `No`)
           if (result === `Yes`) {
             allActions.splice(id, 1);
             await Configuration.setGlobal(`actions`, allActions);
--- a/src/webviews/actions/index.js
+++ b/src/webviews/actions/index.js
@@ -133,6 +133,7 @@ module.exports = class SettingsUI {
     let allActions = Configuration.get(`actions`);
     let currentAction;
     let uiTitle;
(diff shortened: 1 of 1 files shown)


src/webviews/actions/index.js
```diff
@@ -273,46 +279,63 @@ module.exports = class SettingsUI {
...
     ui.addField(field);
 
-    let {panel, data} = await ui.loadPage(`Work with Actions`);
+    while (stayOnPanel === true) {
+      let {panel, data} = await ui.loadPage(uiTitle);
+
+      if (data) {
+        switch (data.buttons) {
+        case `deleteAction`:
+          const result = await vscode.window.showInformationMessage(`Are you sure you want to delete this action?`, `Yes`, `No`)
```
worksofliam, line 295: Perhaps this should be a modal instead so it has the full attention from the user?

```js
await vscode.window.showInformationMessage(`Are you sure you want to delete this action?`, { modal: true }, `Yes`, `No`)
```

worksofliam (approved):
Tested and am happy with this! I love UX changes. They are my favourite. I did leave one comment, but I am happy if it doesn't change. Let me know when you are happy and I will merge.

src/webviews/actions/index.js
chrjorgensen: I actually had this in the code for a while, but I removed it again when I was trying to keep the changed values when clicking Delete. 😄 
I will add it again...

PR #780, opened 2022-07-26
Fix member operations

chrjorgensen:
### Changes

This PR will
- Enhance the member name validation when creating a member - like the validation when copying a member. This will also allow creating a member without extension.
- Allow copy of an empty member without showing error. Previously this generated an error even though the member was created.
- Correctly copy to a member without extension. Previously the new member had the extension from the old member.
- Add a prompt for confirmation when copying to an existing member.
- Correctly rename a member to member name without extension. Previously the extension was not removed.
- Keep the user input on screen until the create, copy or rename member operation has completed. This allows the user to edit the member name to fix the situation.

### Checklist

* [x] have tested my change


worksofliam (changes requested):
The code looks good, but it looks like the member path validator needs to better support not requiring an extension.

I tried copying and creating members with an extension but looks like this error is being thrown:

image

chrjorgensen:
@worksofliam I did not change the member name validation code - but I found that it requires a `.` after the name to accept it as a member without type. And I actually quite like that, since it forces the user to specify the member name in a special format to create it without type.

WDYT?

But I'm not a fan of the way it reports an error in the name specified - IMO it should keep the window and member name when reporting the name error, for the user to edit the name or press `Esc` to cancel the operation. If you agree, I will code this in the create and copy member functions.

chrjorgensen:
New commit added - to keep the user input on screen until the create or copy member operation has completed.

chrjorgensen:
Another commit - keep user input until member is renamed.

chrjorgensen:
Just found an error - copying a member to itself by only changing the type was not caught (but failed of course).
Check added in commit [d858386](https://github.com/halcyon-tech/vscode-ibmi/pull/780/commits/d858386efed9e57b8cd6898c91e263d9558c23cf)

worksofliam:
@chrjorgensen Hi! Just want to let you know that I will get around to this after tomorrow.

worksofliam (changes requested):
> This will also allow creating a member without extension.

I am not able to create a member without an extension. This is likely due to the member path parser.

image

> Correctly copy to a member without extension. Previously the new member had the extension from the old member.

When I attempt to copy a member with no extension, I get this error

image

> Keep the user input on screen until the create, copy or rename member operation has completed. This allows the user to edit the member name to fix the situation.

This I like a lot! Great change.

chrjorgensen:
@worksofliam For creating a member without extension, [see my comment above.](https://github.com/halcyon-tech/vscode-ibmi/pull/780#issuecomment-1197808101)

There were an error when the new name was not in the correct format - this has been fixed. If you still encounter error copying a member with no extension, please specify the steps to reproduce the error. 😉

chrjorgensen:
@worksofliam Btw, great live PR session today! 👌

worksofliam:
@chrjorgensen Will test this tomorrow as requires a little bit of time to test all the items.

worksofliam (approved):
Sorry it's taken me a while to get around to this. It's looking great!

- 👍 allow creating a member without extension.
- 👍 Allow copy of an empty member without showing error.
- 👍 Correctly copy to a member without extension
- 👍 Add a prompt for confirmation when copying to an existing member.
- 👍 Correctly rename a member to member name without extension.
- 👍 Keep the user input on screen until the create, copy or rename member operation has completed.

PR #798, opened 2022-08-07
TypeScript conversion

worksofliam:
### Changes

This is almost a complete re-write. I did not want to just add the TS compiler to the base and support both JS and TS. Instead, I am copying over one file at a time and re-writing into TS. This will allow for improvements, but also likely introduce bugs which is a risk we're going to have to take for this. Conversion first, cleanup and testing after.

* Re-written how `instance.ts` works. Instead of using mostly static export functions, it's actually using OO correctly this time.
* Making serious use of `.d.ts`. Check out `index.d.ts`
* I am doing all the extension APIs first (`api` directory) and then worrying about the VS Code API stuff later.
* Still lots to do.


Wright4i:
Wow! Great initiative - definitely happy to test this version when you're ready

worksofliam:
@Wright4i It's been a long time coming. It might be weeks honestly, it's quite a drag and it's mostly the type of thing where I need to be heads down for hours at a time.

Gotta try and fit in in with everything else, of course.

worksofliam:
Closing in favour of #852

worksofliam:
Closed in place of #852

PR #818, opened 2022-08-17
Support QsysFs INB files for IBM i Notebooks

Wright4i:
### Changes

Added support for saving source type `INB` files in Qsys. Before it would attempt to save the JSON from IBM i Notebooks up to the record length, causing malformed JSON and a loss of all data stored in your INB file.

**Writing to Qsys**
If extension is `INB` and the sourceData array has 1 element of valid JSON then it will use regex to split on recordLength instead of `\n`.

**Reading from Qsys**
If extension is `INB` join srcdta using ` `` ` instead of `\n` back to the body

### Checklist

* [x] have tested my change
* [x] updated relevant documentation [vscode-ibmi-notebooks commit/future PR](https://github.com/halcyon-tech/vscode-ibmi-notebooks/commit/c934df3a403b58dd4e1eaf4c5fcbce54b9c6acb0)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,3 +35,4 @@ Thanks so much to everyone [who has contributed](https://github.com/halcyon-tech
 * [@EddieSmith](https://github.com/EddieSmith)
 * [@fathert](https://github.com/fathert)
 * [@Teqed](https://github.com/Teqed)
+* [@Wright4i](https://github.com/Wright4i)
--- a/src/filesystems/qsys/complex.js
+++ b/src/filesystems/qsys/complex.js
@@ -18,9 +18,9 @@ module.exports = class ComplexQsysFs {
    */
   async readFile(uri) {
     const connection = instance.getConnection();
-    const {asp, library, file, member} = connection.parserMemberPath(uri.path);
+    const {asp, library, file, member, extension} = connection.parserMemberPath(uri.path);
 
-    const memberContent = await contentApi.downloadMemberContentWithDates(asp, library, file, member);
+    const memberContent = await contentApi.downloadMemberContentWithDates(asp, library, file, member, extension);
 
     return new Uint8Array(Buffer.from(memberContent, `utf8`));
   }
@@ -40,9 +40,9 @@ module.exports = class ComplexQsysFs {
    */
   writeFile(uri, content, options) {
     const connection = instance.getConnection();
-    const {asp, library, file, member} = connection.parserMemberPath(uri.path);
+    const {asp, library, file, member, extension} = connection.parserMemberPath(uri.path);
 
-    return contentApi.uploadMemberContentWithDates(asp, library, file, member, content.toString(`utf8`));
+    return contentApi.uploadMemberContentWithDates(asp, library, file, member, extension, content.toString(`utf8`));
   }
 
   /**
--- a/src/filesystems/qsys/complex/content.js
+++ b/src/filesystems/qsys/complex/content.js
@@ -19,14 +19,16 @@ module.exports = class IBMiContent {
    * @param {string} lib 
    * @param {string} spf 
    * @param {string} mbr 
+   * @param {string} ext
    */
-  static async downloadMemberContentWithDates(asp, lib, spf, mbr) {
+  static async downloadMemberContentWithDates(asp, lib, spf, mbr, ext) {
     const connection = instance.getConnection();
     const content = instance.getContent();
 
     lib = lib.toUpperCase();
     spf = spf.toUpperCase();
     mbr = mbr.toUpperCase();
+    ext = ext.toUpperCase();
 
     const tempLib = connection.config.tempLibrary;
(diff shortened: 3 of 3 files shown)


Wright4i:
@worksofliam the reason I marked this as a draft - this works when using qsys/complex, but nothing I can do for qsys/basic. In essence, you need to have `config.enableSourceDates` turned on.

Should I make a new config option and let the code drop into using qsys/complex on `config.enableSourceDates || config.enableINBSupport` or just update the description for Source Dates to mention IBM i Notebooks?

Or perhaps a different idea?

Wright4i:
Here's a test showing INB loading correctly from Qsys (left) and the JSON that was split at 100 (right).
qsys_inb_test

worksofliam:
@Wright4i Neat! I've got to be careful with this, since it's affecting quite a serious bit of code. It makes sense that you're changing the lines to not go over the record lengths, but I imagine there is some impact on how this affects source dates.

> Should I make a new config option and let the code drop into using qsys/complex

Let's not do this. I think something that needs to be investigated before we look into this is: changing the document content (the JSON) before it even touches `writeFile` or `readFile`. 

I am not sure how I feel about checking for a specific extension in those methods either. Seems kind of hacky and would like to investigate if this is something that we can achieve alone in vscode-ibmi-notebooks before touching this filesystem implementation.

Let me know what you think.

Thanks!

Wright4i:
@worksofliam the extension check was really to make sure this change doesn't affect anything else as I was also concerned about an impact on source dates and non-INB files. There's no reason this wouldn't work without the extension check and then any valid 1 line JSON document would save properly to Qsys.

I'll see what it would take to modify the JSON before the writeFile. This might be the **better solution** I'll report back.

I started here as this was the exact spot in the code making malformed JSON today. I made 2 notebooks and lost everything _then_ found the documentation mentioning don't use source members 😅. 

This bit of code I especially don't like in the current content.js. Just quietly truncating any source you have over the record length. It could use some improvement as well.
```sqlrpgle
      if (sourceData[i].length > recordLength) {
        sourceData[i] = sourceData[i].substring(0, recordLength);
      }
```

Wright4i:
@worksofliam so following up on modifying the JSON before the writeFile.

To do that the change would need to be back in `vscode-ibmi-notebooks` in the `serializeNotebook` function.

There's a few issues though with modifying the serialize function I ran into:

1.  No access to recordLength
2.  Return type expected is `<Uint8Array>` which I can't seem to change as `vscode.workspace.registerNotebookSerializer` needs to process the content in that type
3. JSON doesn't support multi-line values. You can add in `\n` newline characters, but the value itself will still be one long string
4. No way to detect we're trying to save to Qsys

So with those issues this is the best I could come up with after trying a half dozen ideas and it definitely seems worse than modifying writeFile. This is just a POC change, obviously it'd have to be conditioned not to always split (see Issue 4 above)...

```ts
  async serializeNotebook(
    data: vscode.NotebookData,
    _token: vscode.CancellationToken
  ): Promise<Uint8Array> {
    let contents: RawNotebookCell[] = [];

    for (const cell of data.cells) {
      let splitCell = cell.value.match(/.{1,87}/g);
      if (splitCell === null) {
        splitCell = [cell.value];
      }

      for (const split of splitCell) {
        contents.push({
          kind: cell.kind,
          language: cell.languageId,
          value: split
        });
      }
    }

    return new TextEncoder().encode(JSON.stringify(contents,null,1));
  }
```

So I'm splitting the cell pushing a new object everytime a cell.value is over the recordLength (112). I also need to stringify with 1 whitespace to at least make the individual JSON key-value pairs on new lines before it gets encoded.

The result looks like this:
image

It's better than nothing.. but it still is 💩

# Long story short
I'm open to any ideas on how to abstract it / make it less hacky, but I think the change will have to be somewhere in this repo.

-or-

Maybe we give up and instead make a QOL issue ticket to just block Notebooks from even trying to open from a Qsys .inb file and instead suggest an IFS/Local notebook. Then nobody else can fall into the same trap.

worksofliam:
Yeah, this is tricky. IMO, I personally am ok with not supporting QSYS more than I am ok with updating our filesystem provider to support it. I think we are better off blocking notebooks for that filesystem.

> No way to detect we're trying to save to Qsys

Yes, there is a way but I have no idea if you can access it from the notebook.

`document.uri.scheme === 'member'` where `document` is a [`TextDocument`](https://code.visualstudio.com/api/references/vscode-api#TextDocument).

worksofliam:
@Wright4i Ok, here's an idea but think I found an issue in the VS Code API.

There is a `onWillSaveTextDocument` event as part of [`workspace`](https://code.visualstudio.com/api/references/vscode-api#workspace), but not a `onWillSaveNotebookDocument`. This is the event that might allow us to check the scheme, and if it is `member`, then reformat the JSON before writing to the filesystem.

worksofliam:
Discussed with @Wright4i what the next steps are. Will close.

PR #832, opened 2022-08-24
Add a disconnect button to status bar

sebCIL:
### Changes

Add a disconnect button to status bar.
2022-08-24_12h13_05

### Checklist

* [X] have tested my change

### DIFF
--- a/package.json
+++ b/package.json
@@ -373,10 +373,10 @@
 					"default": true,
 					"description": "When enabled, files will automatically be opened when copied or moved/renamed."
 				},
-				"code-for-ibmi.showReconnectButton": {
+				"code-for-ibmi.showConnectionButtons": {
 					"type": "boolean",
 					"default": false,
-					"description": "If enabled, a new button will be shown in the toolbar to force reconnect to the IBM i. Requires restart if changed."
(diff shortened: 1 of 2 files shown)


worksofliam (changes requested):
A very good change, but I don't want this button shown by default. In this same file we also add the reconnect button, but it hidden by default. Here's what I think we should:

* Rename config `showReconnectButton` to `showConnectionButtons`
* Like the reconnect button, only show the disconnect button when `showConnectionButtons` is enabled

sebCIL:
Ok, it only show the disconnect button when `showConnectionButtons` is enabled.
I have made the modification in `package.json` too but how to convert `showReconnectButton` to `showConnectionButtons` in the configuration file?

worksofliam:
@sebCIL Awesome!

> I have made the modification in package.json too but how to convert showReconnectButton to showConnectionButtons in the configuration file?

There is no way to convert. `showConnectionButtons` will be treated as a new property and the old one will be ignored in the future.

I will review today!

worksofliam (approved):
Testing both enabled and disabled. Looks great! Thank you!

PR #844, opened 2022-08-30
Select from list of previous current libraries

chrjorgensen:
### Changes

This PR will present the previous current libraries in a list and allow filtering and selection of the libraries in addition to changing current library to library not previously selected:

- Any new value for library name is validated before change.
- Selecting a library from the list will remove the library from the list, push the current library to the list and change the current library to the selected one.
- The final item on the list is `Clear list`.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/Configuration.js
+++ b/src/api/Configuration.js
@@ -79,6 +79,9 @@ module.exports = class Configuration {
 
     /** @type {boolean} */
     this.autoSaveBeforeAction = (base.autoSaveBeforeAction === true);
+
+    /** @type {string[]} */
+    this.prevCurLibs = base.prevCurLibs || [];
   }
 
   /**
--- a/src/api/Configuration.js
+++ b/src/api/Configuration.js
@@ -79,9 +79,6 @@ module.exports = class Configuration {
 
     /** @type {boolean} */
     this.autoSaveBeforeAction = (base.autoSaveBeforeAction === true);
-
-    /** @type {string[]} */
-    this.prevCurLibs = base.prevCurLibs || [];
   }
 
   /**
--- a/src/views/libraryListView.js
+++ b/src/views/libraryListView.js
@@ -3,8 +3,6 @@ const vscode = require(`vscode`);
 
 let instance = require(`../Instance`);
 const Configuration = require(`../api/Configuration`);
-const { list } = require("../../dist/extension");
-const { getConnection } = require("../Instance");
 
 module.exports = class libraryListProvider {
   /**
--- a/src/views/libraryListView.js
+++ b/src/views/libraryListView.js
@@ -27,7 +27,7 @@ module.exports = class libraryListProvider {
         const connection = instance.getConnection();
         const config = instance.getConfig();
         const currentLibrary = config.currentLibrary.toUpperCase();
-        let prevCurLibs = config.prevCurLibs;
+        let prevCurLibs = Object.values(instance.getStorage().get(`prevCurLibs`));
         let list = [...prevCurLibs];
 
         if (list.length > 0) {
@@ -50,7 +50,7 @@ module.exports = class libraryListProvider {
           const newLibrary = quickPick.selectedItems[0].label;
           if (newLibrary) {
             if (newLibrary === `Clear list`) {
-              await config.set(`prevCurLibs`, []);
+              await instance.getStorage().set(`prevCurLibs`, {});
               list = [];
               quickPick.items = list.map(lib => ({ label: lib }));
               vscode.window.showInformationMessage(`Cleared list.`);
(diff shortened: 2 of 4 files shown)


src/api/Configuration.js
```diff
@@ -79,6 +79,9 @@ module.exports = class Configuration {
 
     /** @type {boolean} */
     this.autoSaveBeforeAction = (base.autoSaveBeforeAction === true);
+
+    /** @type {string[]} */
+    this.prevCurLibs = base.prevCurLibs || [];
```
worksofliam, line 84: I don't think this config should be saved into the connection settings. I think it is more similar to how we store previously opened files. See `instance.storage.get('sourceList')` as an example and let's use this instead for storing previous current libraries.

src/views/libraryListView.js
```diff
@@ -3,6 +3,8 @@ const vscode = require(`vscode`);
 
 let instance = require(`../Instance`);
 const Configuration = require(`../api/Configuration`);
+const { list } = require("../../dist/extension");
```
worksofliam, line 6: Guessing this should not be here since we don't want to pull in from `dist` 😄

src/views/libraryListView.js
```diff
@@ -24,18 +26,62 @@ module.exports = class libraryListProvider {
...
+            if (newLibrary === `Clear list`) {
+              await config.set(`prevCurLibs`, []);
+              list = [];
+              quickPick.items = list.map(lib => ({ label: lib }));
+              vscode.window.showInformationMessage(`Cleared list.`);
+            } else {
+              if (newLibrary !== currentLibrary) {
+                let newLibraryOK = true;
+                try {
+                  await connection.remoteCommand(`CHGCURLIB ${newLibrary}`);
```
worksofliam, line 63: I am not sure this is required. We have an internal API in this view `validateLibraryList` which should achieve the same purpose.

```js
const badLibs = await this.validateLibraryList([newLibrary]);
const isBad = badLibs.includes(newLibrary); // went wrong
```

worksofliam (changes requested):
This is a great UX addition. I have requested some changes to the config and validation, but other than that.. love it!

src/views/libraryListView.js
chrjorgensen: I don't know how line 6 and 7 ended up being added to this file - it wasn't me! 😭 😆 
Perhaps VS Code auto-added them because of some reference during development?

They will of course be removed!

src/views/libraryListView.js
chrjorgensen: Yeah, saw that - but the API doesn't return the cause of any problem discovered, i.e. why the library can't be used as current library. So the user won't know what caused the rejection - invalid name, non-existing library, authority, using `QTEMP` as current library, whatever the reason `CHGCURLIB` wouldn't work.

My solution presents the error message from IBM i - although it could be improved by removing the CPFxxxx in the beginning of the error message string.

IMHO some API's for validating libraries, objects and members could provide better feedback to the user - and code. But that is outside this scope...

src/api/Configuration.js
chrjorgensen: I agree, that was also my first thought - but couldn't make it work. Will try again...

chrjorgensen:
@worksofliam I have made some changes - please review. 😃 

Note that I have not (yet) changed the `CHGCURLIB` validation to use the `validateLibraryList` API - see my comments above.

To use separators in the QuickPick list, I had to run `npm update` to get the latest `vscode` type definitions. This made some changes to the file `package-lock.json` - should this file be committed?

worksofliam:
@chrjorgensen 

> To use separators in the QuickPick list, I had to run npm update to get the latest vscode type definitions. This made some changes to the file package-lock.json - should this file be committed?

Yes, do that! We're likely due an upgrade anyway :)

> The redundant code for adding separators to the list - see [cb7da0c](https://github.com/halcyon-tech/vscode-ibmi/pull/844/commits/cb7da0c23d4d78f1e02fb55c0c33349989a361d4).

I see what you're trying to do. We can clean this up later!

Send up the changes to the package files and I'll do a final review!

chrjorgensen:
@worksofliam Packages updated

worksofliam (changes requested):
I love it! There is one thing I would like changed.

When I open the input box, I would like the current library as the initial value instead of it being blank.

image

worksofliam:
Side note: I love the use of `$(trash)` here!

chrjorgensen:
@worksofliam How about this:

billede

Current library is now presented and selected at the top and pressing Enter will keep the library as current library.
The placeholder text has also been changed to show it is a filter.
I have also restructured the logic for building the quickpick list.

Is it okay?

worksofliam:
@chrjorgensen that is beautiful! I will check it out today and we'll get it merged.

worksofliam:
@chrjorgensen Looking great! Thank you for this great PR.

PR #853, opened 2022-09-04
Option to show library description and attribute in User Library List view

chrjorgensen:
### Changes

This PR will add a configuration value to allow showing library description and attribute in the User Library List view.

This was made optional, since it is quite slow to retrieve the description and attribute for the libraries when SQL is not enabled.

Two techniques were tested during development:

1. retrieving information for one library at a time and
2. retrieving information for all libraries and filter according to the library list

The second option was chosen since there did seem to be any speed difference and this option was easier to code.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMiContent.js
+++ b/src/api/IBMiContent.js
@@ -248,9 +248,9 @@ module.exports = class IBMiContent {
     
   }
 
-
   /**
-   * @param {string[]} libraries
+   * Get list of libraries with description and attribute
+   * @param {string[]} libraries Array of libraries to retrieve
    * @returns {Promise<{name: string, text: string, attribute: string}[]>} List of libraries
    */
   async getLibraryList(libraries) {
@@ -282,7 +282,6 @@ module.exports = class IBMiContent {
       };
 
       results = results.filter(object => (libraries.includes(this.ibmi.sysNameInLocal(object.ODOBNM))));
-
     };
 
     if (results.length === 0) return [];
--- a/src/api/IBMiContent.js
+++ b/src/api/IBMiContent.js
@@ -273,7 +273,7 @@ module.exports = class IBMiContent {
         const library = libraries[i];
         await this.ibmi.remoteCommand(`DSPOBJD OBJ(QSYS/*ALL) OBJTYPE(*LIB) DETAIL(*TEXTATR) OUTPUT(*OUTFILE) OUTFILE(${tempLib}/${TempName})`);
       };
-      let results = await this.getTable(tempLib, TempName, TempName, true);
+      results = await this.getTable(tempLib, TempName, TempName, true);
 
       if (results.length === 1) {
         if (results[0].ODOBNM.trim() === ``) {
@@ -281,7 +281,7 @@ module.exports = class IBMiContent {
         }
       };
 
-      results = results.filter(object => (libraries.includes(object.ODOBNM)));
+      results = results.filter(object => (libraries.includes(this.ibmi.sysNameInLocal(object.ODOBNM))));
 
     };
 
@@ -289,7 +289,7 @@ module.exports = class IBMiContent {
 
     return results
       .map(object => ({
-        name: this.ibmi.sysNameInLocal(object.ODOBNM),
+        name: config.enableSQL ? object.ODOBNM : this.ibmi.sysNameInLocal(object.ODOBNM),
         attribute: object.ODOBAT,
         text: object.ODOBTX
       }))
--- a/src/api/IBMiContent.js
+++ b/src/api/IBMiContent.js
@@ -248,6 +248,56 @@ module.exports = class IBMiContent {
     
   }
 
+
+  /**
+   * @param {string[]} libraries
+   * @returns {Promise<{name: string, text: string, attribute: string}[]>} List of libraries
+   */
+  async getLibraryList(libraries) {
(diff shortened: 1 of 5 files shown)


worksofliam:
Looks interesting... will text over the next few days! Initial reaction is worried about the use of the `SPLIT` function in SQL as not everyone has that (yet)

chrjorgensen:
> Initial reaction is worried about the use of the `SPLIT` function in SQL as not everyone has that (yet)

Hm, this was built on an IBM i 7.3 system. What is the oldest IBM i release we support? Maybe we should write this in the documentation or README - or I may gotten this info previously but have just forgotten it... 😆 

I can add a check for the `SPLIT` function and if not found, redirect to using the (slower) DSPOBJD method.

worksofliam:
@chrjorgensen oh yep.. I should read the docs with both eyes. We only support 7.3 and up!

chrjorgensen:
@worksofliam No need for testing for SQL `SPLIT` function then, i.e. less work for me! 😆

worksofliam:
@chrjorgensen I am going to get the TS PR merged in first before I get this done. This will be next.

Thank you!

chrjorgensen:
@worksofliam Just a quick screenshot of what this PR is about...

billede

src/api/IBMiContent.js
```diff
@@ -248,6 +248,56 @@ module.exports = class IBMiContent {
     
   }
 
+
+  /**
+   * @param {string[]} libraries
```
worksofliam, line 253: I think a description above `@param` would be useful to describe this method.

src/api/IBMiContent.js
```diff
@@ -248,6 +248,56 @@ module.exports = class IBMiContent {
...
+        select os.OBJNAME as ODOBNM
+             , coalesce(os.OBJTEXT, '') as ODOBTX
+             , os.OBJATTRIBUTE as ODOBAT
+          from table( SYSTOOLS.SPLIT( INPUT_LIST => '${libraries.toString()}', DELIMITER => ',' ) ) libs
+             , table( QSYS2.OBJECT_STATISTICS( OBJECT_SCHEMA => 'QSYS', OBJTYPELIST => '*LIB', OBJECT_NAME => libs.ELEMENT ) ) os
+      `;
+      results = await this.runSQL(statement);
+    } else {
+      for (let i = 0; i < libraries.length; i++) {
+        const library = libraries[i];
```
worksofliam, line 273: Is this unused?

src/views/libraryListView.js
```diff
@@ -278,21 +278,21 @@ module.exports = class libraryListProvider {
...
-      currentLibrary.description = `(current library)`;
-      items.push(currentLibrary);
-
-      const libraryList = config.libraryList;
-
-      for (let library of libraryList) {
-        library = library.toUpperCase();
-        items.push(new Library(library));
+      if (config.showDescInLibList === true) {
+        libraries = await content.getLibraryList(new Array(config.currentLibrary).concat(config.libraryList));
```
worksofliam, line 289: Perhaps instead of using `concat`, we could use the spread syntax?

```
[config.currentLibrary, ...config.libraryList]
```

src/views/libraryListView.js
```diff
@@ -278,21 +278,21 @@ module.exports = class libraryListProvider {
...
+      } else {
+        libraries = new Array(config.currentLibrary).concat(config.libraryList).map(lib => { return { name: lib, text: ``, attribute: `` }});
```
worksofliam, line 291: Perhaps we could do the same here?

```
[config.currentLibrary, ...config.libraryList].map()
```

worksofliam (changes requested):
Functionally it's looking good. Most of the changes I am requesting is just code-related.

src/api/IBMiContent.js
chrjorgensen: No, this is used when showing library text without having SQL enabled...

src/views/libraryListView.js
chrjorgensen: The slow way...! 😊

src/api/IBMiContent.js
chrjorgensen: Fixed in [fe0b4af](https://github.com/halcyon-tech/vscode-ibmi/pull/853/commits/fe0b4afa5147d7658a3f985fc33c7e3f0f39f034)

src/views/libraryListView.js
chrjorgensen: Fixed in [d5f3d0a](https://github.com/halcyon-tech/vscode-ibmi/pull/853/commits/d5f3d0aedebbac40c5b40d46a7f4595ad82fb55b)

src/views/libraryListView.js
chrjorgensen: Fixed in [d5f3d0a](https://github.com/halcyon-tech/vscode-ibmi/pull/853/commits/d5f3d0aedebbac40c5b40d46a7f4595ad82fb55b)

worksofliam:
Looking good!! I suppose the only downside this is that it has to go and fetch the list again when we move items up/down. Then again, it's not an issue because we can just turn it off. Feature flags are my favourite!

A great PR - thank you @chrjorgensen 

image

image

PR #858, opened 2022-09-07
Speedup display of library text and fixes

chrjorgensen:
### Changes

This PR will
- speed up the library list view by omitting multiple (unnecessary) calls to get library information
- fix error when library was current library and also in library list (it would not show in the list part, only as current library)
- fix error displaying `(current library)` (when library was also in library list, `(current library)` would show in both places)

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMiContent.js
+++ b/src/api/IBMiContent.js
@@ -283,15 +283,13 @@ module.exports = class IBMiContent {
 
     if (results.length === 0) return [];
 
-    return results
-      .map(object => ({
-        name: config.enableSQL ? object.ODOBNM : this.ibmi.sysNameInLocal(object.ODOBNM),
-        attribute: object.ODOBAT,
-        text: object.ODOBTX
-      }))
-      .sort((a, b) => {
-        return (libraries.indexOf(a.name) - libraries.indexOf(b.name));
-      });
+    results = results.map(object => ({
+      name: config.enableSQL ? object.ODOBNM : this.ibmi.sysNameInLocal(object.ODOBNM),
+      attribute: object.ODOBAT,
+      text: object.ODOBTX
+    }));
+
+    return libraries.map(lib => {return results.find(info => info.name === lib)});
   }
 
   /**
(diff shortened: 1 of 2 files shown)


worksofliam (changes requested):
@chrjorgensen Seems to be a little bug when the library list has an entry when the library does not exist - and when this happens the User Library List view does not render anymore. Perhaps if this happens, we still should display the entry in the list and either: a) show no description or b) the description mentions that it was not found

image

chrjorgensen:
@worksofliam Good catch! 👍 

I have now changed `getLibraryList` to always return all the libraries in the list - if a library does not exist, it will have the text
`*** NOT FOUND ***` (commit [0b76ead](https://github.com/halcyon-tech/vscode-ibmi/pull/858/commits/0b76eade50b9acecd68269ac01bf8442555a365f))

I have also added a button for refresh of the library list and a command and button to validate the library list:

billede

Now the user can more easily get the current descriptions of the libraries in the list and validate / remove any libraries no longer on the server.

worksofliam:
Neat, that is a nice change.

I will get a review in ASAP this week!

Also, do you think that refresh and validation should be one button? I'm okay to have them as two buttons, but wonder if the UX is practically the same. Any ideas?

chrjorgensen:
A refresh will show the user if any libraries have been deleted or renamed - without removing them from the library list.
A validate will remove any libraries that have been deleted or renamed - and the library list will be changed and the user can't keep his/her list, if the library delete/rename was a mistake.

I understand your reasoning, though - one could argue that validate and refresh should be one operation. However, currently I think it should be up to the user to decide what to do with the library list: Either refresh and get a warning or sanitize the list and remove any libraries no longer available.

If screen space is an issue, the validate button could be moved to the secondary actions area in the view? The refresh button should always be available, like it is in the other views,

chrjorgensen:
@worksofliam @Wright4i 
I agree on your thoughts and have renamed the function `Validate library list` to `Cleanup library list` to inform the user, that the library list will be cleaned and may change.

I also changed the button to a secondary action, which now shows in the `...` - together with the variable maintenance button.
I think the change gives a cleaner look for the library list view:

billede

What do you think?

Wright4i:
Works for me!

Issue #861, opened 2022-09-10
Allow filters in Object Browser to change position


System-i-Developer:
The IFS Browser allows filters to be moved up and down the list.  It would be great if the same thing could be available for the Object Browser. Right now you can only change it by directly editing the settings.json file.

worksofliam:
Added in 1.5.14

PR #862, opened 2022-09-10
Allow filters to change position

chrjorgensen:
### Changes

This PR will

- add commands to move a filter up and down in the object browser
- add commands to move a filter to top or bottom of the list in the object browser
- add command to sort the filters by name in the object browser

It will solve the request made in issue #861 

### Checklist

* [x] have tested my change

### DIFF
--- a/src/views/objectBrowser.js
+++ b/src/views/objectBrowser.js
@@ -97,6 +97,26 @@ module.exports = class objectBrowserTwoProvider {
         }
       }),
 
+      vscode.commands.registerCommand(`code-for-ibmi.sortFilters`, async (node) => {
+        const config = instance.getConfig();
+
+        let objectFilters = config.objectFilters;
+
+        try {
+          objectFilters.sort(function(a, b){
+            const x = a.name.toLowerCase();
+            const y = b.name.toLowerCase();
+            if (x < y) {return -1;}
+            if (x > y) {return 1;}
+            return 0;
+          });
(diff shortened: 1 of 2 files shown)


src/views/objectBrowser.js
```diff
@@ -53,6 +53,70 @@ module.exports = class objectBrowserTwoProvider {
...
+            const x = a.name.toLowerCase();
+            const y = b.name.toLowerCase();
+            if (x < y) {return -1;}
+            if (x > y) {return 1;}
+            return 0;
+          });
+          await config.set(`objectFilters`, objectFilters);
+          if (Configuration.get(`autoRefresh`)) this.refresh();
+        } catch (e) {
+          console.log(e);
```
worksofliam, line 116: What is the use of the try/catch here? During testing, did something crash for you here?

worksofliam (changes requested):

src/views/objectBrowser.js
chrjorgensen: The code was copied from `sortIFSShortcuts` including the try/catch, I tested without try/catch and no errors were encountered, so the try/catch has been removed.

Wright4i:
@worksofliam another good PR by @chrjorgensen :)

Tested out fine. Filters move up and down. Saved their position on exit/window reload. Handled up/down/top/bottom gracefully whether the filter was collapsed or not and once reloaded was in the new position with the same state (open or collapsed).

PR #871, opened 2022-09-14
Add option to show names in uppercase

chrjorgensen:
### Changes

This PR will add a configuration option to show names and types in lowercase or original (upper/mixed) case:

billede

The option will be enabled by default to make the extension work as before this option. The user will have to disable it to have names shown in uppercase / mixed case.

Here is a sample screenshot with the option disabled:

billede

And another screenshot of a member list (IBM provides the text for files and members in QSYSINC as uppercase, this is not done by the extension):

billede

IMHO this option should be disabled by default to make the extension more familiar to old-school IBM i people. I myself prefer names in the /QSYS.LIB filesystem (objects and members) to be shown in uppercase. This would however change the visual of this extension, and that may not be what you want. So I have set it to `true` - enabled - for now.

There is no change in the IFS browser, everything is shown as stored in the server. I use mixed case here and think most people do the same.

WDYT?

### Checklist

* [x] have tested my change

### DIFF
--- a/package.json
+++ b/package.json
@@ -393,10 +393,10 @@
 					"default": true,
 					"description": "If enabled, listings will refresh when items are interacted with (create, copy, delele, etc). Turn this off if you find performance is bad."
 				},
-				"code-for-ibmi.ObjectBrowser.showNamesInUppercase": {
+				"code-for-ibmi.ObjectBrowser.showNamesInLowercase": {
 					"type": "boolean",
-					"default": false,
-					"description": "If enabled, names and types of object and members will be shown in uppercase (like in 5250 sessions)"
+					"default": true,
+					"description": "If enabled, names and types of object and members will be shown in lowercase. If disabled, the names will be shown as is on the server (like in 5250 sessions)"
 				},
 				"code-for-ibmi.connections": {
 					"type": "array",
--- a/package.json
+++ b/package.json
@@ -393,6 +393,11 @@
 					"default": true,
 					"description": "If enabled, listings will refresh when items are interacted with (create, copy, delele, etc). Turn this off if you find performance is bad."
 				},
+				"code-for-ibmi.ObjectBrowser.showNamesInUppercase": {
+					"type": "boolean",
+					"default": false,
+					"description": "If enabled, names and types of object and members will be shown in uppercase (like in 5250 sessions)"
+				},
 				"code-for-ibmi.connections": {
 					"type": "array",
 					"items": {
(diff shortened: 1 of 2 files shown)


chrjorgensen:
@worksofliam I have made this option a global one, since I think it should be consistent across systems. Do you have another opinion?

The same could be said of the option to show library descriptions and attributes in PR #853 - so now I'm in doubt about that one, if it also should be global...

worksofliam:
I really like this PR. This is something that gets brought up a lot outside of GitHub.

Though, I have to say that I think the config is a little backwards. The config is titled "show names in uppercase" and if it is disabled, then force everything to lowercase. If we enable it, it is still possible for system objects to have mixed-case (`"myobj"` for example is a valid object name and would not be uppercase when enabled)

What I think makes more sense is:

A new flag named "show names in lowercase" and if it is enabled, then force everything to lowercase. This means, when it is disabled, it will show everything the way the system displays it (uppercase, mixed-case, etc).

Let me know what you think.

chrjorgensen:
@worksofliam I totally agree on switching the config to the opposite "Show names in lowercase"! But what should be the default for existing extension users - disabled (false), showing names in uppercase or enabled (true), showing names in lowercase as done until now?

It was made like this since this was a new config option to enable my liking of uppercase names - and I was (after more than three decades on this platform!) not aware of the ability for "mixed" case names! :laughing: 

I will change the config when I know what you want as the default for it...

worksofliam:
@chrjorgensen 

> But what should be the default for existing extension users - disabled (false), showing names in uppercase or enabled (true), showing names in lowercase as done until now?

I guess the default should be what the existing behaviour is, which would be: show the names in lowercase by default!

chrjorgensen:
@worksofliam The config option naming and default has been reverted. I have also changed the original comment describing the PR to reflect this change.

Wright4i:
@worksofliam LGTM. Primary concern in my testing was how things like `Go To File...` or `Rename Member` would handle the naming with the new config disabled. Everything worked as expected. Yes, even mixed case came back mixed!
image

I'm on the fence on enabled by default/not.
* On the one hand lowercase is more conducive to the goal of making vscode-ibmi friendly to newcomers that have vscode experience, but no RPG experience.
* On the other hand if the system is  *SCREAMING* at you all the time anyway then those kids better get used to it 🤣

In the end it doesn't really matter. It's trivial to change a setting and anyone used to vscode knows to go check out an extensions settings. I don't think I have a single extension configured with "out of the box" settings on my workspace!

worksofliam:
@Wright4i Neat, thank you!! I am aiming for lowercase by default because that is how the existing functionality works by default.

PR #878, opened 2022-09-21
Add option to sort objects by name or type in object browser

chrjorgensen:
### Changes

This PR will add an option for sorting the objects in object browser by name or type:

billede

When this option is enabled, objects are sorted by name and type. If disabled, objects are sorted by type and name - the type being the MI object type, so the list will be sorted like in PDM.

The option is enabled by default to preserve the current behavior of this extension.

### Checklist

* [x] have tested my change


src/api/IBMiContent.js
```diff
@@ -499,4 +507,104 @@ module.exports = class IBMiContent {
 
     return errors;
   }
+
+  /**
+   * @returns {Map([])} objTypes
+   */
+  getObjectTypeMap() {
+    // https://www.ibm.com/docs/api/v1/content/ssw_ibm_i_73/rbam6/rbam6objecttypes.htm
+    return new Map([
```
worksofliam, line 516: Is there a reason you are using a map over a key-value object here?

src/api/IBMiContent.js
```diff
@@ -350,6 +351,8 @@ module.exports = class IBMiContent {
         }
       }
 
+      const internalObjType = this.getObjectTypeMap();
```
worksofliam, line 354: Since the result of `getObjectTypeMap` is always the same, maybe we should just make it global somehow?

src/api/IBMiContent.js
chrjorgensen: The reason was... my lack of Javascript knowledge and relying on W3Schools... 😆 

But after some investigating on Maps, it seems to me that Maps are more modern than key-value objects - and may perform faster: [https://bretcameron.medium.com/how-javascript-maps-can-make-your-code-faster-90f56bf61d9d](https://bretcameron.medium.com/how-javascript-maps-can-make-your-code-faster-90f56bf61d9d)

Should we keep it?

src/api/IBMiContent.js
chrjorgensen: Any suggestion on how to make it global and constant? I put it into a separate function at the bottom of the source, since I don't think it should be in the middle of the code.

worksofliam: Let me have a think about this and get back to you!

src/api/IBMiContent.js
worksofliam: Yes, let's keep it. Thanks!

worksofliam:
I will look into how where we should put this global object and then we should be good to get going!

worksofliam:
Will leave update today. Sorry it's been a minute.

worksofliam:
@chrjorgensen Here is what I think:

See the `schemas` directory. We keep constants and initial values in there. Perhaps the `getObjectTypeMap` method could be extracting into a new ts file and put inside this directory and then brought into `IBMiContent.js`.

What do you think of that?

chrjorgensen:
@worksofliam I'm sorry, but my TypeScript knowledge is not enough to make a ts import file - so I made a JavaScript import file instead. 😞

worksofliam:
@chrjorgensen no problem! I will let this in, and then I'm going to do another PR where I convert it to TS. I will let you review that!

src/api/IBMiContent.js
```diff
@@ -360,7 +364,12 @@ module.exports = class IBMiContent {
         }))
         .sort((a, b) => {
           if (a.library.localeCompare(b.library) != 0) return a.library.localeCompare(b.library)
-          else return a.name.localeCompare(b.name);
+          else if (sortOrder === `name`) return a.name.localeCompare(b.name)
+          else {
```
worksofliam, line 368: Perhaps the logic here is round the wrong way?

worksofliam (changes requested):
@chrjorgensen Looking good, except one thing! Is the ordering around the wrong way? It looks like it was `Z-A` instead of `A-Z`.

image

src/api/IBMiContent.js
chrjorgensen: No - the sort order shown is by internal(!) object type (the hex value from getObjectTypesMap!

This is the same order as PDM and familiar to IBM i developers. Programs come first, then service programs etc.

worksofliam: Right!! Sorry, I haven't used it in years 😂 

It works great. I am going to try it with a larger library (QSYS) and then will likely merge!

worksofliam (approved):

Issue #883, opened 2022-09-26
Detect record length, even for source files with no existing data


bdietz400:
With an existing source file with record length of 112.

right click to create a new member and paste some code that extends past column 80.
i see "error" that 'content past record lenght of 80'

Extension version: 1.5.15
VS Code version: Code 1.71.2 (74b1f979648cc44d385a2286793c226e611f59e7, 2022-09-14T21:03:37.738Z)
OS version: Windows_NT x64 10.0.19044
Modes:
Sandboxed: No

worksofliam:
@bdietz400 Do you have source dates enabled or disabled?

worksofliam:
Guessing enabled.

Please share the contents of your Code for IBM i output channel after restarting VS Code and recreating the issue.

<img width="939" alt="image" src="https://user-images.githubusercontent.com/3708366/178739532-03c8ad30-84de-4698-9904-65938f2ef651.png">

bdietz400:
Enable source dates is checked.
re create progressing....

bdietz400:
![past80](https://user-images.githubusercontent.com/7241553/192364711-e661ecb2-7446-4996-b21f-951339c3139f.jpg)
[past80.txt](https://github.com/halcyon-tech/vscode-ibmi/files/9649232/past80.txt)

worksofliam:
So I can see it's running this:

```
SELECT LENGTH(srcdta) as LENGTH FROM ILEDITtmp.VSCODE_SOURCE112_CODEPAST80 limit 1
```

And returning

`LENGTH: 0`.

This might be because it's a brand new source file with no existing members in! Does this error also appear on new members in existing source files with members in them?

bdietz400:
it happens when there are existing member and a new one is created.
once the member exists there are no problems

worksofliam:
@Wright4i I forget.. did we release this yet?

Wright4i:
@worksofliam we in fact have not released this yet. Actually, that reminds me.. we also need to release the ibm-i-development-pack after [PR#5](https://github.com/halcyon-tech/ibm-i-development-pack/pull/5). I watched you demo then have to separately add in the Display Renderer extension!

worksofliam:
@Wright4i i can do that soon!

PR #884, opened 2022-09-26
Fix for #883 - Uses ROW_LENGTH from SYSTABLES for new source members

Wright4i:
### Changes

Fix for #883. Uses QSYS2/SYSTABLES to get ROW_LENGTH-12 when unable to get a row length using LENGTH(srcdta) when using SQL (Source Dates enabled)

image

### Checklist

* [x] have tested my change

### DIFF
--- a/src/filesystems/qsys/complex/content.js
+++ b/src/filesystems/qsys/complex/content.js
@@ -41,7 +41,12 @@ module.exports = class IBMiContent {
       if (result.length > 0) {
         recordLengths[alias] = result[0].LENGTH;
       } else {
-        recordLengths[alias] = DEFAULT_RECORD_LENGTH;
+        const result = await content.runSQL(`SELECT row_length-12 as LENGTH FROM QSYS2.SYSTABLES WHERE TABLE_SCHEMA = '${lib}' and TABLE_NAME = '${spf}' limit 1`);
+        if (result.length > 0) {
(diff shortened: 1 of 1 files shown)


worksofliam:
I like this but I am concerned about the performance of SYSTABLES. I have a faint memory that it performs terribly when there are thousands of members. 

I have some system with that setup, so I will have to try it out!

Wright4i:
SYSTABLES is being used in getMemberList ([IBMiContent.js)](https://github.com/halcyon-tech/vscode-ibmi/blob/9a951855d82d6b6f44df7e01292e67ac85e6dda6/src/api/IBMiContent.js#L374) already and we're only doing the extra call if the original check on the length of srcdta fails. I think any performance issue would be negligible in this instance.

Open to ideas though. Only other idea I have right now would be force-saving an empty member then doing the length(srcdta) check, but that sounds even less performant.

Wright4i:
But since you mention it.. I happen to have one with a few thousand source members in it too! 3473 to be exact. 107ms to pull row_length back from SYSTABLES and I don't think anyone has ever described this server as being snappy lol
image

worksofliam:
That's a good check @Wright4i! I'm gonna have a little test tonight/tomorrow and we'll get this merged ASAP. Thank you!

worksofliam:
Thank you @Wright4i!!

This is an awesome change. Thank you for the quick turn around.

PR #923, opened 2022-10-27
Remove CL language provider

worksofliam:
### Changes

We are removing the feature flag to enable the CL content assist / completion provider. Since this functionality is moving over to a new extension, vscode-clle, we no longer need to do it here. This extension will still need to install the required program (`GENCMDXML`) for the CL content assist to work, so that functionality will stay.

Now it will work like so:

1. User connects to system
2. Checks if `IBM.vscode-clle` is installed 
3. If it is, ask the user if they want to install the CL language tools onto their system (already happening)
3. If yes, check if `GENCMDXML` is installed into temporary library
4. vscode-clle will use the Code for IBM i API to call `GENCMDXML`.

### Tasks

* [x] Remove completion provider for CL
* [x] Replace check if functionality should be enabled
    * [ ] Remove feature flag
    * [ ] Check if `IBM.vscode-clle` is installed instead
* [x] vscode-clle released

### Checklist

* [x] have tested my change


worksofliam:
I have tested myself and am allowing a merge without external review.

PR #924, opened 2022-10-27
Fixed "Add to library list" action

sebjulliand:
"Add to library list" action is broken as soon as a library is selected in the view. Once this happen, the command receives the selected node in parameter, and never prompts for a name anymore, which results in the error message "Library is too long." being displayed right away.

### Changes

Checking if the "newLibrary" parameter is actually a string fixes the issue.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/views/libraryListView.js
+++ b/src/views/libraryListView.js
@@ -130,7 +130,7 @@ module.exports = class libraryListProvider {
 
         let libraryList = [...config.libraryList];
 
-        if(newLibrary == ``){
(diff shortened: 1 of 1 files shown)


worksofliam:
We just fixed this in another view as well. Thanks for your PR @sebjulliand - I'll have a review done and merged tomorrow for a release.

I really appreciate it!

sebjulliand:
My pleasure. Hopefully there will be more to come :)

worksofliam (approved):
Looking good!

PR #929, opened 2022-10-28
Converted helpView to TypeScript

sebjulliand:
### Changes

* Converted helpView to TypeScript (let's start small)
* Made `instance` a const

### Checklist

* [x] have tested my change


worksofliam (approved):
Testing and it's looking good to me. Thanks you!

PR #930, opened 2022-10-30
Fixed "Name of profile" prompt

sebjulliand:
### Changes
The "Name of profile" input box would show "[Object object]" as the default value when no "last profile name" is already stored.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/views/profilesView.js
+++ b/src/views/profilesView.js
@@ -33,7 +33,7 @@ module.exports = class profilesProvider {
         let currentProfiles = config.connectionProfiles;
 
         const profileName = profileNode ? profileNode.profile : await vscode.window.showInputBox({
-		      value: currentProfile,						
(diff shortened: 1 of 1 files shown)


src/views/profilesView.js
```diff
@@ -33,7 +33,7 @@ module.exports = class profilesProvider {
         let currentProfiles = config.connectionProfiles;
 
         const profileName = profileNode ? profileNode.profile : await vscode.window.showInputBox({
-		      value: currentProfile,						
+	  value: typeof currentProfile === `object` ? `` : currentProfile,
```
worksofliam, line 36: @sebjulliand I think it should be the overway around as it makes more sense.

```js
typeof currentProfile === `string` ? currentProfile : ``
```

What do you think?

sebjulliand: Oh my, yes that's way more logical ☺

PR #939, opened 2022-11-03
Enable switching between source member edit modes

worksofliam:
### Changes

At request of @jkyeung, I have add the newer diff-based source date approach as an optional feature. There are two reasons I have done this:

* Allow users to test this feature.
* Allow users who want to test this feature also stay up to date with other features.

### Checklist

* [x] have tested my change


worksofliam:
@jkyeung please try this build and if you're happy with the way to switch between editing modes, then we can merge and deploy it. Thanks.

jkyeung:
The interface for switching between the two modes is good! Go for it!

(As a more general comment, I personally have not been able to develop a feel for what settings are "extension" settings and which are "connection" settings. Perhaps it would be more obvious if I regularly worked with more than one connection. But given the status quo, it makes sense to add this new thing to the same place where the source dates are enabled.)

worksofliam:
@jkyeung Awesome - glad it's looking good. I will merge this in tomorrow.

Thanks!

PR #944, opened 2022-11-04
Converted LocalLanguageActions to TypeScript

sebjulliand:
@worksofliam I started from scratch since it was a small change. It's cleaner this way.

### Changes
* Converted `LocalLanguageActions` to TypeScript
* `type` in `Action` is now optional (or `type` would have to be defined on every predefined action)

### Checklist
* [x] have tested my change


worksofliam (approved):
Looking and working good. Thank you!

PR #959, opened 2022-11-11
Converted Colors/ColorProvider to TypeScript

sebjulliand:
### Changes
* Converted Colors to TypeScript
* Renamed Colors to SEUColors
* Converted ColorProvider to TypeScript
* Renamed ColorProvider to SEUColorProvider

### Checklist
* [x] have tested my change


worksofliam:
So I am not sure where I stand with this PR. I am glad it's converted to TS, but this feature really should not even exist as part of this extension. I am very grateful, but I believe no more effort should be put into supporting SEU colours.

worksofliam:
@halcyon-tech/core I'd love an extra opinion on whether we should merge this PR based on my comment above.

chrjorgensen:
> @halcyon-tech/core I'd love an extra opinion on whether we should merge this PR based on my comment above.

@worksofliam I personally never liked or used SEU colours - so for my part this extension could just ignore (or even remove) them.
But this is not a new feature, just a rewrite by @sebjulliand of existing code, so I think it should be merged.

Later we can decide whether SEU colours should still be supported or not... 😉

Wright4i:
@worksofliam I unfortunately have to deal with source colors in old programs every day. I do actively remove them as I come across source with it. Like @chrjorgensen said I'm happy to see this merged in as-is as it's a rewrite of existing code. 

An option to auto-remove source coloring would certainly be nice. Maybe that's an idea for the rpgle linter when formatting document with a rule to flag the color codes as warnings?

worksofliam:
@sebjulliand Whenever you have a moment, please fix the conflict and I will do some testing. Thanks!

sebjulliand:
Will do!

PR #960, opened 2022-11-14
Fixed ILEObject icon

sebjulliand:
### Changes
ILEObject icon would always be `circle-large-outline` when the `showNamesInLowercase` setting was true, because the object's type would be lowercase and wouldn't match the icons array's.

### Checklist
* [x] have tested my change


worksofliam:
Nice! Thanks @sebjulliand :)

PR #980, opened 2022-12-09
Added methods/facilitators to exported API

sebjulliand:
### Changes
Added a few methods in exported API:
* Getters for connection `name`, `host`, `user `and `port`
* `uploadFiles`
* `downloadFile`
* `onEvent` to easily subscribe to connection events

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -830,22 +830,6 @@ export default class IBMi {
     this.client.getFile(this.fileToPath(localFile), remoteFile);
   }
 
-  getHost() : string{
-    return this.currentHost;
-  }
-
-  getPort() : number{
-    return this.currentPort;
-  }
-
-  getUser() : string{
-    return this.currentUser;
-  }
-
-  getConnectionName() : string{
-    return this.currentConnectionName;
-  }
-
   fileToPath(file : string | vscode.Uri) : string{
     if(typeof file === "string"){
       return file;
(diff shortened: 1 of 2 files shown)


src/api/IBMi.ts
```diff
@@ -822,5 +822,36 @@ export default class IBMi {
...
 
+  getUser() : string{
+    return this.currentUser;
+  }
+
+  getConnectionName() : string{
+    return this.currentConnectionName;
+  }
+
+  fileToPath(file : string | vscode.Uri) : string{
```
worksofliam, line 849: @sebjulliand Any chance you can explain what your use case for this is? I feel like this is mostly a helper function for others as it isn't really used anywhere in this extensions.

worksofliam (approved):
Looks great! Just one comment above and we should be good to go.

src/api/IBMi.ts
sebjulliand: @worksofliam of course! I exposed the current connection name at that time to be able to store this information and reuse it in my extension (mainly to implement a "reconnect" and "autoconnect" feature, which I'll be happy to port back in vscode-for-i if you feel like it would be useful).

However, with the new typings export mechanism, these helpers (connectionName, user, host and port) don't seem to be relevant anymore as the fields are exposed in the generated .d.ts files. I'll remove them and keep upload and download 😉

worksofliam:
Thanks @sebjulliand :) If this being released urgent? I am aiming to release this (and a typings update) later in the week.

sebjulliand:
@worksofliam no urgency, I can work with my own typing for now. Take your time 👍🏻

PR #988, opened 2022-12-17
Converted error handlers to TypeScript

sebjulliand:
@worksofliam I managed to test the old handler and it looks fine. However, I didn't manage to get some `EXPANSION` to test new handler. How do you get that?

### Changes
* Converted error handlers (new & old) to TypeScript
* Updated `CompileTools.ts` accordingly

### Checklist
* [x] have tested my change


worksofliam:
@sebjulliand EXPANSION is generated from any of the SQL precompilers.

sebjulliand:
Gotcha, thanks; seems like it runs just fine too 👌

PR #990, opened 2022-12-18
Converted deployment to TypeScript

sebjulliand:
### Changes
* Converted `deployment` to TypeScript
* Refactored to split logic into several methods
* Exported a bare `deploy` function that may be exported to other extensions later on
* Fixed remote path output on Windows when deploying `all`
* Prevented `.gitignore` itself from being deployed
* Bumped TS target to `es2019`

### Checklist
* [x] have tested my change
  * [x] Deploy all
  * [x] Deploy staged
  * [x] Deploy unstaged
  * [x] Deploy changed


src/api/local/deployment.ts
```diff
@@ -0,0 +1,477 @@
...
+      } else {
+        throw new Error(`No repository found for ${parameters.localFolder.fsPath}`);
+      }
+    } else {
+      throw new Error(`No repositories are open.`);
+    }
+  }
+
+  async function deployAll(parameters: DeploymentParameters) {
+    const name = parameters.localFolder.path.split('/').reverse().pop();
```
worksofliam, line 355: I am not sure this is a valid way to get the `name`. It is blank on Mac. 

image

worksofliam: Value of `localFolder.path`: `/Users/barry/Repos/ibmi-company_system`

worksofliam: You might be better off with something like: `require(`path`).basename(`/Users/barry/Repos/ibmi-company_system`)`

worksofliam (changes requested):
Tested all deployment options and the only issue is as described with the `name` in the `All option.

worksofliam:
LGTM! Thank you @sebjulliand 🥳

PR #995, opened 2022-12-21
Added Filter and error parser to exported API

sebjulliand:
### Changes
* Extracted and exported `Filter` interface (used by the `fs` extension)
* Added a few types declarations where relevant

### Checklist
* [x] have tested my change


worksofliam:
@sebjulliand Looks like we have some conflicts here from the last PR. Can you fix these whenever you have a moment?

Thanks!
Liam

sebjulliand:
@worksofliam while this PR is still open, I added the error parser to the exported API as well. Hopefully that's OK with you 😉

src/typings.ts
```diff
@@ -96,4 +97,18 @@ export interface IFSFile {
 export interface IBMiError{
   code: string
   text: string
+}
+
+export interface Filter{
```
worksofliam, line 102: I see this interface was added, but I don't see this `Filter` interface used anywhere in this extension.

Is this intentional?

worksofliam (changes requested):
Check out my question above. Thanks!

src/typings.ts
sebjulliand: Yes, it's intentional.

I left a comment in the `ObjectBrowser`. I renamed `Filter` class to `FilterItem` and commented that it will implement Filter (it does, but it's implicit at the moment).
Since JS doesn't support implementing an interface, it's all I could do.
This is more or less a `todo` for when I'll convert `ObjectBrowser` to TypeScript 😅

worksofliam:
LGTM!!

PR #999, opened 2022-12-25
Implemented MD5 based deployment

sebjulliand:
🎅🎄Ho ho ho, Merry Christmas! 🎄🎅
### Changes
Added a new deployment method called `MD5 compare`. It uses files' MD5 hash from local workspace and remote deployment location to:
* Upload only changed files since last deployment
* Delete remote files not found in the local workspace

It also clears empty folders in the remote deployment location.

The feature uses the `md5sum` shell command on the remote system. The extension will check if the command is available in order to enable the feature.

The `deploy` method has also been added to the exported function in the extension's API, to let other extension run unattended deployments.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/local/deployment.ts
+++ b/src/api/local/deployment.ts
@@ -1,10 +1,7 @@
 
-import path, { basename } from 'path';
+import path, { basename, dirname } from 'path';
 import vscode from 'vscode';
-
-import IBMi from '../IBMi';
 import { getLocalActions } from './actions';
-
 import { ConnectionConfiguration } from '../Configuration';
 import { LocalLanguageActions } from '../../schemas/LocalLanguageActions';
 import { GitExtension } from '../import/git';
@@ -13,6 +10,9 @@ import Instance from '../Instance';
 import { Ignore } from 'ignore'
 import ignore from 'ignore'
 import { NodeSSH } from 'node-ssh';
+import { readFileSync } from 'fs';
+import Crypto from 'crypto';
+import IBMi from '../IBMi';
 
 export namespace Deployment {
   interface Upload {
@@ -25,7 +25,12 @@ export namespace Deployment {
     method: Method
     localFolder: vscode.Uri
     remotePath: string
-    ignoreRules?: Ignore    
+    ignoreRules?: Ignore
+  }
+
+  interface MD5Entry {
+    path: string
+    md5: string
   }
 
   const BUTTON_BASE = `$(cloud-upload) Deploy`;
@@ -39,7 +44,8 @@ export namespace Deployment {
     "all",
     "staged",
     "unstaged",
-    "changed"
+    "changed",
+    "md5synch"
   }
 
   export function initialize(context: vscode.ExtensionContext, instance: Instance) {
@@ -81,9 +87,7 @@ export namespace Deployment {
             `Ignore`
           ).then(async result => {
             if (result === `Yes`) {
-              await connection.sendCommand({
-                command: `mkdir -p "${possibleDeployDir}"`
-              });
+              await createRemoteDirectory(possibleDeployDir);
 
               existingPaths[workspace.uri.fsPath] = possibleDeployDir;
               try {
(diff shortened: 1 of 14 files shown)


src/api/local/deployment.ts
```diff
@@ -172,13 +176,18 @@ export namespace Deployment {
...
+          { method: Method.unstaged, label: `Working Changes`, description: `Unstaged changes in git` },
+          { method: Method.staged, label: `Staged Changes`, description: `` },
+          { method: Method.all, label: `All`, description: `Every file in the local workspace` }
+        ];
+
+        if (getConnection().remoteFeatures.md5sum) {
+          methods.push({ method: Method.md5synch, label: `MD5 compare`, description: `Synchronizes using MD5 hash comparison` });
+        }
+
+        const method = (await vscode.window.showQuickPick(methods.sort((m1, m2) => m1.label.localeCompare(m2.label)),
```
worksofliam, line 190: I am not sure how I feel about this sort.

This means `All` is always showing up at the top, and it's rare that we would want 'all' every time. I think we should aim to have `Changed` or `MD5 Compare` at the top.

src/api/local/deployment.ts
```diff
@@ -172,13 +176,18 @@ export namespace Deployment {
...
-          ],
+        const methods = [
+          { method: Method.changed, label: `Changes`, description: `${changes.size} change${changes.size > 1 ? `s` : ``} detected since last upload. ${!changes.size ? `Will skip deploy step.` : ``}` },
+          { method: Method.unstaged, label: `Working Changes`, description: `Unstaged changes in git` },
+          { method: Method.staged, label: `Staged Changes`, description: `` },
+          { method: Method.all, label: `All`, description: `Every file in the local workspace` }
+        ];
+
+        if (getConnection().remoteFeatures.md5sum) {
+          methods.push({ method: Method.md5synch, label: `MD5 compare`, description: `Synchronizes using MD5 hash comparison` });
```
worksofliam, line 187: Personally, I think we need a different name than 'MD5 Compare'. Perhaps even just `Compare` is ok, as the description has got it covered.

worksofliam (changes requested):
This is very awesome change. I've tested in with a few different situations and it is looking good. My only concern right now is that I don't want the 'md5 compare' upload method to upload items defined in my `.gitignore`. 

It doesn't look like the `deployMD5` is looking at the ignore rules. I think if a file in the `uploads` is in the ignore rules, then it shouldn't be uploaded.

What do you think?

src/api/local/deployment.ts
sebjulliand: Sounds good to me! I'll put both on top (`Compare` then `Changed`).

src/api/local/deployment.ts
sebjulliand: Let's go with `Compare` then. 👍

sebjulliand:
> This is very awesome change. I've tested in with a few different situations and it is looking good. My only concern right now is that I don't want the 'md5 compare' upload method to upload items defined in my `.gitignore`.
> 
> It doesn't look like the `deployMD5` is looking at the ignore rules. I think if a file in the `uploads` is in the ignore rules, then it shouldn't be uploaded.
> 
> What do you think?

You're right to ask, and actually, `Compare` does not uploads what's in `.gitignore`. At least, it's not supposed to. The exclusion is done in the `findFiles` method which lists all the local files but those matching a pattern from `.gitignore`.

src/api/local/deployment.ts
```diff
@@ -474,4 +578,39 @@ export namespace Deployment {
...
+  function toRelative(root: vscode.Uri, file: vscode.Uri) {
+    return path.relative(root.path, file.path).replace(/\\/g, `/`);
+  }
+
+  async function findFiles(parameters: DeploymentParameters, includePattern: string, excludePattern?: string) {
+    const root = parameters.localFolder.fsPath;
+    return (await vscode.workspace.findFiles(new vscode.RelativePattern(root, includePattern),
+      excludePattern ? new vscode.RelativePattern(root, excludePattern) : null))
+      .filter(file => {
+        if (parameters.ignoreRules) {
```
sebjulliand, line 607: @worksofliam: this is where the files matching `.gitignore` get excluded for the `Compare` method.

worksofliam: Thanks for pointing that out for me!

worksofliam:
Review in coming later today.

worksofliam:
LGTM. Thank you @sebjulliand - this is really an outstanding change.

worksofliam (approved):

PR #1001, opened 2022-12-27
Fixed launchActionsSetup crashing

sebjulliand:
### Changes
Fixed `launchActionsSetup` crashing right at the beginning because of a wrong call to `getWorkspaceFolders` prefixed by `module.exports` (how I let that slipped is still a mystery...sorry!).

### Checklist
* [x] have tested my change


worksofliam (approved):
I have verified this fix is working. Thank you!

PR #1014, opened 2023-01-18
Activate extension on startup finished

sebjulliand:
### Changes
Changed extension's activation event to [onStartupFinished](https://code.visualstudio.com/api/references/activation-events#onStartupFinished), so it can handle whatever it needs to right away.

### Checklist
* [x] have tested my change

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,8 +32,7 @@
 		"theme": "dark"
 	},
 	"activationEvents": [
-		"onView:connectionBrowser",
-		"onCommand:code-for-ibmi.connect"
+		"onStartupFinished"
 	],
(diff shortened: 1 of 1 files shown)


worksofliam:
@sebjulliand I do know that in recent VS Code releases we no longer need the command activations, but do you think it's okay that we remove the `onView`?

sebjulliand:
> @sebjulliand I do know that in recent VS Code releases we no longer need the command activations, but do you think it's okay that we remove the `onView`?

@worksofliam I think so...`onStartupFinished` seems to be the earliest yet safest stage to activae the extension without issue.

PR #1027, opened 2023-01-27
Handle permission denied in IFS browser

sebCIL:
### Changes

Handle permission denied in IFS browser.
Show an error message if permission is denied.
image

### Checklist

* [X] have tested my change
* [X] have added myself to the contributors' list in [CONTRIBUTING.md](https://github.com/halcyon-tech/vscode-ibmi/blob/master/CONTRIBUTING.md)

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -456,26 +456,38 @@ export default class IBMiContent {
    * @param remotePath 
    * @return an array of IFSFile
    */
-  async getFileList(remotePath: string): Promise<IFSFile[]> {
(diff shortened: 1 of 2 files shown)


worksofliam (changes requested):
@sebCIL This is good change and it makes sense, but I don't think this function should be returning `IBMiError`. I think it needs actually throw an error instead.

```
throw new Error(fileListResult.stderr);
```

.. which should then be handled inside `ifsBrowser.js`.

Let me know what you think.

Thanks!

sebCIL:
You're right, it is better.
I will do that.

sebCIL (commented):
@worksofliam I have made the change.

PR #1043, opened 2023-02-04
Add Compile support for PFSQL type source members

bojasv:
added RUNSQLSTM configuration for PFSQL type source members

### Changes

Description of change here.

### Checklist

* [x] have tested my change

### DIFF
--- a/package.json
+++ b/package.json
@@ -647,6 +647,22 @@
 							"command": "?RUNSQLSTM SRCFILE(&OPENLIB/&OPENSPF) SRCMBR(&OPENMBR) COMMIT(*NONE) NAMING(*SQL)"
 						},
 						{
+							"type": "member",
+							"extensions": [
+								"PFSQL"
+							],
+							"name": "Run SQL Statements (RUNSQLSTM)",
+							"command": "?RUNSQLSTM SRCFILE(&OPENLIB/&OPENSPF) SRCMBR(&OPENMBR) COMMIT(*NONE) NAMING(*SQL)"
+						},
+						{
+							"type": "member",
+							"extensions": [
+								"pfsql"
+							],
+							"name": "Run SQL Statements (RUNSQLSTM)",
+							"command": "?RUNSQLSTM SRCFILE(&OPENLIB/&OPENSPF) SRCMBR(&OPENMBR) COMMIT(*NONE) NAMING(*SQL)"
+						},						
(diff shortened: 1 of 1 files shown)


chrjorgensen:
@worksofliam IMHO This is a very user specific source member  type, which does not belong together with the IBM-defined member types. The RUNSQLSTM command is used quite often, but this special member type is not.

Too many special member types does clutter the command list. WDYT?

I do appreciate the will and energy used to contribute to the extension, though! :+1:

priceaj:
whats PFSQL?

bojasv:
Hi @chrjorgensen Thank you for your comment. Before this, I never realized that a member type not listed for the OS could be supported by a native command (RUNSQLSTM), esp. that I can't find any relevant reference to PFSQL on IBMi. 

Only after reading thru the help menu, I realized that apart from the OS-supplied standard member types, we could have our own SRCMBR types as well, but then how does the OS recognizes them the way we expect it to work? It's new and surprising for me. 

For PFSQL, the member type got validated by RUNSQLSTM's smooth execution and never let me question its authenticity. 

@worksofliam Perhaps we could have an easy way to let users configure the extn for user-defined member types, but to me it looks an trivial issue so far.

priceaj:
can you not already do this when creating a user defined action?

bojasv:
> can you not already do this when creating a user defined action?

Oh nice, spare the newbie please ;)

priceaj:
So if you click "Actions" at the bottom, then you can user define a new Member Action, and you can key in whatever extension you want

worksofliam:
@bojasv first, let me say thank you very much for your PR. It was closed before I really got to take a look at it. It looks like our other community members have helped me out here with a swift response.

@chrjorgensen you are correct. The default actions should only reference the 'valid' IBM i types, in which `PFSQL` is not one.

@priceaj thanks for directing @bojasv on how to create an action!

chrjorgensen:
@bojasv

> Only after reading thru the help menu, I realized that apart from the OS-supplied standard member types, we could have our own SRCMBR types as well, but then how does the OS recognizes them the way we expect it to work? It's new and surprising for me.

The member type can be anything you like - as you just discovered. And it is not validated by any command!

It is only used by the development tools like this extension, Rational Developer for i, PDM etc. to link to the correct command for the source.

PR #1075, opened 2023-02-18
Added "protected" option to filters (revival)

sebjulliand:
### Changes
Added a `protected` setting to Object filters, as a possible answer to issue https://github.com/orgs/halcyon-tech/discussions/1048.

When a filter is protected:
- a lock icon is shown next to its name
- no modifying action can be taken on any of its children
- members can be opened but cannot be saved
- members listed from a `protected` filter search cannot be saved

The `ctrl+alt+o` shortcut is now `Go to file (read only)` and allows to open members in protected mode (i.e. save is disabled).

### Checklist

* [x] have tested my change


chrjorgensen:
@sebjulliand @worksofliam I did a quick checkout of Seb's branch to see the function - but on my computer no filters are shown after connection to any system, and the debug console output shows this:

```
rejected promise not handled within 1 second: Error: ES Modules may not assign module.exports or exports.*, Use ESM export syntax, instead: ./src/filesystems/qsys/complex/index.js
extensionHostProcess.js:100
stack trace: Error: ES Modules may not assign module.exports or exports.*, Use ESM export syntax, instead: ./src/filesystems/qsys/complex/index.js
	at Object.set [as exports] (d:\Data\Develop\code-for-ibmi\dist\extension.js:42426:21)
	at ./src/filesystems/qsys/complex/index.js (d:\Data\Develop\code-for-ibmi\dist\extension.js:36286:16)
	at __webpack_require__ (d:\Data\Develop\code-for-ibmi\dist\extension.js:42384:42)
	at loadAllofExtension (d:\Data\Develop\code-for-ibmi\dist\extension.js:29392:35)
	at Login.LoginToPrevious (d:\Data\Develop\code-for-ibmi\dist\extension.js:39369:11)
	at process.processTicksAndRejections (node:internal/process/task_queues:96:5)
	at async d:\Data\Develop\code-for-ibmi\dist\extension.js:30524:25
	at async s._executeContributedCommand (c:\program files\vscodium\resources\app\out\vs\workbench\api\node\extensionHostProcess.js:96:113532)
	at async s._executeContributedCommand (c:\program files\vscodium\resources\app\out\vs\workbench\api\node\extensionHostProcess.js:96:113532)
```

This was the output when starting debug - seems okay to me...

```
 *  Executing task: npm run webpack 


> code-for-ibmi@1.7.3 webpack
> webpack --mode development

asset extension.js 1.48 MiB [emitted] (name: main) 1 related asset
runtime modules 1.25 KiB 5 modules
cacheable modules 1.43 MiB
  modules by path ./node_modules/ 1.01 MiB
    javascript modules 919 KiB 60 modules
    ./node_modules/@bendera/vscode-webview-elements/dist/bundled.js 116 KiB [built] [code generated]
    ./node_modules/ssh2-streams/package.json 590 bytes [built] [code generated]
  modules by path ./src/ 427 KiB
    modules by path ./src/api/ 200 KiB 20 modules
    modules by path ./src/filesystems/ 33.2 KiB 9 modules
    modules by path ./src/views/ 92.2 KiB 7 modules
    modules by path ./src/webviews/ 41.4 KiB 6 modules
    modules by path ./src/languages/ 13.5 KiB 4 modules
    modules by path ./src/*.ts 37.6 KiB 3 modules
    modules by path ./src/schemas/*.ts 9.03 KiB 2 modules
+ 18 modules
webpack 5.75.0 compiled successfully in 5005 ms
 *  Terminal will be reused by tasks, press any key to close it. 
```

Anything I need to do prior to running the code in the branch?

sebjulliand:
> @sebjulliand @worksofliam I did a quick checkout of Seb's branch to see the function - but on my computer no filters are shown after connection to any system, and the debug console output shows this:
> ...
> Anything I need to do prior to running the code in the branch?

Good catch @chrjorgensen ! Even though it transpiled, the import in the "complex" QSYS FS provider was not using the correct syntax. That complex FS provider is activaed when `Source Dates` are enabled in the connection settings. I did not enable it in my dev env so I didn't face this issue.

But I fixed it, so it should be fine now 😁

chrjorgensen:
@sebjulliand It works now like a charm! 🎉 Very nice!

Some ideas and opinions:
- It would be nice if the protection could be set for the connection, thus applied to all filters for the connection.
- It would be nice if the protection could be enabled/disabled after member is opened - check for protection at save instead of open.
- The error message when trying to save in read only mode suggests "Retry", which does not quite make sense to me:
billede

But you rock! 👏

worksofliam:
Initial review:

* New filter type of 'protect', which then adds a URI fragment to all Uri's below that filter.
* This code is the one actually preventing the edit if it's protected: https://github.com/halcyon-tech/vscode-ibmi/pull/1075/files#diff-60902256ae7cd43976de2ae42faea36561dd9b8a75d327d17a1cf4786fdf3feaR42
* `protected` parameter also added to any function which can generate a 'protected' Uri.

worksofliam (commented):
Initial review:

* New filter type of 'protect', which then adds a URI fragment to all Uri's below that filter.
* This code is the one actually preventing the edit if it's protected: https://github.com/halcyon-tech/vscode-ibmi/pull/1075/files#diff-60902256ae7cd43976de2ae42faea36561dd9b8a75d327d17a1cf4786fdf3feaR42
* protected parameter also added to any function which can generate a 'protected' Uri.

src/views/objectBrowser.js
```diff
@@ -1059,33 +1061,36 @@ module.exports = class objectBrowserTwoProvider {
...
   /**
-   * @param {{name: string, library: string, object: string, types: string[], member: string, memberType: string}} filter
+   * @param {ConnectionConfiguration.ObjectFilters} filter
    */
   constructor(filter) {
     super(filter.name, vscode.TreeItemCollapsibleState.Collapsed);
 
-    this.contextValue = `filter`;
+    this.protected = filter.protected;
+    this.contextValue = `filter${this.protected ? "_readonly" : ""}`;
```
worksofliam, line 1070: This is likely the change with the biggest impact. I understand why the commands in package.json were changed now.

sebjulliand:
> @sebjulliand It works now like a charm! 🎉 Very nice!
> 
> Some ideas and opinions:
> 
>     * It would be nice if the protection could be set for the connection, thus applied to all filters for the connection.
> 
>     * It would be nice if the protection could be enabled/disabled after member is opened - check for protection at save instead of open.
> 
>     * The error message when trying to save in read only mode suggests "Retry", which does not quite make sense to me:
>       billede
> 
> 
> But you rock! 👏

Thanks @chrjorgensen !
The error message comes from VSCode, because I throw an error to prevent the save from happening. But I may be able to display something more user friendly. I'll look into that.

Disabling the protection afterwards may prove to be tricky; I'd have to keep a reference to the filter the member comes from, in the member's URI instead of just setting a "read only" flag in the URI. So this may not be possible at the moment.

A connection level protection is possible though, as I can always get a reference to the connection. I may add that too.

Again, thanks for the feedback!

sebjulliand:
Since I can't say "no" to @chrjorgensen ...
- I added a `Read only mode` to the connection settings, preventing IFS files and source members from being saved, globally.
image
- When a member is opened in read only mode because it's been opened from a protected filter, unprotecting the filter will allow the opened member to be saved.
- Instead of throwing an error that shows a confusing notification, I display more user friendly dialogs when the user attempts to save a read only member
image

- If the connection is in read only mode or the filter is protected, these warning dialogs will offer an option edit the settings.
image
image

On the technical side, I switched from using the URI fragment to using the query part of it. It makes much more sense and there are standards to encode and parse it 🙂

chrjorgensen:
@sebjulliand You are amazing!! 😃 Can't wait to try this - but be careful, I have many wishes! 😉

chrjorgensen:
@sebjulliand Thanks for following up on my suggestions! I have tested your latest commits, and there seems to be only a few quirks:

- You're fine messages are shown when I try to save a member in readonly mode - but the "member dirty" indicator on the member tab title disappears, as if the member is back to normal:
billede
=> 
billede
This could be misinterpreted by the user, who could think this was the original content.
Can we keep the "dirty" indicator?

.vscode/settings.json
```diff
@@ -0,0 +1,2 @@
+{
```
chrjorgensen, line 1: What is the purpose of this file?

chrjorgensen (commented):
Looking good to me - just a few comments...

chrjorgensen:
@sebjulliand I just created PR #1092 and I can see that we both changed the same files in the Object Browser and Search - will be interesting to see if git(hub) can handle this... 🙄

chrjorgensen:
@sebjulliand Another idea that I nearly forgot: It would be nice to have some indicator in the status bar, when the connection is readonly. The filter has a lock before the name when it is protected, but you can't see when the connection is in readonly mode.

.vscode/settings.json
sebjulliand: My mistake!

sebjulliand:
> @sebjulliand Thanks for following up on my suggestions! I have tested your latest commits, and there seems to be only a few quirks:
> 
>     * You're fine messages are shown when I try to save a member in readonly mode - but the "member dirty" indicator on the member tab title disappears, as if the member is back to normal:
>       billede
>       =>
>       billede
>       This could be misinterpreted by the user, who could think this was the original content.
>       Can we keep the "dirty" indicator?

No, I'm afraid this one won't be possible. We have no control over the dirty indicator. The only way to keep it on is to...throw en error and have that confusing error dialog showing up 😅

sebjulliand:
@chrjorgensen there you go:
image

That was a nice idea :)
Of course, changing the setting will update the icon on the fly.

chrjorgensen:
@sebjulliand I knew you could do it! 😆

chrjorgensen:
@sebjulliand And now you probably already know my next request? 😉 
(hint: connection browser...)

sebjulliand:
> @sebjulliand And now you probably already know my next request? 😉 (hint: connection browser...)
> 
> Don't worry, I'll stop after that one...

Humor me 😏

chrjorgensen:
No humor, but you have to guess:

Which one of these connections are readonly?

billede

sebjulliand:
> No humor, but you have to guess:
> 
> Which one of these connections are readonly?
> 
> billede

Well, it's the one with the lock, obviously! 😁
image

Just kidding, good thinking @chrjorgensen . I read "object browser" instead of "connection browser" yesterday's evening, so I was confused. It got better after a good night of sleep!

chrjorgensen:
@sebjulliand Thanks - it looks and works great in my tests.
@worksofliam You're assigned so I leave it up to you to approve the PR...

src/webviews/filters/index.js
```diff
@@ -84,6 +87,11 @@ module.exports = class FiltersUI {
     field.default = filter.memberType || `*`;
     field.description = `Member type. Can be multi-generic value. Examples: <code>RPG*</code> or <code>SQL*LE</code>. A single <code>*</code> will return all member types.`;
     ui.addField(field);
+    
+    field = new Field(`checkbox`, `protected`, `Protected`);
+    field.default = filter.protected ? "checked" : "";
```
worksofliam, line 92:
```suggestion
    field.default = filter.protected ? `checked` : ``;
```

worksofliam (dismissed):
Looks good to me. Please fix the merge conflicts and we should be good to go. Great work

worksofliam:
@sebjulliand I will retest again today with these new changes.

worksofliam (approved):
I fixed the conflicts myself, then fixed a bug I caused, and now it's ready to go. Woo!!!

Sorry this took weeks @sebjulliand.

Please, whenever you have a moment, let's add something about this PR into the docs.

PR #1092, opened 2023-02-23
Search enhancements

chrjorgensen:
### Changes

This PR will enhance searching by
- Sort the search results (IFS search result could previously be mixed)
- Show subdirectories in IFS search result (previously only the basename was shown)
- Show member and type in upper- or lowercase in search results, like names are shown in the Object Browser
- Add a tooltip to the search results, showing the full path of the member/file
- Add a button to collapse all elements in the search result

### Checklist

* [x] have tested my change

### DIFF
--- a/src/views/ifsBrowser.js
+++ b/src/views/ifsBrowser.js
@@ -498,10 +498,10 @@ module.exports = class ifsBrowserProvider {
                   message: `'${searchTerm}' in ${path}.`
                 });
 
-                const results = await Search.searchIFS(instance, path, searchTerm);
+                let results = await Search.searchIFS(instance, path, searchTerm);
 
                 if (results.length > 0) {
-
+                  results = results.map(a => ({...a, label: a.path.replace(`${path}${path.endsWith(`/`) ? `` : `/`}`, ``)}));
                   setSearchResults(searchTerm, results.sort((a, b) => a.path.localeCompare(b.path)));
 
                 } else {
--- a/src/views/ifsBrowser.js
+++ b/src/views/ifsBrowser.js
@@ -470,22 +470,22 @@ module.exports = class ifsBrowserProvider {
 
         if (connection.remoteFeatures.grep) {
 
-          let path;
+          let searchPath;
           if (node)
-            path = node.path;
+            searchPath = node.path;
           else {
-            path = config.homeDirectory;
-            path = await vscode.window.showInputBox({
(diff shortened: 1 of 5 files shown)


src/views/ifsBrowser.js
```diff
@@ -498,11 +498,11 @@ module.exports = class ifsBrowserProvider {
                   message: `'${searchTerm}' in ${path}.`
                 });
 
-                const results = await Search.searchIFS(instance, path, searchTerm);
+                let results = await Search.searchIFS(instance, path, searchTerm);
 
                 if (results.length > 0) {
-
-                  setSearchResults(searchTerm, results);
+                  results = results.map(a => ({...a, label: a.path.replace(`${path}${path.endsWith(`/`) ? `` : `/`}`, ``)}));
```
worksofliam, line 504: I am trying to understand what your `.replace` call is doing. 

Does this just get the basename of the path? Perhaps `path.basename` should be used instead?

worksofliam (commented):
@chrjorgensen

Check out my question in the review. Other than that, looks great!

src/views/ifsBrowser.js
chrjorgensen: The `.replace` call is removing the search start path from the front of the absolute paths returned by `grep`, in order to show found files in subdirectories as relative to the search start path.

`basename` will not show the subdirectories, only the filenames - and what was used before (in the `HitSource` constructor).

worksofliam:
```suggestion
                  results = results.map(a => ({...a, label: a.path.replace(`${path}${path.endsWith(`/`) ? `` : `/`}`, ``)}));
```

Ok, perhaps want we should use here is `path.posix.relative` instead of basename?

```
> path.posix.relative('/home/JGORZINS/projects/manzan/', '/home/JGORZINS/projects/manzan/ile/src/pub_db2.sqlc')
'ile/src/pub_db2.sqlc'
```

worksofliam (commented):
Check out my note about `path.posix.relative`- did some testing and it is a great change! Thank you as always.

src/views/ifsBrowser.js
chrjorgensen: Jeez, I even looked at the path.posix functions, but missed out on the `path.posix.relative` function - fixed...

worksofliam (approved):

PR #1095, opened 2023-02-25
Implemented Connect to Previous connection

sebjulliand:
### Changes
Enabled command `code-for-ibmi.connectPrevious` to actually connects to the IBM i previously connected to.
The last connection name is stored using the new `GlobalStorage` class. The `Storage` class is now an abstraction and the previous `Storage` class has been renamed to `ConnectionStorage`.

Connecting to the previous connection can be achieved by:
- Executing the command from the Command Palette
- Clicking on the "Connect to previous" button in the Connection browser menu
image

If no previous connection exists in `GlobalStorage` or if the previous connection is deleted, the button/command palette action won't be available.

### Checklist
* [x] have tested my change

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "code-for-ibmi",
-	"version": "1.7.3",
+	"version": "1.7.4",
 	"lockfileVersion": 2,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "code-for-ibmi",
-			"version": "1.7.3",
+			"version": "1.7.4",
 			"license": "MIT",
 			"dependencies": {
 				"@bendera/vscode-webview-elements": "^0.6.3",
--- a/package.json
+++ b/package.json
@@ -786,19 +786,19 @@
 		"commands": [
 			{
 				"command": "code-for-ibmi.debug.setup.remote",
-				"title": "Setup remote certificates",
+				"title": "Setup Remote Certificates",
 				"category": "IBM i Debug",
 				"enablement": "code-for-ibmi:connected == true"
 			},
 			{
 				"command": "code-for-ibmi.debug.setup.local",
-				"title": "Import local certificate",
+				"title": "Import Local Certificate",
 				"category": "IBM i Debug",
 				"enablement": "code-for-ibmi:connected == true"
 			},
 			{
 				"command": "code-for-ibmi.debug.start",
-				"title": "Start debug service",
+				"title": "Start Debug Service",
 				"category": "IBM i Debug",
 				"enablement": "code-for-ibmi:connected == true"
 			},
@@ -811,7 +811,7 @@
 			},
 			{
 				"command": "code-for-ibmi.debug.endDebug",
-				"title": "Stop debugging",
+				"title": "Stop Debugging",
 				"category": "IBM i",
 				"icon": "$(debug-disconnect)",
 				"enablement": "code-for-ibmi:connected == true"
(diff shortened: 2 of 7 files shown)


chrjorgensen:
@sebjulliand Very elegant how you use the `commandPalette` and `code-for-ibm-i.hasPreviousConnection` to control when the command is available! 👌 I have been annoyed over all the extension commands, that only worked in a context but still showed up in the command palette - this is how to reduce that clutter...

chrjorgensen:
@sebjulliand Some usability thoughts:
- What was the previous connection? Only the extension knows, it can't be seen on the UI.
- How about presenting the previous connections in a quickpick list - like I did it with `Change current library` - in order of last connected time. Then the user can see the connection history and quickly jump to any of the previous connections. This would also include filtering of the connections in the list. Here is the `Change current library` quickpick:
billede

sebjulliand:
@chrjorgensen Thanks for the feedback!
I think at some point I'll go through all the commands and disable those unavailable from the command palette, depending on the context.

About your other suggestions:
- I'll make the previous connection stand out in the connection browser
- The quickpick is a neat idea! I may remove the button I added to the connection browser in this case, as it would not allow to connect to the previous connection instantly. It would show the quickpick, which would serve the same purpose as clicking the connection in the connection browser 😋 So I'll probably remove it and implement your quickpick idea. Of course, if there's only one connection declared in the list, we'll connect right away.

sebjulliand:
Following @chrjorgensen's feedback:
- `connectPrevious` command has been split in two. There is now `connectTo` (who works as `connectPrevious` did) and `connectToPrevious` who actually connects to the last IBM i connected to.
- `connectToPrevious` is only available from the Connection Browser as a button
- The very last connected system bears a yellow icon
image
- `connectTo` will display a quick pick input to select amongst all the previous connections made, ordered by time of last connection
image
- `connectTo` is bound to the `ctrl/cmd+alt+c` key for quick access.
- Both commands are only available if at least one previous connection is registered and active (i.e. its not been deleted in the meantime).

worksofliam:
Code looks really good from me. I'm out for a couple of days until Wednesday so I can't promise I will be able to test until then.

@chrjorgensen Feel free to give it a go and let me know of your findings based on your previous feedback.

sebjulliand:
@worksofliam you're good to go. I face a "merge from hell" situation so I had to work my way around it but it's all good now 😊

worksofliam (approved):
A simple but welcome UX change. Thanks @sebjulliand! 

@chrjorgensen Do you want the honour of merging?

chrjorgensen:
> @chrjorgensen Do you want the honour of merging?

Yes, please - thank you. 😃

chrjorgensen:
@sebjulliand There is something missing in this change: When switching between systems, the previous connection is not updated correctly, only on first connection after start.

Try this:

1. Start Code for IBM i.
2. Connect to a connection other than the previous connection.
3. Disconnect by clicking the disconnect button in the bar.
4. Refresh the connection browser - the previous connection is updated.
5. Connect to a connection other than the previous connection.
6. Disconnect by clicking the disconnect button in the bar.
7. Refresh the connection browser - the previous connection is NOT updated!

sebjulliand:
Right, I'll make a new PR. Turns out this one was created and merged before the "Disconnect gracefully" one was created an merged. 😅

sebjulliand:
Oh and actually the `connected` event doesn't get triggerred when reconnecting (it's in the same place in the code).
I'll fix that too!

chrjorgensen:
Ah, good find! Yeah, I thought so - that some changes had collided. I knew you wouldn't have let this pass...  😆

PR #1112, opened 2023-03-03
Default deploy directory on input

worksofliam:
### Changes

When calling the command to set the deploy directory without a path, set the default in the user input.

### Checklist

* [ ] have tested my change

### DIFF
--- a/src/api/local/deployment.ts
+++ b/src/api/local/deployment.ts
@@ -479,24 +479,28 @@ export namespace Deployment {
   }
 
   async function setDeployLocation(node: any, workspaceFolder?: WorkspaceFolder) {
-    const path = node?.path || await vscode.window.showInputBox({
+    const connection = instance.getConnection();
+    const possibleDeployDir = path.posix.join(`/`, `home`, connection!.currentUser, `builds`, workspaceFolder ? workspaceFolder.name : `codebase`);
+
+    const deployDir = node?.path || await vscode.window.showInputBox({
       prompt: `Enter IFS directory to deploy to`,
+      value: possibleDeployDir
     });
 
-    if (path) {
+    if (deployDir) {
       const storage = instance.getStorage();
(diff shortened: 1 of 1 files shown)


src/api/local/deployment.ts
```diff
@@ -479,24 +479,28 @@ export namespace Deployment {
   }
 
   async function setDeployLocation(node: any, workspaceFolder?: WorkspaceFolder) {
-    const path = node?.path || await vscode.window.showInputBox({
+    const connection = instance.getConnection();
+    const possibleDeployDir = path.posix.join(`/`, `home`, connection!.currentUser, `builds`, workspaceFolder ? workspaceFolder.name : `codebase`);
```
sebjulliand, line 483: How about using the actual and/or configured home directory?
```suggestion
    const possibleDeployDir = path.posix.join(connection?.config?.homeDirectory || `/home`, connection!.currentUser, `builds`, workspaceFolder ? workspaceFolder.name : `codebase`);
```

sebjulliand (changes requested):

src/api/local/deployment.ts
worksofliam: @sebjulliand Yes. The reason we build the path is because `config.homeDirectory` is not actually the home directory all the time. It acts more like the 'working directory', and we don't want to build path to be based on the working directory.

sebjulliand: Sounds about right.
Now, call me paranoid, but I'd run `pwd` or `echo ~` and get the output here to get the actual home directory. I know most of the time `/home/<username>` is the actual home dir but...better safe than sorry? 😁

worksofliam:
Next step for this PR is to determine the best way to get the user's home directory. Converting to draft for now.

worksofliam:
Looks like Seb actually solved in the tar re-write.

Issue #1098, opened 2023-02-28
Initial Command to set environment i.e. library list by profile


JDubb-At-Nelnet:
**Is your feature request related to a problem? Please describe.**
Problem - we would like to use a CL command supplied by LMi - ACMSLIB/ACMSLIBL - in order to set the library list upon successful connection.  A chglibl command would also be possible, as well as calling a CL program that sets the users library list using traditional library list cl commands.  Would like this to be tied to the profile - as in a separate command definition could be specified for each profile, so that when a profile changes the command for that profile runs.  This would allow easy rollout of library list changes to all developers using the tool.

**Describe the solution you'd like**
In the library list explorer, provide a new button (Initial Command) that pops open an initial command entry dialogue box.  The command would be saved to the current profile, and modifying the command after the fact would be possible via clicking the same 'initial command' button.  The initial command would run any time a new connection is made, or the user switches profile since the command is tied to individual profiles.  Any libraries in the library list set up in the users job description, or in the library list tool itself could potentially be overwritten by the initial command.

**Describe alternatives you've considered**
Tried setting a library list in .bashrc using 'cl: ACMSLIB/ACMSLIBL ....' but that doesn't work.

**Additional Context**
Other tools (rdi) allow this.

worksofliam:
I have been trying things and I've come up with this conclusion:

We need to write a new program that takes in a user defined command, for example:

```
PRINTLIBL CMD(CHGLIBL LIBL(COOLLIB DEVLIB PRDLIB) CURLIB(WHATEVER))
```

This would

* run the provided command, which would change the library list of the job
* `PRINTLIBL` would then write the user portion and current library of the job to standard out
* Code for IBM i would read standard out and update to reflect the changes

It would be better if we can determine if this is possible entirely from a shell script, which would save a lot of effort. Will ponder that.

worksofliam:
QSH `liblist` does not listen to changes after jobs are run.

worksofliam:
Another idea, likely easier to implement:

SQL procedure that uses `QCMDEXC`(?) to call the provided change libl command and then returns the new library list as a result set.

worksofliam:
This looks promising. Next step is to somehow integrate this with profiles.

<img width="1084" alt="image" src="https://user-images.githubusercontent.com/3708366/224129563-a41ebfba-f981-
44d0-918f-29b13693e222.png">

worksofliam:
Introducing.... Command Profiles

[...]

PR #1123, opened 2023-03-09
Command Profiles

worksofliam:
### Changes

To address #1098, I have come up with a way to set the library list based on a command. This PR adds a few things:

* New backend procedure to run a command and return the resulting library list
   * This gets installed onto the system if it does not exist
   * New API will get exposed through IBMiContent!
* Adds a new type of profile, named 'Command Profile'
   * a command profile only has two attributes, a name and a command
   * Unlike regular profiles, they only update the library list and the current library
* Command Profiles show up in the Profiles view with a different icon
   * Command Profiles also have a simple edit UI

### To do

* [ ] Documentation needs to be updated after this PR has been approved, but before it gets merged.

### Checklist

* [x] have tested my change


chrjorgensen:
@worksofliam I will review this asap, hopefully tomorrow...

chrjorgensen (changes requested):
@worksofliam LGTM and working as expected. 👌 

Only thing that bothers me is that after creating a command profile and running it, there is no way to get back to the previous setup of the library list - not even deleting the command profile works, the library list after reconnect or restart is still what the command profile returned.

What if the current library list is saved into a profile `Previous setup` when creating the first command profile? That way the user can get back to the previous setup before starting to use command profiles. I think many users will try this out, not realizing that their current setup will be gone...

worksofliam:
@chrjorgensen Good point. Perhaps there should be a default profile named 'default', which will just reset the user library list + current library back to the base.

I will do that.

worksofliam:
@chrjorgensen I have implemented a base profile that will reset some of the connection attributed. Let me know what you think.

worksofliam:
@chrjorgensen 

> How do one activate the Default (reset to Default Profile)? Nothing happens on click and there is no arrow like on the other profiles.

As shown in the image below, I did not set an icon for this button on the 'Default' item. This is because I want used to explicitly see at a glance that it will reset unlike the other buttons to switch profile. This is also only shown on hover.

image

> Should the object filters and IFS shortcuts be reset (i.e. removed) when the default profile is activated?

In my opinion, yes. As you know, object filters and IFS shortcuts can be associated with filters and I personally tend to find that when I want to create a new profile, I don't have many (if any) of the old filters as part of it.

Also, when actually resetting back to the default, we make sure the user is aware of this change by warning them:  

image

Please let me know what improvements I can make here. Perhaps before resetting back to the default, we should ask the user if they want to save into their current profile too?

chrjorgensen (commented):
@worksofliam I just realized that the `Reset to Default Profile` is a button...! 😄 

It was not obvious to me since this is text and the other profiles have an icon... I would expect to have an icon to active the reset, just as the other profiles. Maybe you could change the name of the profile from `Default` to `Reset to Default Profile` and add the same icon to activate as for the other profiles? This would not be dangerous for the user, since he/she has to confirm the operation in a modal pop-up window.

I do not agree on your view on the reset of library list and filters - I think it depends on how one has used Code for IBM i so far. I have not used `Profiles` so far, but use the same library list almost all the time and have this setup to my preference, probably since I don't work on systems with more than one application or customer. Now, if I saw this new fancy button for creating `Command Profiles` and wanted to give it a shot, I would expect to have some way to come back to my preference - not to have to recreate them. But as it is now, I don't see how I can restore my preference except for modifying library list after my memory?

chrjorgensen:
Let's get the opinion of @sebjulliand  - maybe we should have a referendum? 😄

worksofliam:
> Maybe you could change the name of the profile from Default to Reset to Default Profile and add the same icon to activate

This is a better idea! I will do this next.

> But as it is now, I don't see how I can restore my preference except for modifying library list after my memory?

Correct. The user needs to ensure that they previously save their settings, which is why we warn them that it will clear their existing settings.

chrjorgensen:
> This is a better idea! I will do this next.

Great, I will test this asap.

> Correct. The user needs to ensure that they previously save their settings, which is why we warn them that it will clear their existing settings.

But there is no warning when the icon to create a command profile is clicked and a command profile is saved. Only when the user wants to come back to the previous setting, there is a warning that library list and all filters will be reset and their longtime setup lost! That would annoy me very much as a user.

I think the problem is that `Command Profiles` is a term from the `Profiles` functionality - but it appears in the library list view, together with the `Save Current Settings to New Profile` button. As a user, I would expect any buttons in the library list view to affect only the library list (which `Command Profiles` do) and NOT affect my filters.

I've never really become 'friends' with Profiles and have not really used them. I could use the `Command Profile` functionality to make predefined library lists, but I would like to have my filters stay the same whatever library list was set.

worksofliam:
@chrjorgensen 

> Maybe the confusion could be reduced by only having this button on the Profiles view and not in the Library List view

This button will no longer show on the Library List view.

> Before I continue my test, please fix the following error, that I just discovered

You can now use the Enter key or the Save button to save the Command Profile. Cancel will cancel as expected.

chrjorgensen (changes requested):
@worksofliam This is so much better! 👌 

`Profiles` view now first appears when the user saves the current library list to Default.
`Command Profile` button appears only in `Profiles` view and the user thus has an escape route.
`Create/edit Command Profile` now saves on Enter.

Now there's only one issue before I can approve: There is a check for command profile name is entered, but there is no check for any command is entered!

When this is fixed I can approve the PR.

worksofliam (commented):
Ready for another review!

chrjorgensen (approved):
@worksofliam It works and I approve and merge the code.

Some suggestions for next time you're bored 😉:
1. Keep the view until everything is okay and commend profile is created, or the user presses `Cancel`. As now, an error message appears but the view is gone, and the user has to start over.
2. Make the command `Create/Edit command profile` unavailable in the Command Palette until Profiles view is active. You removed the button from the Library List view, but the command is still available.

Issue #1074, opened 2023-02-17
Default shell configuration


worksofliam:
Of course the recommended shell on IBM i is bash.

Perhaps we need:

1. a command to allows the user to change their default shell to bash
2. check the default shell when the user connects to the system, prompt if it's a dumb shell

These checks would happen in `IBMi#connect` (inside of `IBMi.ts`)

beckhamrryyaann:
This looks fun.  I will try to implement this.

beckhamrryyaann:
@worksofliam Hi,  Let me know if this is what you are expecting:  Please advise.
In the connect method of IBMi.ts
* It appears I can use the `sendCommand()` property to send commands to system.  I can run the command `echo $SHELL` and that will give me path of users current shell. 
*  If the value is equal to `/QOpenSys/pkgs/bin/bash` than I know they are using the correct shell.
*   If not, I send show Alert dialog that lets them press 'Yes'.
*  If Yes is pressed, then I can send the sql command, `CALL QSYS2.SET_PASE_SHELL_INFO('*DEFAULT', '/QOpenSys/pkgs/bin/bash')` to set their shell to use bash.

Thanks
Ryan

worksofliam:
@beckhamrryyaann that sounds about right! Thanks for looking into this. I am looking forward to seeing your PR.

beckhamrryyaann:
I about had the pr ready but got to thinking I just can’t assume bash is installed on system So i am working through that.

chrjorgensen:
@beckhamrryyaann Closed by #1134

worksofliam:
@ThomasBarberot Is your environment setup as defined in the CCSID docs? https://halcyon-tech.github.io/docs/#/pages/tips/ccsid

ThomasBarberot:
Yes, that why I'm surprised.
- The system value QCCSID is set to 1147
- My user profil is set to CCSID(*SYSVAL)
- The system level environment variable PASE_LANG is set to FR_FR
- The system level environment variable QIBM_PASE_CCSID is set to 1208
When I change the default QSH to bash, there is an issue, and when I reset it to the default value, all working good.

chrjorgensen:
@ThomasBarberot I don't see how `bash` could affect the download and upload of members - it's just a PASE program like the default `bsh` shell program and doesn't change the settings of the PASE environment.

[...]

PR #1134, opened 2023-03-14
Give user option to set bash as their default shell when connecting.

beckhamrryyaann:
### Changes

Adds feature described in #1074.  Give user option to set bash as their default shell. 

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -573,37 +573,40 @@ export default class IBMi {
           }
 
           // Check users default shell.
-          // If not bash and bash is installed, give user option to set bash as default shell.
+          // give user option to set bash as default shell.
           try {
-            const bashShellPath = '/QOpenSys/pkgs/bin/bash';
-            const commandShellResult = await this.sendCommand({
-              command: `echo $SHELL`
-            });
-            if (!commandShellResult.stderr) {
-              let userDefaultShell = commandShellResult.stdout.trim();
-              if (userDefaultShell === bashShellPath &&
-                  this.remoteFeatures['bash']) {
-                vscode.window.showInformationMessage(`IBM recommends using bash as your default shell.`, `Set shell to bash?`, `Read More`,).then(async choice => {
-                  switch (choice) { 
-                    case `Set shell to bash?`:
-                      statement = `CALL QSYS2.SET_PASE_SHELL_INFO('*CURRENT', '/QOpenSys/pkgs/bin/bash')`;
-                      output = await this.sendCommand({
-                        command: `LC_ALL=EN_US.UTF-8 system "call QSYS/QZDFMDB2 PARM('-d' '-i')"`,
-                        stdin: statement
-                      });
+            // make sure sql is enabled and bash is installed on system
+            if (this.config.enableSQL && 
+                this.remoteFeatures[`bash`]) {
+              const bashShellPath = '/QOpenSys/pkgs/bin/bash';
+              const commandShellResult = await this.sendCommand({
+                command: `echo $SHELL`
+              });
+              if (!commandShellResult.stderr) {
+                let userDefaultShell = commandShellResult.stdout.trim();
+                if (userDefaultShell !== bashShellPath) {
+                  vscode.window.showInformationMessage(`IBM recommends using bash as your default shell.`, `Set shell to bash?`, `Read More`,).then(async choice => {
+                    switch (choice) { 
+                      case `Set shell to bash?`:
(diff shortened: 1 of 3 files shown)


beckhamrryyaann:
The initial PR does not attempt to install bash on the users system if not installed.  The code just gracefully continues on with the rest of the connection logic.  Please advise if this is not acceptable behavior.

src/api/IBMi.ts
```diff
@@ -570,6 +571,45 @@ export default class IBMi {
...
+          // Check users default shell.
+          // If not bash and bash is installed, give user option to set bash as default shell.
+          try {
+            const bashShellPath = '/QOpenSys/pkgs/bin/bash';
+            const commandShellResult = await this.sendCommand({
+              command: `echo $SHELL`
+            });
+            if (!commandShellResult.stderr) {
+              let userDefaultShell = commandShellResult.stdout.trim();
+              if (userDefaultShell === bashShellPath &&
```
chrjorgensen, line 584: This should be `userDefaultShell !== bashShellPath` instead of using `===`!
Now it prompts when my shell is bash already... 😉

src/api/IBMi.ts
```diff
@@ -570,6 +571,45 @@ export default class IBMi {
             // Oh well!
             console.log(e);
           }
+
+          // Check users default shell.
+          // If not bash and bash is installed, give user option to set bash as default shell.
+          try {
```
chrjorgensen, line 577: You don't consider the connection property called `enableSQL`... your check for bash should only be done if SQL is enabled, since you use SQL to change the shell for the user.

chrjorgensen (changes requested):
@beckhamrryyaann Thanks for making this PR - just a few issues as seen above.

beckhamrryyaann:
Oh my !  How embarrassing.  I changed it to === during testing so I could
see the message without having to change my default shell over and over on
my system.   I will makes these corrections hopefully tonight and resubmit
the PR.

src/api/IBMi.ts
```diff
@@ -570,6 +571,45 @@ export default class IBMi {
...
+                  switch (choice) { 
+                    case `Set shell to bash?`:
+                      statement = `CALL QSYS2.SET_PASE_SHELL_INFO('*CURRENT', '/QOpenSys/pkgs/bin/bash')`;
+                      output = await this.sendCommand({
+                        command: `LC_ALL=EN_US.UTF-8 system "call QSYS/QZDFMDB2 PARM('-d' '-i')"`,
+                        stdin: statement
+                      });
+
+                      if (output.stdout) {
+                        vscode.window.showInformationMessage(`Default shell in now bash!`);
```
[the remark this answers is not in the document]
beckhamrryyaann: Latest commit fixes

src/api/IBMi.ts
beckhamrryyaann: Latest commit fixes.

src/api/IBMi.ts
beckhamrryyaann: Latest commit address issue.  First thing the enhancement does is check to make sure bash is installed and also that sql is enabled.  If both are not true, the enhancement code is skipped.  No calls to server are made so should be more efficient.

chrjorgensen: It's working now and I'll approve the changes. If you're up to it, you can make a new PR and use the `chsh` (if installed) to change the user shell instead of using SQL - as described in https://ibmi-oss-docs.readthedocs.io/en/latest/user_setup/README.html#step-4-change-your-default-shell-to-bash. That way you're not dependent on whether SQL is enabled or not...

chrjorgensen (approved):

chrjorgensen:
@beckhamrryyaann LGTM - approved and will merge. Thanks for contributing!

src/api/IBMi.ts
beckhamrryyaann: @chrjorgensen I'm definitely up for it.  How do you feel if I include both methods (sql or chsh) and use which ever one is usable.  I would default to using sql first like I have, but if sql is disabled and chsh is available I could use that as a fall back?  Are you ok with this?

chrjorgensen: @beckhamrryyaann I definitely think you should go for `chsh` and drop SQL - then you can ignore the `enableSQL` setting. Youjust have to check for `chsh` instead... 😜 

Please make a new PR - this one has been merged already...

beckhamrryyaann: Ok.   That sounds good.  I will try to get that this week.

PR #1136, opened 2023-03-14
Fix library list validation: find all invalid libraries

chrjorgensen:
### Changes

This PR will fix validation of the library list. Previously it would stop after first invalid library, but this fix will find all the invalid libraries in the list - as was intended...

### Checklist

* [x] have tested my change


worksofliam (approved):
LGTM! Code is beautiful (can't go wrong with one-liners!) and runs as expected.

I am assigning to @sebjulliand to determine when is best to merge.

PR #1145, opened 2023-03-19
Initial work for l10n - views

worksofliam:
### Changes

Shows off a possible way to support l10n. I am using a custom implementation instead of the VS Code l10n API due to the fact the configuration files use the base message as the base text and the id for the locale in other languages. This uses just an ID and then falls back to the default (en) if cannot be found in the target locale.

I made that decision based on this thread: https://github.com/microsoft/vscode-discussions/discussions/517

This will not handle the `package.json` and we will have to use the VS Code method to do that (perhaps a separate PR also.)

Looking for feedback on:

* The interfaces
* The `t` method


chrjorgensen:
@worksofliam @sebjulliand
I started translating texts to my own locale Danish, thus beginning the journey to have C4i fully support l10n.

The `t` function works fine - I just changed the `replace()` with RegExp to `replaceAll()` (had to upgrade from ES2019 to ES2021).

What about panel texts? Menu items? Command names? Tooltips?
There are texts in many places other than in the code. Do we convert all? And how?

worksofliam:
@chrjorgensen 

Great works! And nice cleaning. Not so bad moving to a new version of ES2019.

> Do we convert all? And how?

Yeah.. this is going to be a challenge. It wouldn't be the end of the world if we missed some. A good way to get more contributors I think 😄

chrjorgensen:
@worksofliam @sebjulliand
I changed the locale to a configuration setting, thus able to select a different language not related to VS Code. If no locale has been selected, it will use the env language as before. The locale is set only once and change will require restart.
Any comments on this? Good? Bad? Anything missed? Any opinions are welcome...

chrjorgensen:
@sebjulliand In this work all files containing texts will have to be changed, including the JavaScript files. You're planning to convert these into TypeScript - will we collide on this? How do we avoid most collisions?

This localization project is a massive task that will take a long time to finish. And I'm off on vacation after next week. What are your thoughts on this?

chrjorgensen:
@sebjulliand Nice change in [b41cc5a](https://github.com/halcyon-tech/vscode-ibmi/pull/1145/commits/b41cc5a7650165d7f5e3e19e6eaedd4b34e16e20) - thanks! 😃

chrjorgensen:
@worksofliam @sebjulliand All views have now been localized - next up are/would be the webviews.
It would be great with a review and I have changed the PR from draft to ready for review...

How far should we go with this before merging into master branch? The longer we wait the more likely it is to have collisions for this PR or other PR's in the works.

src/locale/index.ts
```diff
@@ -0,0 +1,26 @@
...
+import { en } from "./ids/en";
+
+const locales: {[locale: string]: Locale} = {
+  en,
+  da
+}
+
+export type Locale = {[id: string]: string};
+
+const currentLocale = String(GlobalConfiguration.get(`locale`) || env.language);
```
worksofliam, line 13: We really should be letting VS Code decide what the locale is. This will always default to English.

worksofliam:
@chrjorgensen I added some code to make the default locale whatever VS Code is using, and then also added a handler so when the config changes it updates our internal locale reference.

@sebjulliand Over to you for French!

sebjulliand:
french-it-up

worksofliam:
@chrjorgensen merge when you feel comfortable…

src/locale/ids/fr.ts
```diff
@@ -0,0 +1,184 @@
+import { Locale } from "..";
+
+export const en: Locale = {
```
chrjorgensen, line 3:
```suggestion
export const fr: Locale = {
```

@sebjulliand You forgot to change the locale code from `en` to `fr`.

You must also add the `fr` locale code to the files `src/locale/index.ts` and `package.json`...

chrjorgensen (changes requested):

src/locale/ids/fr.ts
sebjulliand: Ugh, I guess I was way to excited to ship the translations 😅

chrjorgensen (approved):
@sebjulliand Perfect - approving and merging.

PR #1146, opened 2023-03-20
Make webpanels retain their context

sebjulliand:
### Changes
Set panels' `retainContextWhenHidden` property to `true` to make them retain there context.
With this, if the view looses the focus, it will retain its context (instead of being reset).

Webpanels will now also be opened in the main section since there is no risk of loosing the context when switching tabs.

### Checklist
* [x] have tested my change


chrjorgensen (commented):
@sebjulliand LGTM - a simple, but very nice change to have the custom UI panels open up in the main window instead of beside the editors and in an often narrow panel! 👌 

@worksofliam WDYT? I will make it up to you to review and approve this PR...

worksofliam:
I very much love the fullscreen due to the fact I added some CSS to add a max width! It looks so cool now. 

image

worksofliam (approved):
Looks good to me! Tested and works as expected. Thanks!

Issue #1122, opened 2023-03-09
Replace SQL usage of fetching system values with QSH `sysval`


worksofliam:
https://www.ibm.com/docs/en/i/7.3?topic=utilities-sysval

worksofliam:
@chrjorgensen I know you helped work on this in the `IBMi.ts` connection process. What do you think?

chrjorgensen:
I think we should use the fastest method, be it QSHELL `sysval` or SQL.

I already have some thoughts about the connection process and all the calls made to the server before the connection is complete. This takes quite some time for me and is a bit annoying. So I've done some testing measuring the time used per call and are considering how to reduce the number of calls by combining the requests. Nothing is ready yet - and maybe never will... 😆

worksofliam:
@chrjorgensen I wonder if we could actually do some of the optional feature checking after the connection has been established. 

E.g. 

1. find out all the import stuff first (SQL, CCSID and data areas come to mind), then load the extension
2. let the user do stuff while the backend determines the remainder of the optional functionality

Likely another big change on top of all our PRs right now :joy:

I was also just thinking tonight that is becoming a little long, even for fast connections

chrjorgensen:
@worksofliam My idea is to store the server settings, which are rarely changed, in some cache store in the client, and have a `Quick Connect` checkbox in the connection settings, that when checked will use the values in the cache (if available) instead of going to the server. The user can then remove the checkmark to refresh or force the client to get the newest values from the server.

Values that could be stored in cache are (in no particular order)
- bad data areas
- SQL
- CCSID
- encoding values for #@$
- ASP info
- all remote features
- temp library creation
- temp library features creation, i.e. GENCMDXML and GETNEWLIBL

The cache would be written after each connect to be available at next connection, whether the `Quick Connect` was selected or not. Then the user could just check it to have faster connections in the future.

Other stuff that could be run in the background on the server
- cleaning temp library
- cleaning temp directory

That would reduce connection time considerably...

PR #1150, opened 2023-03-21
Add Quick Connect

chrjorgensen:
### Changes

This PR will add a `Quick Connect` setting in the connection configuration to speed up connecting to the system:

billede

When selected, the following server settings are taken from the previous connection to the system:
- bad data areas
- all remote features
- ASP info
- CCSID
- encoding values for #@$

These values are stored in connection storage on each connection to the system. So if no previous connection has been done, even though the `Quick Connect` is checked, all settings will be retrieved from the server and stored in the connection configuration to be used on the next connection.

This setting is active by default. The user can disable it permanently in the configuration, or the user can reload all settings from server into cache by right-clicking on the connection and select `Connect and Reload Server Settings`.

If the remote feature keys do not match the keys in the cache, the remote features will be checked on the server disregarding this setting.

For more information, see issue #1122.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/Configuration.ts
+++ b/src/api/Configuration.ts
@@ -45,6 +45,7 @@ export namespace ConnectionConfiguration {
     debugUpdateProductionFiles: boolean;
     debugEnableDebugTracing: boolean;
     readOnlyMode: boolean;
+    quickConnect: boolean;
     [name: string]: any;
   }
 
@@ -119,7 +120,8 @@ export namespace ConnectionConfiguration {
       debugIsSecure: (parameters.debugIsSecure === true),
       debugUpdateProductionFiles: (parameters.debugUpdateProductionFiles === true),
       debugEnableDebugTracing: (parameters.debugEnableDebugTracing === true),
-      readOnlyMode: (parameters.readOnlyMode === true)
+      readOnlyMode: (parameters.readOnlyMode === true),
+      quickConnect: (parameters.quickConnect === true)
     }
   }
 
--- a/src/api/Configuration.ts
+++ b/src/api/Configuration.ts
@@ -121,7 +121,7 @@ export namespace ConnectionConfiguration {
       debugUpdateProductionFiles: (parameters.debugUpdateProductionFiles === true),
       debugEnableDebugTracing: (parameters.debugEnableDebugTracing === true),
       readOnlyMode: (parameters.readOnlyMode === true),
-      quickConnect: (parameters.quickConnect === true)
+      quickConnect: (parameters.quickConnect === true || parameters.quickConnect === undefined)
     }
   }
 
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -88,7 +88,9 @@ export default class IBMi {
       'GENCMDXML.PGM': undefined,
       'GETNEWLIBL.PGM': undefined,
       'QZDFMDB2.PGM': undefined,
-      'startDebugService.sh': undefined
+      'startDebugService.sh': undefined,
+      attr: undefined,
+      iconv: undefined
     };
 
     this.variantChars = {
@@ -448,7 +450,8 @@ export default class IBMi {
         }
 
         // Check for installed components?
-        if (quickConnect === true && cachedServerSettings?.remoteFeatures) {
+        // For Quick Connect to work here, 'remoteFeatures' MUST have all features defined and no new properties may be added!
+        if (quickConnect === true && cachedServerSettings?.remoteFeatures && Object.keys(cachedServerSettings.remoteFeatures).sort().toString() === Object.keys(this.remoteFeatures).sort().toString()) {
           this.remoteFeatures = cachedServerSettings.remoteFeatures;
         } else {
           progress.report({
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -9,6 +9,8 @@ import { CompileTools } from "./CompileTools";
 import { ConnectionData, CommandData, StandardIO, CommandResult, IBMiMember, RemoteCommand } from "../typings";
 import * as configVars from './configVars';
 import { instance } from "../instantiate";
+import IBMiContent from "./IBMiContent";
+import { GlobalStorage, CachedServerSettings } from './Storage';
 
 export interface MemberParts extends IBMiMember {
   basename: string
@@ -160,6 +162,9 @@ export default class IBMi {
 
         //Load existing config
         this.config = await ConnectionConfiguration.load(this.currentConnectionName);
+
+        // Load cached server settings.
+        const cachedServerSettings: CachedServerSettings = GlobalStorage.get().getServerSettingsCache(this.currentConnectionName);
 
         progress.report({
           message: `Checking home directory.`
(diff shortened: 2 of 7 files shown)


worksofliam:
@chrjorgensen This is awesome. I personally love that this is an opt in thing. Please fix up the conflicts and then I will retest again.

sebjulliand (changes requested):
That's an awesome change and should indeed speed up connection times. Great job!

Maybe you can make a change to enhance the UX a bit more. Having to uncheck the property in the settings, connect and check again the property seems a bit cumbersome. How about having a `Connect and reload settings` command that could force the reloading of the settings even if quick connect is on? This could be a right click action in the connection browser.

chrjorgensen:
@sebjulliand I now added a connection right-click option to connect and reload server settings:

billede

A new command `ConnectToAndReload` has been created, but is not available in the command palette, only on the right-click menu.

I had to add one parameter to the connect functions to specify that server settings should be reloaded, despite quick connect was set for the connection. Could I have done it better in another way?

src/api/Storage.ts
```diff
@@ -4,7 +4,8 @@ const PREVIOUS_CUR_LIBS_KEY = `prevCurLibs`;
 const LAST_PROFILE_KEY = `currentProfile`;
 const SOURCE_LIST_KEY = `sourceList`;
 const DEPLOYMENT_KEY = `deployment`;
-const DEBUG_KEY = `debug`
+const DEBUG_KEY = `debug`;
+const SERVER_SETTINGS_CACHE_KEY = `serverSettingsCache`;
```
sebjulliand, line 8: Since your setting's key depends on a name, make this a function and use it later on, so you don't repeat the key contruction code over and over.
```suggestion
const SERVER_SETTINGS_CACHE_KEY = (name : string) => `serverSettingsCache_${name}`;
```

src/api/Storage.ts
```diff
@@ -73,7 +82,18 @@ export class GlobalStorage extends Storage {
   async setLastConnections(lastConnections: LastConnection[]) {
     await this.set("lastConnections", lastConnections.sort((c1, c2) => c2.timestamp - c1.timestamp));
   }
-}
+
+  getServerSettingsCache(name: string) {
+    return this.get<CachedServerSettings>(SERVER_SETTINGS_CACHE_KEY.concat(`_${name}`));
```
sebjulliand, line 87: Continued from suggestion's above:
```suggestion
    return this.get<CachedServerSettings>(SERVER_SETTINGS_CACHE_KEY(name));
```

src/api/Storage.ts
```diff
@@ -73,7 +82,18 @@ export class GlobalStorage extends Storage {
...
+  }
+
+  async setServerSettingsCache(name: string, serverSettings: CachedServerSettings) {
+    await this.set(SERVER_SETTINGS_CACHE_KEY.concat(`_${name}`), serverSettings);
```
sebjulliand, line 91:
```suggestion
    await this.set(SERVER_SETTINGS_CACHE_KEY(name), serverSettings);
```

src/api/Storage.ts
```diff
@@ -73,7 +82,18 @@ export class GlobalStorage extends Storage {
...
+  }
+
+  async deleteServerSettingsCache(name: string) {
+    await this.set(SERVER_SETTINGS_CACHE_KEY.concat(`_${name}`), undefined);
```
sebjulliand, line 95:
```suggestion
    await this.set(SERVER_SETTINGS_CACHE_KEY(name)), undefined);
```

sebjulliand (approved):
It's all good to me; just a suggestion to consider.

sebjulliand:
> I had to add one parameter to the connect functions to specify that server settings should be reloaded, despite quick connect was set for the connection. Could I have done it better in another way?

No, that's perfect the way you did it @chrjorgensen !

sebjulliand:
@chrjorgensen, @worksofliam : shouldn't `Quick connect` actually be turned on by default when creating a connection? I'd tend to think so. What do you guys think?

worksofliam:
@sebjulliand I am guessing this quick connect would only work after the initial connection so we have some system information.

I haven't tested this PR yet

sebjulliand:
> @sebjulliand I am guessing this quick connect would only work after the initial connection so we have some system information.

If it's on but nothing's been cached yet, it will perform a normal connection to gather and cache the values. That's why I think it doesn't hurt to have it turned `on` by default.

chrjorgensen:
Yeah, Quick connect will not break if activated on a connection never used before. For each setting it will check if there's a value in the cache from a previous connection and only then use that value. Otherwise it will go to the server as usual.
So having this on by default will not break anything.

I'm undecided whether this setting is best suited as a connection setting or a global setting. Since it won't break anything and servers very, very rarely change settings, this could just as well be a global setting. What do you guys think?

sebjulliand:
When you put it this way, I think it should even not be an option at all, but just the built-in behavior. I can't think of a scenario where you'd need to reload the cached settings every time. Since we have the command to reload them on demand, there may be no point to have it as an option at all.

worksofliam:
> When you put it this way, I think it should even not be an option at all, but just the built-in behavior.

I somewhat agree with it. In fact, I was thinking an option like 'always do a full connect' (which means never use our cached settings) in the VS Code settings would work here. If it's false, it will: use the cached settings if they exist, otherwise do full connect and store them. If it's true.. always do a full connect

chrjorgensen:
@sebjulliand This PR was so far from master that I had to rebase the changes onto master branch and force push.
I also added a check for remote features in the cache, since new features have been and will be added in the future, and then the cache is not valid.

src/api/IBMi.ts
```diff
@@ -375,209 +383,230 @@ export default class IBMi {
...
-                    vscode.window.showInformationMessage(`Failed to delete the data area QSYS/QCPFRMIMPF. Code for IBM i may not work as intended.`);
-                  });
-                break;
-              case `Read more`:
-                vscode.env.openExternal(vscode.Uri.parse(`https://github.com/halcyon-tech/vscode-ibmi/issues/476#issuecomment-1018908018`));
-                break;
-            }
+        // Check for bad data areas?
+        if (quickConnect === true && cachedServerSettings?.badDataAreasChecked === true) {
+        } else {
```
sebjulliand, line 388: Having an empty block may be confusing, don't you think?
```suggestion
        if (!quickConnect || !cachedServerSettings?.badDataAreasChecked) {
```

sebjulliand (changes requested):
Just a small suggestion; I can't wait to merge this PR, it's so convenient!

src/api/IBMi.ts
chrjorgensen: @sebjulliand I think it's in the eyes of the beholder - to me it's more confusing with two NOT's and an OR... and I would prefer to keep the test for Quick Connect the same in all places. But I will add a comment inside the empty block to clarify it's empty on purpose.

chrjorgensen:
@sebjulliand I clarified the empty code block - and had to do a new rebase and force-push.

I also made the Quick Connect active by default, per our discussion, and clarified how to reload the server settings in the configuration description.

Looking forward to your review! 😃

chrjorgensen:
@sebjulliand @worksofliam I agree that the Quick Connect option could be removed and just be the built-in behavior, and this could very well be done in the near future. But I'm a careful guy and like the user to have an option to disable this new functionality, if it somehow would create some kind of problem that we have not yet encountered. Running it as an option for a while is more secure and less disturbant for the user.

sebjulliand (changes requested):
Everything is fine, save for one detail; it's not directly related to your PR, but it's been highlighted by it 😉

In `src/webviews/settings/index.ts`, the target configuration doesn't get loaded and saved in all cases using the `ConnectionConfiguration.load` and `ConnectionConfiguration.update` method. Since `ConnectionConfiguration.load` isn't used to load the configuration, some default values are not applied for undefined settings, like `Quick Connect`. Consequently, when you edit the connection settings for a connection for which `Quick Connect` is undefined, the checkbox is unchecked. But when the settings will be loaded during the connection, it will actually be `true`. So the settings page is not the exact reflect of the actual settings in this case 😉

Using `ConnectionConfiguration`'s `load` and `update` will also simplify the connection settings page code a little bit and remove some redundant code.

chrjorgensen:
@sebjulliand I see the problem and is on it...

chrjorgensen:
@sebjulliand The initialization problem in settings view is now fixed - please test thoroughly, since this is not a trivial code rewrite.

worksofliam:
@chrjorgensen I had a quick review and I think the whole idea of restarting can be removed now - you're right. I would need to confirm this for `enableSQL` as we use that in a lot of places!

src/webviews/settings/index.ts
```diff
@@ -171,34 +163,20 @@ export class SettingsUI {
...
 
-          if (restart) {
-            vscode.window.showInformationMessage(`Some settings require a restart to take effect. Reload workspace now?`, `Reload`, `No`)
-              .then(async (value) => {
-                if (value === `Reload`) {
-                  await vscode.commands.executeCommand(`workbench.action.reloadWindow`);
-                }
-              });
+          Object.assign(config, data);
+          await ConnectionConfiguration.update(config);
```
sebjulliand, line 171: We should use `instance.setConfig` here. It will call `ConnectionConfiguration.update` and also update the current connection's config if applicable. Otherwise, `instance.getConfig` will return a stale config.
```suggestion
          await instance.setConfig(config);
```

chrjorgensen:
@sebjulliand Suggestion has been accepted! 😉

sebjulliand (approved):
I'm all for it!

src/api/Configuration.ts
```diff
@@ -119,7 +120,8 @@ export namespace ConnectionConfiguration {
       debugIsSecure: (parameters.debugIsSecure === true),
       debugUpdateProductionFiles: (parameters.debugUpdateProductionFiles === true),
       debugEnableDebugTracing: (parameters.debugEnableDebugTracing === true),
-      readOnlyMode: (parameters.readOnlyMode === true)
+      readOnlyMode: (parameters.readOnlyMode === true),
+      quickConnect: (parameters.quickConnect === true || parameters.quickConnect === undefined)
```
worksofliam, line 124: I tested this and it shows that Quick Connect will be on by default.

worksofliam (changes requested):
Wrong button

chrjorgensen:
> Code looks good, but the 'Connect and Reload' right click option doesn't work as expected. Seems like the UI isn't updating after the connection is made?
> 
> Video uploading here: https://youtu.be/wntqwrjuFdM

@worksofliam I see that - looking into it...

chrjorgensen:
@worksofliam I'm glad I asked you to have a look and do some testing! 😃 There were a problem with two optional parameters to `connect`, one of which were created in master branch after the other were created in my PR branch - and since they were both boolean and optional, no warning were issued during development!

It should have been fixed now, please retest...

worksofliam (approved):
Yup, looks like it is working as expected now. I really like this feature - thanks for your hard work @chrjorgensen!

I will give you the pleasure of merging.

chrjorgensen:
> I will give you the pleasure of merging.

Thanks @worksofliam and @sebjulliand

PR #1152, opened 2023-03-22
COBOL language id can also be uppercase

worksofliam:
### Changes

* BitLang COBOL implements the COBOL language id as `COBOL`, whereas our regex was only checking for `cobol.`
* I also fixed the default actions for COBOL, as it looks like `CRTBNDCBL` and `CRTCBLMOD` does not like unicode at all - but the good news is that `CRTSQLCBLI` does!

I have tested and is now working as expected.

### Checklist

* [x] have tested my change


chrjorgensen:
@worksofliam Why is the regex for language ID case sensitive - wouldn't it be better to ignore the casing?

worksofliam:
@chrjorgensen Good point. Got a demo to do today, so I will make sure to get it fixed after that!

chrjorgensen (dismissed):
@worksofliam Code is working as expected. I didn't test the default COBOL action setup, though...

PR #1154, opened 2023-03-22
Allow local actions to download directories post build

worksofliam:
### Changes

Now both files and directories can be downloaded as part of a local build.

### Checklist

* [x] have tested my change


src/api/CompileTools.ts
```diff
@@ -435,7 +435,15 @@ export namespace CompileTools {
...
                 const downloads = chosenAction.postDownload.map(
-                  downloadPath => clinet.getFile(path.join(localDir, downloadPath), path.posix.join(remoteDir, downloadPath))
+                  downloadPath => {
+                    const localPath = path.join(localDir, downloadPath);
+                    const remotePath = path.posix.join(remoteDir, downloadPath);
+                    if (downloadPath.endsWith(`/`)) {
+                      return client.getDirectory(localPath, remotePath)
+                    } else {
+                      return client.getFile(localPath, remotePath)
+                    }
```
sebjulliand, line 445: I'd suggest testing if the remote path is actually a directory, so we don't just trust how the path ends, and uses don't have to worry about it.
```suggestion
                    const isDirectory = (await connection.sendCommand({
                        command: `[ -d "${remotePath}" ]`
                    }))?.code === 0;
                    if (isDirectory) {
                      return client.getDirectory(localPath, remotePath)
                    } else {
                      return client.getFile(localPath, remotePath)
                    }
```

sebjulliand (changes requested):
Just a change to consider. Otherwise, it's great!

src/api/CompileTools.ts
worksofliam: Can't await in the `map` here. 😢 

I like your idea, perhaps we can determine this upfront before the map.

sebjulliand: How about this:
```typescript
// Then we download the files that is specified.
const downloads = chosenAction.postDownload.map(
  async (downloadPath) => {
    const localPath = path.join(localDir, downloadPath);
    const remotePath = path.posix.join(remoteDir, downloadPath);
    const isDirectory = (await connection.sendCommand({
      command: `[ -d "${remotePath}" ]`
    }))?.code === 0;
    if (isDirectory) {
      return client.getDirectory(localPath, remotePath)
    } else {
      return client.getFile(localPath, remotePath)
    }
  }
);
```

worksofliam: @sebjulliand 

I tried this and this snippet is always returning `code` as `1`, so it doesn't work. I am afraid I don't know this bash syntax well enough to understand it.

```
                    const isDirectoryCall = (await connection.sendCommand({
                      command: `[ -d "${remotePath}" ]`
                    }));
```

sebjulliand: `[ -d "${remotePath}" ]` tests if the path is a directory and returns 0 if it is.
If you open a shell and try `[ -d "/tmp" ] && echo it's a directory` it should echo `it's a directory`.

I tried it before submitting the suggestion, using `connection.sendCommand`, and got the expected result (i.e. 0 for a directory and 1 for a file).

How did it look in the output for you?!

worksofliam: You can see here what the value of `remotePath` is followed by the output of `sendCommand` with that syntax: 

image

sebjulliand: image

worksofliam: image

sebjulliand: Oh I got it! The path is not resolved when used in a command. Which means that this:
`/xxx/yyy/zzz/.logs` does not get resolved to `/xxx/yyy/zzz/logs`. The command will try to check if the file/directory actually named `.logs` in `/xxx/yyy/zzz` is a directory.

Here's an alternative: run `cd ${remotePath}` and check if the return code is 0. This should work, even for symbolic links actually!.

sebjulliand: (sorry, I forgot the quotes up there: `[ -d "/tmp" ] && echo "it's a directory"`)

worksofliam: The `cd` worked :) New commit incoming

src/schemas/actions.json
```diff
@@ -65,7 +65,7 @@
             "$id": "#/actions/items/anyOf/0/properties/postDownload",
             "type": "array",
             "title": "Post Download",
-            "description": "Remote files to download when the Action is complete.",
+            "description": "Remote files/folders to download when the Action is complete. If items ends with `/`, a directory is assumed.",
```
sebjulliand, line 68: It may be not worth keeping the last part since it's automatically detected now!
```suggestion
            "description": "Remote files/folders to download when the Action is complete.",
```

sebjulliand (dismissed):
Just a comment, but that's me being picky 😅

worksofliam:
@sebjulliand thanks! I'm gonna merge now.

PR #1161, opened 2023-03-23
Missing await on QSysFS basic write

sebjulliand:
### Changes
When source date support is disabled, QSysFS writeFile method will not wait for the upload to complete before returning.

### Checklist
* [x] have tested my change


worksofliam (approved):
Releasing 1.7.7 for this.

Issue #1173, opened 2023-03-25
Problems tab opening new file instead of using open one


worksofliam:
This is happening due to a (what I think is a bug) in the VS Code API, where the diagnostic API is taking URI query parameters into account when looking for open editors:

Issue raised here: https://github.com/microsoft/vscode/issues/178322

PR #1174, opened 2023-03-25
Fix use of filter parameter on Uris

worksofliam:
### Changes

* Solves issue that filter query parameter was causing diagnostics to no longer match up (#1173)
* Fixes consistent use of readonly flag where `filter` had a second implication (go and find out if this filter is readonly or not)
* No longer allow readonly sources/objects to have Actions run against them

### Checklist

* [x] have tested my change

### DIFF
--- a/src/views/objectBrowser.js
+++ b/src/views/objectBrowser.js
@@ -1157,7 +1157,7 @@ class Member extends vscode.TreeItem {
 
     this.contextValue = `member${filter.protected ? `_readonly` : ``}`;
     this.description = member.text;
-    this.resourceUri = getMemberUri(member, { filter: filter.name });
+    this.resourceUri = getMemberUri(member, filter.protected ? {readonly: true} : undefined);
     this.path = this.resourceUri.path;
     this.tooltip = `${this.resourceUri.path}${member.text ? `\n(${member.text})` : ``}`;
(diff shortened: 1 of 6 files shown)


src/views/objectBrowser.js
```diff
@@ -1157,7 +1157,7 @@ class Member extends vscode.TreeItem {
 
     this.contextValue = `member${filter.protected ? `_readonly` : ``}`;
     this.description = member.text;
-    this.resourceUri = getMemberUri(member, { filter: filter.name });
+    this.resourceUri = getMemberUri(member, filter.protected ? {readonly: true} : undefined);
```
sebjulliand, line 1160: Keep it simple 😉
```suggestion
    this.resourceUri = getMemberUri(member, {readonly: filter.protected});
```

sebjulliand (approved):
LGTM! Just a possible simplification on a line.

src/views/objectBrowser.js
worksofliam: There is a reason for this. I didn't want `readonly=false` on the URIs, which would cause the problem with the diagnostics again.

image

sebjulliand: Ugh, right, my bad 😅

PR #1175, opened 2023-03-25
Source date save now checks before using `TRANSLATE` scalar

worksofliam:
### Changes

SQL `TRANSLATE` is expensive when inserting a lot of data. This was very visible when working with large files. Translate is only required for source colour support, so therefore the save should only use translate when source colour is enabled.

I don't use source colours and this has sped up the save process significantly.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/filesystems/qsys/extendedContent.ts
+++ b/src/filesystems/qsys/extendedContent.ts
@@ -149,7 +149,7 @@ export class ExtendedIBMiContent {
         // Row length is the length of the SLQ string used to insert each row
         const rowLength = recordLength + 55;
         // 450000 is just below the maxiumu length for each insert.
-        const perInsert = Math.floor(450000 / rowLength);
+        const perInsert = Math.floor(400000 / rowLength);
 
         const rowGroups = sliceUp(rows, perInsert);
         rowGroups.forEach(rowGroup => {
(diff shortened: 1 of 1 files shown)


src/filesystems/qsys/extendedContent.ts
```diff
@@ -121,9 +123,17 @@ export class ExtendedIBMiContent {
             sourceData[i] = sourceData[i].substring(0, recordLength);
           }
 
-          rows.push(
-            `(${sequence}, ${sourceDates[i] ? sourceDates[i].padEnd(6, `0`) : `0`}, translate('${escapeString(sourceData[i])}', ${SEU_GREEN_UL_RI}, ${SEU_GREEN_UL_RI_temp}))`,
-          );
+          // We only want to do the translate when source colours at enabled.
+          // For large sources, translate adds a bunch of time to the saving process.
+          if (sourceColourSupport)
```
chrjorgensen, line 128: Before testing this, shouldn't the function `downloadMemberContentWithDates` have the same change - test for `sourceColourSupport`? Otherwise we convert some source without converting it back. I think we should be consistent here...

chrjorgensen (changes requested):

src/filesystems/qsys/extendedContent.ts
worksofliam: Yep, completely missed this. Sorry about that. Will get it fixed and ask for a review again.

PR #1195, opened 2023-03-29
Do not allow the same webview to be opened multiple times

sebjulliand:
### Changes
This PR will keep track of opened webviews and prevent multiple occurrences of the same webview (i.e. with the same title) to be opened.
If a webview is already opened and a request to open it again is issued, the opened webview will be revealed.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/CustomUI.ts
+++ b/src/api/CustomUI.ts
@@ -127,14 +127,28 @@ export class Section {
   }
 }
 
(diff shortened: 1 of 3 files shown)


worksofliam:
Code looks ok. Need to give it a functional test and then merge should be ok :)

worksofliam (approved):
Works as expected. Merging!

PR #1196, opened 2023-03-29
Fixed member opening after create/copy operations

sebjulliand:
### Changes
The member URI used to open the created/copied was wrong and prevented it from being opened right after its creation.
The commands now use `vscode.open` and `getMemberUri` to open the new member.

Member copy/creation process has been enhanced:
- Target name/path gets validated by vscode `showInput` API
- User can choose to retry in case of failure
- The process runs with a progress notification

### Checklist
* [x] have tested my change

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
 	"icon": "icon.png",
 	"displayName": "Code for IBM i",
 	"description": "Maintain your RPGLE, CL, COBOL, C/CPP on IBM i right from Visual Studio Code.",
-	"version": "1.7.11",
+	"version": "1.7.12",
 	"keywords": [
 		"ibmi",
 		"rpgle",
--- a/src/api/CompileTools.ts
+++ b/src/api/CompileTools.ts
@@ -282,7 +282,7 @@ export namespace CompileTools {
             case `member`:
               const memberDetail = connection.parserMemberPath(uri.path);
               evfeventInfo.library = memberDetail.library;
-              evfeventInfo.object = memberDetail.member;
+              evfeventInfo.object = memberDetail.name;
               evfeventInfo.extension = memberDetail.extension;
               evfeventInfo.asp = memberDetail.asp;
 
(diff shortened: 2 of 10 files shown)


chrjorgensen:
@sebjulliand I just spotted this PR and looked at the changes - and something caught my eye: Member without extension is no longer supported after this PR.

I noticed it because I did quite some coding to ensure the extension supported exactly that. Since you can add source members without extension (actually the default on the ADDPFM command), I figured our extension should do the same, and I added some code for supporting copying to a member without extension. Are we removing this support?

Before this PR the user could add, copy to or rename to a member without extension by specifying the member name followed by a `.` and nothing more. The `.` was a requirement to make the user be deliberate in having no extension for the member and not just have forgotten to enter an extension.

It is probably rarely used, but since it's allowed on the server, not supporting this could cause trouble for some users - if they have a policy for some members having no extension (like a README or similar). They will no longer be able to use such a naming standard after this PR.

sebjulliand:
Thanks for the clarification @chrjorgensen . I didn't understand at first why the method checking a member path would look for the `.` but not if an actual extension was provided after the `.` .

I'll remove the extra control looking for the actual extension name and leave a comment there to make sure we know it's not a bug but a feature 😁

chrjorgensen:
> I didn't understand at first why the method checking a member path would look for the `.` but not if an actual extension was provided after the `.` .

I thought it was obvious...! 😆 No, my bad for not documenting this in the first place, and making you waste some time.

sebjulliand:
> I thought it was obvious...! 😆 No, my bad for not documenting this in the first place, and making you waste some time.

Hey no worries at all, I didn't waste any time at all! It took me 2 seconds to remove that extra control and I learned a new thing 😉

worksofliam (approved):
@chrjorgensen I am very grateful for your review here. Specifically this catch:

> Before this PR the user could add, copy to or rename to a member without extension by specifying the member name followed by a . and nothing more. The . was a requirement to make the user be deliberate in having no extension for the member and not just have forgotten to enter an extension.

I tested this to ensure it continued to work, and it does - thanks for fixing that @sebjulliand!

Nonetheless, I am also so grateful for this fix. Fixing the UX is always one of my favs!

PR #1212, opened 2023-04-03
Fix #1200 Compiling Errors not highligthted for source files having l…

### LINKED ISSUE #1200: Compiling Errors not highligthted for source files having long paths (longer than the record length of the events file)
**Bug Description**

Source files that has errors and has a long path whose length is longer than the record length of the events file, when compile it the extension is not capable of highlighting it over the source code tab. If you go to problems tab you can see the messages but if any of this is clicked and an error is produced.

**Steps to reproduce the behavior**

1. Create a local project structure with many directories with large names or a big one using the max length allowed. This can be done working directly under the IFS instead of create a project.
2. Create a file with a long name inside the directory.
3. Deploy and compile.

**Expected behavior**
Errors highligthted with squiggly lines showed in the source file tab that was compiled.

**Screenshots**

image

image

image

**Environment**
 - Extension version 1.7.11
 - IBM i OS version 7.5

**Additional context**

The error parsers that process the messages from the events file are not taking care of the FILEIDCONT record type.

angelorpa:
…ong paths (longer than the record length of the events file)

### **Changes**

Add support for highlighting compiling errors over source files that have long paths (longer than the record length of the events file). The code added checks for the source file path length in the FILEID register type and compare to the max allowed length (255) that a file path can be in a FILEID record type, so this way to know when to check and process for FILEIDCONT records to get the complete source file path.

### **Checklist**

* [x] have tested my change

### **Testing**

_compiling source file members: sqlrpgle , rpgle with new and old handlers_
image
image
image

_source files from IFS with long paths including spaces in it_
image

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -7,6 +7,9 @@ vsc-extension-quickstart.md
 **/jsconfig.json
 **/*.map
 **/.eslintrc.json
+**/tsconfig.json
+src/**
+types/**
 docs/**
 .github/**
 node_modules


worksofliam (changes requested):
@angelorpa This is awesome. The code is looking pretty good to me. I would really appreciate it if you could add some comments in your else blocks here.

It will help us understand this code if we have to come back and look at in in the future.

Thank you!!

worksofliam:
I tested further and it's working as expected. This is a fairly impactful change. I would also appreciate reviews from @chrjorgensen and @sebjulliand.

chrjorgensen (changes requested):
@angelorpa @worksofliam Code looks good and works with no problem.

When the missing comments has been added we can approve this and get it merged.

sebjulliand (changes requested):
Code looks good and efficient.
However, for the sake of not duplicating code around, I may suggest that you create and export a function in `src/api/errors/handler.ts` to compute the source file path and call it in both `old.ts` and `new.ts`.

angelorpa:
> Code looks good and efficient.
> However, for the sake of not duplicating code around, I may suggest that you create and export a function in `src/api/errors/handler.ts` to compute the source file path and call it in both `old.ts` and `new.ts`.

I agree, in fact I was thinking on it. Let me create a EventsFileHandler.ts to expose some utilities functions and me be in a future we will use it as a unified event file handler.

worksofliam:
Converted to draft while we wait for additional changes.

angelorpa:
I did created a new PR  #1276 , a simple solution without write a new EventsFileHandler.ts, maybe when i have more free time i could write it.

On the other hand, there is a specially case that produce a wrong result, and occurs when a space that is part of the path is located at the end of a FILECONT line. The issue itself is produced by the getTable function used to get the lines of the events file because it removes the trailing blanks.  One posible solution is get the data from members using SQL this could be done using alias, a second solution could be change the CPYTOIMPF and modify the RMVBLANK parameter to not remove the trailing blanks.
https://github.com/halcyon-tech/vscode-ibmi/blob/9fa13fcbc94296310072abe1ff271526f0a70436/src/api/CompileTools.ts#L95-L100
https://github.com/halcyon-tech/vscode-ibmi/blob/9fa13fcbc94296310072abe1ff271526f0a70436/src/api/IBMiContent.ts#L248-L254

https://github.com/halcyon-tech/vscode-ibmi/blob/9fa13fcbc94296310072abe1ff271526f0a70436/src/api/IBMiContent.ts#L268-L272

PR #1179, opened 2023-03-26
Proposal: API tests

worksofliam:
### Changes

Our API is getting large.

I am proposing a testing suite that will allow us to test:

* Our export API
* Commands that we register

I am not using jest, mocha or vitest here so we can simply integrate easily into how we do development. The reason I am not using the testing suite that comes with VS Code (`vscode-test` package) is because of our dependency on the connection. I've built this so the tests can be triggered when we make a connection (plus we could even automate the connection with our `connectDirect` API)

Something that needs to be figured out:

* Not shipping our test code in production builds
* This can be done in our webpack config I bet

PR #1224, opened 2023-04-06
Methods for object resolves

worksofliam:
### Changes

Adds new methods to help with resolving paths to source members and streamfiles.

### Checklist

* [ ] have tested my change


src/api/IBMiContent.ts
```diff
@@ -531,6 +531,59 @@ export default class IBMiContent {
...
+        } catch (e) {
+          console.log(e);
+        }
+      }
+    }
+
+    return undefined;
+  }
+
+  async streamfileResolve(name: string, directories: QsysPath[]): Promise<string|undefined> {
```
worksofliam, line 564: @edmundreinhardt @SanjulaGanepola Looks like there is no way to `/usr/bin/find` to take a list of file names.

SanjulaGanepola: Taking a look at the documentation, the find command supports the OR operator. I think we can handle a list of file names using this?

image

worksofliam: @SanjulaGanepola Thanks - that syntax works. Working on that now.

```
-bash-5.1$ /usr/bin/find /QSYS.LIB/LIAMA.LIB/QRPGLESRC.FILE/ /QSYS.LIB/LIAMA.LIB/QRPGLEREF.FILE -name 'BADEMPS.*' -o -name 'BANG.*'
```

chrjorgensen:
@worksofliam Code looks good to me - but I have no way of testing this, since the new methods are not being called anywhere.

Btw, is this first step to make copy member resolution work like the RPG compiler? 🤞

worksofliam:
@chrjorgensen 

> Btw, is this first step to make copy member resolution work like the RPG compiler?

That's exactly right!

I do hope in the future we can write tests for this also. (#1179)

worksofliam:
I have added tests for these new APIs.

@SanjulaGanepola Note that `find` searches IFS directories recursively.

worksofliam:
@chrjorgensen There is a reason we're standardising here with these APIs - I don't want to write the same script in every extension (this will be used in RPGLE, CL, etc!)

Though that is very cool that you have figured out how to keep it simple - how do think we could go about using your idea inside of our API without `find`?

chrjorgensen:
@worksofliam Since copy members can be specified in different ways, it's probably a good idea to check the
[/COPY documentation](https://www.ibm.com/docs/en/i/7.3?topic=directives-copy-include) for how unspecified elements are handled.

IMHO Especially the following are important details:
- The default sourcefile searched in the /QSYS.LIB file system - QRPGLESRC
- When the library list are searched
- default suffixes for IFS include files
- Maybe also that the IFS is also searched when no member in /QSYS.LIB is found

chrjorgensen:
> @chrjorgensen There is a reason we're standardising here with these APIs - I don't want to write the same script in every extension (this will be used in RPGLE, CL, etc!)
> 
> Though that is very cool that you have figured out how to keep it simple - how do think we could go about using your idea inside of our API without `find`?

Oh, I didn't mean a script in a file - the code in my suggestion can be written to the shell in one line and be executed:

```shell
for f in a b c; do echo $f; done
```

So your code should build the shell string with the names and libraries/directories included in the list, like "a b c" etc.

> @chrjorgensen The specifics of those rules will be handled in the right language server. The rules you mentioned are specific to RPGLE, whereas CL and COBOL have different rules. The way I implemented these APIs will allow each language server to be flexible and implement the rules separately, but still rely on the same resolver

I suggest using a `for` loop with `if` to avoid too many repetitive calls to the server, thus optimizing the performance.

Since the language servers have different rules for search order and missing elements, isn't your interface a bit wrong:
having separate filenames and directories parameters does not allow specific ordering in the search.

IMHO the parameter to the `resolve` function should simply be a list of streamfiles, allowing for both /QSYS.LIB and IFS names (the member resolve function would add the `/QSYS.LIB/` etc to generate path names from the member and file names). Then it's up to the language servers to create the list of streamfiles (including members) to search for and hand the list to the `resolve` function, which would then build the shell command string and return the first file found.


I ❤️ having these idea exchanges! 😉

worksofliam:
@chrjorgensen 

> the code in my suggestion can be written to the shell in one line and be executed

That is very cool! I will look into making this change since that would even remove the `find` requirement! Would this also work outside of bash in a non-interactive shell?

> isn't your interface a bit wrong: having separate filenames and directories parameters does not allow specific ordering in the search.

This is only for the IFS resolve, which is a bit odd, but it will search in the orders provided to the user. E.g. it will search `file1` in `dir1`, and then `file2` in `dir1`, then `file1` in `dir2`, followed by `file2` in `dir2`.

> IMHO the parameter to the resolve function should simply be a list of streamfiles,

I understand this. This makes sense for the IFS resolve API for sure and I am going to change it based on your recommendation - this is how the member resolve already works. You provide one name and a list of source files and it builds that string for you.

Thanks so much for your feedback!

worksofliam:
@chrjorgensen Tried to use your suggestion, but it is not working as expected. Currently, `find` only returns the path if the file actually exists, and this suggestion returns the string nonetheless.

```
for f in /QSYS.LIB/QGPL.LIB/H.FILE/MATH.MBR,/QSYS.LIB/QSYSINC.LIB/MIH.FILE/MATH.MBR,/QSYS.LIB/QSYSINC.LIB/H.FILE/MATH.MBR; do echo $f; done
```

worksofliam:
@chrjorgensen Only just saw your comment.. let me try with your new string!

worksofliam:
@chrjorgensen That is awesome! It will likely perform a lot better too! Thank you so much.

It is now ready for review.

src/api/IBMiContent.ts
```diff
@@ -643,6 +642,48 @@ export default class IBMiContent {
...
 
+  async memberResolve(member: string, files: QsysPath[]): Promise<IBMiMember|undefined> {
+    const command = `for f in ${files.map(file => `/QSYS.LIB/${file.library.toUpperCase()}.LIB/${file.name.toUpperCase()}.FILE/${member.toUpperCase()}.MBR`).join(` `)}; do if [ -f $f ]; then echo $f; break; fi; done`;
+
+    const result = await this.ibmi.sendCommand({
+      command,
+    });
+
+    if (result.code === 0) {
+      const [firstMost] = result.stdout.split(`,`);
```
chrjorgensen, line 653: Since you have a break in your `for` loop, you don't need to retrieve the "firstmost" - there will be only one...

src/api/IBMiContent.ts
```diff
@@ -643,6 +642,48 @@ export default class IBMiContent {
...
+
+  async streamfileResolve(name: string, directories: string[]): Promise<string|undefined> {
+    const command = `for f in ${directories.map(dir => path.posix.join(dir, name)).join(` `)}; do if [ -f $f ]; then echo $f; break; fi; done`;
+
+    const result = await this.ibmi.sendCommand({
+      command,
+    });
+
+    if (result.code === 0 && result.stdout) {
+      const [firstMost] = result.stdout.split(`\n`);
```
chrjorgensen, line 679: Same here...

chrjorgensen (changes requested):
@worksofliam I have tested your PR by running the test suite - not using the resolve methods in real life. I just made a little comment, which you may address or not...

When your have considered this, I will approve and merge.

worksofliam (commented):
Thanks for your suggestion @chrjorgensen! I made the fix. Please review whenever you're ready.

worksofliam:
@chrjorgensen Sorry to bug you on this. Can I get one more review?

chrjorgensen (approved):
@worksofliam LGTM - approved...

angelorpa:
seems a bit late to ask, but here i go: this functions must necessarily be implemented using the /usr/bin/find command or coud be done with SQL? maybe using the QSYS2.SYSPARTITIONSTAT for member resolver and QSYS2.IFS_OBJECT_STATISTICS for the streamfile resolver ?

angelorpa:
> @angelorpa Why do you think it is a requirement to use `/usr/bin/find`?
> 
> The `find` command is used to search for files / directories in a directory and all subdirectories - but that it not what these functions are doing: They check for a member or streamfile in a list of source files / directories. And they only return the first member / streamfile found. Like the compilers do...
> 
> `find` is useful when you search for something in a directory tree - and I use it regularly. But it is not suited for this use case. Here it is much faster to let the shell loop through the list and check for the member / file. `find` could potentially take a long time in a big directory tree and "choke" the resolve function...

@chrjorgensen  I think you misunderstood my question or may be i formulated in a wrong way. I understood what this functions have to do, but i want to know if is possible implement this functionality using SQL instead of using the find command. I only was suggesting another way to do it. Tables QSYS2.IFS_OBJECT_STATISTICS and QSYS2.SYSPARTITIONSTAT could be used to search for a streamfile or a member, may be using SQL could be faster, easy to understand and more flexible.

chrjorgensen:
@angelorpa It sure is possible to create a search function using the SQL functions you mention - but that would only cover members, not streamfiles (unless you add the QSYS2.IFS_OBJECT_STATISTICS SQL function and then it get's tricky). Also running SQL has a cost when entering the SQL CLI, which runs in a separate job. SQL is great for retrieving information about something, but we just want to know the existence of a member or streamfile, not all its attributes.

Instead of SQL we went for shell scripting, which is widely used in the industry, especially in Linux, and is quick to check for files (does not start a new job for this). We use the `for` loop and `if` command, and you can google it easily searching for `bash for loop` and `bash if test`. So we actually never leave the shell program (which using SQL or `find` would do) and just let it do the lookups...

angelorpa:
> @angelorpa It sure is possible to create a search function using the SQL functions you mention - but that would only cover members, not streamfiles (unless you add the QSYS2.IFS_OBJECT_STATISTICS SQL function and then it get's tricky). Also running SQL has a cost when entering the SQL CLI, which runs in a separate job. SQL is great for retrieving information about something, but we just want to know the existence of a member or streamfile, not all its attributes.
> 
> Instead of SQL we went for shell scripting, which is widely used in the industry, especially in Linux, and is quick to check for files (does not start a new job for this). We use the `for` loop and `if` command, and you can google it easily searching for `bash for loop` and `bash if test`. So we actually never leave the shell program (which using SQL or `find` would do) and just let it do the lookups...

nice, thanks for the feedback.

Issue #1051, opened 2023-02-10
Go to file...button. Autotype suggestions as you typing path to source file.


beckhamrryyaann:
I would love to see an enhancement to the Go to file... button.  This button allows you to type in a source member path either in the IFS or QSYS filesystems.  It appears the feature tries to do some sort of auto suggestion but I'm not sure what its doing.
As you type in the path, it would be nice if the feature started filtering down your options the more you typed.

If I had a library called  LIBRARY1 with source physical file of QRPGLESRC and member name of TESTFILE.sqlrpgle, as i typed
LIB.....it would show list of libraries containing LIB.  Then once i completed my library name and typed the slash "/" it would then start filtering on source physical files and then once i completed source physical file and typed slash "/" it would start filtering the member names.

Also before you type anything in the path box, it should show you a list of the last 10 recent source files you just opened.

I use the Go to file...button many times a day and its frustrating to have to key it in everytime.  I know i could create a filter but alot of times when doing support on the system I'll have to pull up a source member for just a few moments and setting up a filter is not practical for all these instances.

This maybe something I could write myself if you could guide me on where to implement this idea......... if you like this idea.

worksofliam:
> It appears the feature tries to do some sort of auto suggestion but I'm not sure what its doing.

It only lists file that you have previously opened to browsed (in both the Object Browser and IFS Browser).

> If I had a library called LIBRARY1 with source physical file of QRPGLESRC and member name of TESTFILE.sqlrpgle, as i typed
> LIB.....it would show list of libraries containing LIB. Then once i completed my library name and typed the slash "/" it would then start filtering on source physical files and then once i completed source physical file and typed slash "/" it would start filtering the member names.

Ideally this would be the best way to do it, but spamming requests to the system based on what the user types can take a heavily toll on the remote system. I'm open to hear ideas on how we could achieve this without hurting the server too much :)

worksofliam:
@halcyon-tech/core Any ideas on this? Specifically: 

> I'm open to hear ideas on how we could achieve this without hurting the server too much

Wright4i:
> @halcyon-tech/core Any ideas on this? Specifically:
> 
> > I'm open to hear ideas on how we could achieve this without hurting the server too much

Yeah there's no way doing the lookup in real time is going to be fast enough. Either that needs to be cached locally or maybe we could make an MQT? Either option would need a separate command to refresh the cache.

[...]

PR #1242, opened 2023-04-17
Feature/go to file autotype suggestions

sebCIL:
### Changes

Go to file...button. Autotype suggestions as you typing path to source file. Autotype is triggered by typing a '*'.
[Initial issue](https://github.com/halcyon-tech/vscode-ibmi/issues/1051)

### Checklist

* [X] have tested my change

### DIFF
--- a/package.json
+++ b/package.json
@@ -385,12 +385,7 @@
 								"type": "boolean",
 								"default": false,
 								"description": "Tells the debug service to send more data to the client. Only useful for debugging issues in the service."
-							},
-              "code-for-ibmi.goToFileAutoSuggest": {
-                "type": "boolean",
-                "default": false,
-                "description": "Suggesting Library / Source file / Member when typing path to source file."
-              }
+							}
 						}
 					}
 				},
--- a/package.json
+++ b/package.json
@@ -32,7 +32,9 @@
 		"theme": "dark"
 	},
 	"activationEvents": [
-		"onStartupFinished"
+		"onStartupFinished",
+		"onFileSystem:member",
+		"onFileSystem:streamfile"
 	],
 	"main": "./dist/extension.js",
 	"contributes": {
@@ -851,6 +853,12 @@
 				"enablement": "code-for-ibmi:connected"
 			},
 			{
+				"command": "code-for-ibmi.debug.stop",
+				"title": "Stop Debug Service",
+				"category": "IBM i Debug",
+				"enablement": "code-for-ibmi:connected"
+			},
+			{
 				"command": "code-for-ibmi.debug.activeEditor",
 				"title": "Start Debugging Active Source",
 				"category": "IBM i",
@@ -1405,7 +1413,7 @@
 				"title": "Collapse all",
 				"category": "IBM i",
 				"icon": "$(collapse-all)"
-			},			
+			},
 			{
 				"command": "code-for-ibmi.sortIFSFilesByDate",
 				"title": "Date",
@@ -2136,6 +2144,7 @@
 		"@types/glob": "^7.1.3",
 		"@types/node": "^12.20.55",
 		"@types/source-map-support": "^0.5.6",
+		"@types/tar": "^6.1.4",
 		"@types/tmp": "^0.2.3",
 		"@types/vscode": "^1.66.0",
 		"esbuild-loader": "^3.0.1",
@@ -2154,6 +2163,7 @@
 		"csv": "^6.2.1",
 		"ignore": "^5.1.9",
 		"node-ssh": "^11.1.1",
+		"tar": "^6.1.13",
 		"tmp": "^0.2.1",
 		"vscode-diff": "^2.0.2"
 	},
--- a/package.json
+++ b/package.json
@@ -383,7 +383,12 @@
 								"type": "boolean",
 								"default": false,
 								"description": "Tells the debug service to send more data to the client. Only useful for debugging issues in the service."
-							}
+							},
+              "code-for-ibmi.goToFileAutoSuggest": {
+                "type": "number",
+                "default": -1,
+                "description": "Number of characters to enter before suggesting Library / Source file / Member when typing path to source file. (-1 to disable)"
+              }
 						}
 					}
 				},
@@ -795,11 +800,6 @@
 							"command": "?CALL &LIBRARY/&NAME"
 						}
 					]
-				},
-				"code-for-ibmi.goToFileAutoSuggest": {
-					"type": "number",
-					"default": -1,
-					"description": "Number of characters to enter before suggesting Library / Source file / Member when typing path to source file. (-1 to disable)"
 				}
 			}
 		},
--- a/package.json
+++ b/package.json
@@ -795,6 +795,11 @@
 							"command": "?CALL &LIBRARY/&NAME"
 						}
 					]
+				},
+				"code-for-ibmi.goToFileAutoSuggest": {
+					"type": "number",
+					"default": -1,
+					"description": "Number of characters to enter before suggesting Library / Source file / Member when typing path to source file. (-1 to disable)"
 				}
 			}
 		},
--- a/package.json
+++ b/package.json
@@ -387,9 +387,9 @@
 								"description": "Tells the debug service to send more data to the client. Only useful for debugging issues in the service."
 							},
               "code-for-ibmi.goToFileAutoSuggest": {
-                "type": "number",
-                "default": -1,
-                "description": "Number of characters to enter before suggesting Library / Source file / Member when typing path to source file. (-1 to disable)"
+                "type": "boolean",
+                "default": false,
+                "description": "Suggesting Library / Source file / Member when typing path to source file."
               }
 						}
 					}
--- a/src/instantiate.ts
+++ b/src/instantiate.ts
@@ -238,18 +238,18 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
 
               case 2:
                 if (selectionSplit[1].length >= 3) {
-                  resultSet = await content!.runSQL(`SELECT SYSTEM_TABLE_NAME, TEXT_DESCRIPTION FROM QSYS2.SYSFILES WHERE SYSTEM_TABLE_SCHEMA = '${selectionSplit[0]}' AND FILE_TYPE = 'SOURCE' and system_table_name like upper('${selectionSplit[1]}%') ORDER BY SYSTEM_TABLE_NAME`);
+                  resultSet = await content!.runSQL(`SELECT SYSTEM_TABLE_NAME, TABLE_TEXT FROM QSYS2.SYSTABLES WHERE TABLE_SCHEMA = '${selectionSplit[0]}' AND FILE_TYPE = 'S' and SYSTEM_TABLE_NAME like upper('${selectionSplit[1]}%') ORDER BY SYSTEM_TABLE_NAME`);
 
                   if (resultSet.length > 0) {
-                    let listMember: vscode.QuickPickItem[] = []
+                    let listFile: vscode.QuickPickItem[] = []
 
                     resultSet.forEach(row => {
-                      listMember.push({
+                      listFile.push({
                         label: selectionSplit[0] + '/' + String(row.SYSTEM_TABLE_NAME),
                         detail: String(row.TEXT_DESCRIPTION)
                         })
                     })
-                    quickPick.items = [sourceSeparator, ...listMember,fileSeparator,...listItems]
+                    quickPick.items = [sourceSeparator, ...listFile,fileSeparator,...listItems]
                   }
                 }
 
@@ -266,15 +266,15 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
                           ORDER BY TABLE_PARTITION`);
 
                   if (resultSet.length > 0) {
-                    let listFile: vscode.QuickPickItem[] = []
+                    let listMember: vscode.QuickPickItem[] = []
 
                     resultSet.forEach(row => {
-                      listFile.push({
+                      listMember.push({
                         label: selectionSplit[0] + '/' + selectionSplit[1] + '/' + String(row.TABLE_PARTITION) + '.' + String(row.SOURCE_TYPE),
                         detail: String(row.PARTITION_TEXT)
                         })
                     })
-                    quickPick.items = [sourceSeparator,...listFile]
+                    quickPick.items = [sourceSeparator,...listMember]
                   }
                 }
 
--- a/src/instantiate.ts
+++ b/src/instantiate.ts
@@ -229,125 +229,138 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
         if (!list.includes(quickPick.value.toUpperCase())) quickPick.items = [quickPick.value.toUpperCase(), ...list].map(label => ({ label }));
 
         // autosuggest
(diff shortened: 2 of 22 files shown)


src/instantiate.ts
```diff
@@ -188,26 +201,104 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
...
+                })
+                
+                quickPick.items = [sourceSeparator,...listSchema,fileSeparator,...listItems]
+              }
+              
+              break;
+
+            case 2:
+              if (selectionSplit[1].length >= 3) {
+                resultSet = await content!.runSQL(`SELECT SYSTEM_TABLE_NAME, TEXT_DESCRIPTION FROM QSYS2.SYSFILES WHERE SYSTEM_TABLE_SCHEMA = '${selectionSplit[0]}' AND FILE_TYPE = 'SOURCE' and system_table_name like upper('${selectionSplit[1]}%') ORDER BY SYSTEM_TABLE_NAME`);
```
chrjorgensen, line 238: Please use `QSYS2.SYSTABLES` instead of `SYSFILES`- the latter is a new contruct and may not be available on the server connected to by the user.

```suggestion
                resultSet = await content!.runSQL(`SELECT SYSTEM_TABLE_NAME, TEXT_DESCRIPTION FROM QSYS2.SYSTABLES WHERE SYSTEM_TABLE_SCHEMA = '${selectionSplit[0]}' AND FILE_TYPE = 'S
```

chrjorgensen (changes requested):
@sebCIL Thank for this PR! 👍 

I just gave it a spin and have these comments / questions:

- You use SQL but does not consider the configuration setting `enableSQL`. If this is disabled, you can't use SQL - and maybe the auto-suggest function has to be disabled, if you have no other way to get the list?
- You retrieve the list values from the SQL metadata tables - but these are in CCSID 37 (US) and does not take the user language into consideration. See the `getMemberList` function to see an example how to handle this. As is now, I get the following: 
billede
- Beware of the performance: You call the server each time the user enters a key after the first three characters in the name! The server may be slow and could make this client slow for the user. I see three options:
  - Keep the SQL as is and simply filter from the first `SELECT` - you already have the complete list.
  - Put a `LIMIT x` on the SQL to reduce the date returned.
  - Maybe an async call to SQL to collect the list in the background (don't know if it's possible).
- Since this could have a significant impact on the `Go to file` function, it should be a setting in the server configuration, so the user can enable or disable it as appropriate,
- If you can make the auto-suggest perform, you could reduce the required number of characters before auto-suggest is activated. Depending on the naming convention in place, three characters could be too many and make this function useless. Maybe the setting mentioned previously could be the number of characters required?
- Haven't you switched the member and file variables somewhat? In `case 2` you use member variables and in `case 3` you use file variables. I think it should be the other way around, right? Not that it has any influence on the function, it just makes the code confusing.

@worksofliam @sebjulliand Any comments (to the comments)? 😄

sebCIL:
Thanks for the feedback, I'm looking into it

sebCIL:
Ok, i made some changes.

What do you think about it?

src/instantiate.ts
```diff
@@ -188,26 +215,152 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
...
-      quickPick.onDidAccept(() => {
+      quickPick.onDidAccept(() => { 
         const selection = quickPick.selectedItems[0].label;
         if (selection) {
-          if (selection === `Clear list`) {
-            storage.setSourceList({});
-            vscode.window.showInformationMessage(`Cleared list.`);
+          const selectionSplit = selection.split('/')
+          if (selectionSplit.length === 3) {
+            if (selection === `Clear list`) {
```
chrjorgensen, line 353: This should not be conditioned by `selectionSplit.length === 3`!
As is now, if I select `Clear list` without having entered a three part name, I get the following:

billede

and the list is still there the next time I select `IBM i: Go to File...`.

EDIT: Just noted that `Clear list` is not even there when the list is being filtered!

src/instantiate.ts
```diff
@@ -188,26 +215,152 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
...
+                }
+              }
+
+              if (selectionSplit[1].length >= goToFileAutoSuggest) {
+                listDisplay = listFile.filter(file => file.label.startsWith(selectionSplit[0].toUpperCase() + '/' + selectionSplit[1].toUpperCase()))
+              }
+                
+              quickPick.items = [
+                {
+                  label: 'Sources member',
```
chrjorgensen, line 285:
```suggestion
                  label: 'Members',
```

To be consistent...

chrjorgensen (changes requested):
@sebCIL Good effort! I like your idea of caching beforehand - and even though it took almost ten seconds to cache the libraries on one of our systems and the `Go to File` dialog was shown, this is acceptable since I can (eventually) disable autosuggest for that system. I also like it now handles the variant characters correctly instead of showing `#@$`.

There are some problems producing a correct filtered list - see an example here:

billede

Maybe the list parts should be sorted and filtered individually and combined in the end?
The `Clear list` should not be included in the filtered list.

Even though it is not mandatory and we all have our individual preferences, please follow the consensus in this project and add a semicolon to your statements - to indicate end of statement!

package.json
```diff
@@ -795,6 +795,11 @@
 							"command": "?CALL &LIBRARY/&NAME"
 						}
 					]
+				},
+				"code-for-ibmi.goToFileAutoSuggest": {
+					"type": "number",
+					"default": -1,
+					"description": "Number of characters to enter before suggesting Library / Source file / Member when typing path to source file. (-1 to disable)"
```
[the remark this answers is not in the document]
chrjorgensen: And since you actually only test `goToFileAutoSuggest` for values greater than or equal to zero, you should make this configuration setting a boolean.

chrjorgensen (changes requested):

package.json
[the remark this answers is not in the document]
chrjorgensen: See below if this setting is needed at all...

chrjorgensen:
@sebCIL One important thing that I forgot to mention: The `Go to File...` handles both members and streamfiles!
So your autosuggest should not run when the value starts with `/`.

I also got some ideas that you may include if you like them:
1. When you cache libraries, files and members, you could write a message in the bar, e.g. "Caching libraries..." before you start your SQL, and remove the message when the cache is complete. This would tell the use why the dialog does not show immediately for long running SQL's.
2. When navigating to a library or file, the user may not always know the files or members in the library or file. If the user enters `*` immediately after the ´/´, you could show the whole list (and remove the `*` from the value). That way the user can see the content and start filtering. Actually a kind of ad-hoc object filter!

Thinking more on the `*` in idea 2... this could actually be a trigger for the autosuggest! So if you implement this, there is no need for a configuration setting at all... a better solution for the user, I think... WDYT?

Keep up the great work - this is going to be an awesome feature!

src/instantiate.ts
```diff
@@ -188,26 +215,152 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
...
+                    kind: vscode.QuickPickItemKind.Separator
+                },
+                ...listItems
+              ]
+              
+              break;
+
+            case 2:
+              // Create cache
+              if (listFile.length === 0) {
```
chrjorgensen, line 258: To me it seems like the cache is not cleared when the library changes?

src/instantiate.ts
```diff
@@ -188,26 +215,152 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
...
+                  kind: vscode.QuickPickItemKind.Separator
+                },
+                ...listItems
+              ]
+
+              break;
+
+            case 3:
+              // Create cache
+              if (listMember.length === 0) {
```
chrjorgensen, line 300: Same as above - is the member list cleared when the source file changes?

chrjorgensen (changes requested):

package.json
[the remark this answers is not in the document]
sebCIL: > And since you actually only test `goToFileAutoSuggest` for values greater than or equal to zero, you should make this configuration setting a boolean.

I confirm that I test the value to compare it to the number of characters entered : 
`if (selectionSplit[1].length >= goToFileAutoSuggest) {`

Not sure a boolean would be enough !?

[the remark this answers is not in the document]
sebCIL: > You created a global setting, not a connection setting... wink Please change it into a connection setting.
> 
> For an example of a connection setting, search for `showDescInLibList`.

Ok, I moved it.

sebCIL:
> Thinking more on the `*` in idea 2... this could actually be a trigger for the autosuggest! So if you implement this, there is no need for a configuration setting at all... a better solution for the user, I think... WDYT?

This is a good idea, I'll try to implement it.

sebCIL:
I made some changes, is it better?

chrjorgensen:
@sebCIL Sorry for the delay - hope to make a review this weekend. Thanks!

src/instantiate.ts
```diff
@@ -188,26 +200,183 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
...
+        quickPick.placeholder = `Enter file path (Format: LIB/SPF/NAME.ext use '*' for wildcard or /home/xx/file.txt)`;
+      }          
+      
+      quickPick.onDidChangeValue(async () => {
         // INJECT user values into proposed values
         if (!list.includes(quickPick.value.toUpperCase())) quickPick.items = [quickPick.value.toUpperCase(), ...list].map(label => ({ label }));
+
+        // autosuggest
+        if (config && config.enableSQL && goToFileAutoSuggest && (!quickPick.value.startsWith(`/`))) {
+          const asteriskIndex = quickPick.value.indexOf(`*`) ;
```
chrjorgensen, line 237: Would it be better if you remove the `*` typed after filling the list with possible values? This would indicate to the user that the `*` is not part of the name. I see that you remove it when the user selects from the list, but it's probably better to remove it asap.

I wonder what happens when the `*` is kept - will the code be run at every keystroke and query the server?

src/instantiate.ts
```diff
@@ -188,26 +200,183 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
...
+                      })
+                    })
+                  }
+                }
+
+                listDisplay = listFile.filter(file => file.label.startsWith(selectionSplit[0].toUpperCase() + '/' + filterText.toUpperCase()));
+
+                quickPick.items = [
+                  {
+                    label: 'Sources files',
```
chrjorgensen, line 299:
```suggestion
                    label: 'Source files',
```

src/instantiate.ts
```diff
@@ -188,26 +200,183 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
...
+
+      // Create a cache for Schema if autosuggest enabled
+      if (listSchema.length === 0 && goToFileAutoSuggest && config && config.enableSQL) {
+        const resultSetLibrary = await content!.runSQL(`SELECT cast(SYSTEM_SCHEMA_NAME as char(10) for bit data) SYSTEM_SCHEMA_NAME, 
+          ifnull(cast(SCHEMA_TEXT as char(50) for bit data), '') SCHEMA_TEXT 
+        FROM QSYS2.SYSSCHEMAS 
+        WHERE SYSTEM_SCHEMA_NAME NOT LIKE 'Q%' 
+          ORDER BY 1 `);
+
+        quickPick.placeholder = `Caching...`;
```
chrjorgensen, line 217: Setting `quickPick.placeholder` at this point does not seem to work - I don't see it... probably because the dialog is not yet shown.

src/instantiate.ts
```diff
@@ -188,26 +200,183 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
...
+                      label: 'Files',
+                      kind: vscode.QuickPickItemKind.Separator
+                  },
+                  ...listItems
+                ]
+                
+                break;
+
+              case 2:
+                // Create cache
```
chrjorgensen, line 271: You should clear the member cache at this point - or it will not be updated when the source file changes...

chrjorgensen (dismissed):
@sebCIL Thanks for the changes - and sorry for the delay! I have made a thorough review because of the long break with some comments.

Looking forward to get this functionality in the extension! :smile:

src/instantiate.ts
sebCIL: This is a test I forgot to remove.

src/instantiate.ts
sebCIL: I don't know how to remove the `*` ,because if do that, the `quickPick.onDidChangeValue` function will be recalled.

chrjorgensen:
@sebCIL Thanks for following up on this PR! It could help solve some issues that users are facing, e.g. #1563. I will review it this week, hopefully.

src/instantiate.ts
chrjorgensen: Seems like the * is not a problem. We'll leave it here.

chrjorgensen (commented):
@sebCIL Sorry for taking so long - tested it again and it is looking good! 👍
I discovered that this feature is awesome and missed when I switched back to the master branch. I hope to include it in the upcoming release or the next one.

chrjorgensen:
@worksofliam I would very much like to have this awesome (but delayed - my fault!) feature included in the next release.

Will you give it a spin with some fresh eyes? 👀

worksofliam:
@chrjorgensen yup!! Let's stick it into the 2.5.0 milestone and I will review it asap.

chrjorgensen (commented):
No comments - looking good!

chrjorgensen:
@worksofliam Already in the 2.5.0 milestone - you can review... Thanks! 😃

chrjorgensen:
@worksofliam Hold it for a minute - I found an error and will fix it...

chrjorgensen:
@worksofliam Fixed - go ahead with your review when you feel like it... 😃

sebCIL:
> @sebCIL Sorry for taking so long - tested it again and it is looking good! 👍 I discovered that this feature is awesome and missed when I switched back to the master branch. I hope to include it in the upcoming release or the next one.

No problem !
Thanks for your feedback.

chrjorgensen:
@sebCIL I fixed one small issue in commit https://github.com/codefori/vscode-ibmi/pull/1242/commits/4e44f2e8806d735b48016deecfe1b398ed57081e - but have now discovered, that the branch is so far behind master, that we need to squash and rebase the branch on master.

Do you want to do it yourself? It's **urgent** if you want the change merged before the next release!

Or I can do it - but then there's a risk that I will be shown as author (not sure how git will react on this).
You will of course get all credit in the release notes!

chrjorgensen:
@sebCIL I have squashed the commits in this PR into a single commit keeping you as the author. But I had to push the commit into a new PR #1680, which replaces this PR.

Hope you think it's okay.

We all learn along the way and I learned that long-running feature branches requires frequent merge or rebase on master branch, and, if there are many back and forth commits, squashing of the commits before publishing. I was actually surprised that in the end only one file was changed...

Keep up the good work!

@worksofliam I have converted this PR to draft to avoid accidental merging - please use and review PR #1680 instead of this one.

worksofliam:
@chrjorgensen I am guessing this PR should be closed if that is the case.

I can confirm that @sebCIL will still be an author if that's the case.

chrjorgensen:
@worksofliam Okay, I will close this PR to avoid any confusion.

sebCIL:
> @chrjorgensen I am guessing this PR should be closed if that is the case.
> 
> I can confirm that @sebCIL will still be an author if that's the case.

Thx. 
The important thing is that the PR is integrated.

PR #1253, opened 2023-04-21
Feature/optional passphrase when using ssh key in UI

NicolasSchindler:
### Changes

Added the Passphrase option to the login and settings screen.
The logic is the same as for the password. If no passphrase was given when setting up the connection, a popup will appear asking for a passphrase.

### Checklist

* [X] have tested my change
* [X] have added myself to the contributors' list in [CONTRIBUTING.md](https://github.com/halcyon-tech/vscode-ibmi/blob/master/CONTRIBUTING.md)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "code-for-ibmi",
-	"version": "2.10.6-dev.0",
+	"version": "2.10.7-dev.0",
 	"lockfileVersion": 2,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "code-for-ibmi",
-			"version": "2.10.6-dev.0",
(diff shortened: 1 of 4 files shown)


chrjorgensen (changes requested):
@NicolasSchindler Thanks for the PR - nice addition to handle passphrases on SSH keys! 👍 

I only have a few comments:

- This will change the behavior for users currently not using passphrases, since you always ask for a passphrase. It should if possible only ask for passphrase if the key has one.
- How do the user remove a passphrase? If a passphrase is entered by mistake, it seems there is no way to remove it again?
- You're not deleting the passphrase in the `secrets` when the connection is deleted! So a new connection with the same name will use the old passphrase.

PR #1153, opened 2023-03-22
Show more info in tooltip for members and IFS files

chrjorgensen:
### Changes

This PR will enhance the tooltips for members and IFS files by adding
- number of lines (for members)
- size (for IFS files)
- Change date
- owner (for IFS files)

### Checklist

* [x] have tested my change
* [x] eslint is not complaining
* [x] **for feature PRs**: PR only includes one feature enhancement.

PR #1256, opened 2023-04-25
Fix stat on empty or specially named directories

chrjorgensen:
### Changes

This PR will fix running the `stat` command on directories, which are empty or has special characters in the name, e.g. `$`.
The error was introduced in PR #1153, and caused the following error in VS Code:

billede

### Checklist

* [x] have tested my change


sebjulliand (approved):
Works as expected; doesn't crash on an empty directory.
Hopefully the error message will never change or get localized 😅

Issue #1261, opened 2023-04-28
`postDownload` not downloading/creating directories


SJLennon:
**Is your feature request related to a problem? Please describe.**
Slight convenience feature.

**Describe the solution you'd like**
When running "`Launch Actions Setup`" and checking "`ibmi-bob`", shouldn't it create a `.logs` directory if one doesn't exist?

**Describe alternatives you've considered**
I can easily create it manually 

![image](https://user-images.githubusercontent.com/67484051/235255743-10f29c4b-f63f-49ae-adee-88f8e76c6be5.png)

edmundreinhardt:
The .logs directory is created by the BOB tooling and vscode-ibmi download.  There is no need to manually create it.

SJLennon:
@worksofliam 
@edmundreinhardt 

What am I doing wrong?  Here's the scenario:

Cloned the https://github.com/IBM/bob-recursive-example project from Github to a local directory on my PC.
Created Launch Actions Setup and selected ibmi-bob.
Did a build all, which deploys then fails because I forgot to change &lib1 to &LIB1 in iproj.json.
Fixed iproj.json.
Did build all again.
The build completes, but I get this failure:
![image](https://user-images.githubusercontent.com/67484051/235324859-37cc1bd4-b431-486d-9fee-86a07ac39f0a.png)
The .logs directory does not exist on my PC.
The .logs directory is created on PUB400.
![image](https://user-images.githubusercontent.com/67484051/235324963-0545389c-3a76-4ddb-a6f6-0590d3bd9eae.png)

worksofliam:
@SJLennon do you still get this error if you create the directory manually? Perhaps something is happening with the directory creation.

SJLennon:
@worksofliam If I create it manually in File Explorer the download works.

![image](https://user-images.githubusercontent.com/67484051/235325572-2530855f-c707-4c79-b24a-6ce54d8c92b6.png)


![image](https://user-images.githubusercontent.com/67484051/235325551-14006ac4-519f-4d0c-8dbd-c440049daab2.png)

worksofliam:
What action are you running to run Bob? Please share the json, thanks!

SJLennon:
```
[
  {
    "extensions": [
      "GLOBAL"
    ],
    "name": "Build all",
    "command": "OPT=*EVENTF BUILDLIB=&CURLIB /QOpenSys/pkgs/bin/makei build",
    "environment": "pase",
[...]

PR #1264, opened 2023-04-30
Fix post download directories creation

sebjulliand:
### Changes
Fixes https://github.com/halcyon-tech/vscode-ibmi/issues/1261

The post download directories creation would crash, on Windows at least, with the following error:
```
Error (FileSystemError): Unable to resolve filesystem provider with relative file path 'c:\Temp\bob-recursive-example\.logs'
```

The creation process has been fixed and the following rules now apply:
- Post download directories are created if they don't exist
- Post download directories are deleted and created if they exist
- If the post download target exists in the workspace but is not a directory, a warning will ask if it should be deleted first

Also, `.evfevent` has been added as a post download folder for BOB's default actions (it may be parsed in the future).

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/CompileTools.ts
+++ b/src/api/CompileTools.ts
@@ -458,7 +458,7 @@ export namespace CompileTools {
                 const downloadDirectories = chosenAction.postDownload.map(path.parse)
                   .map(pathInfo => pathInfo.dir || pathInfo.base) //Get directories or files' parent directory
                   .filter(Tools.distinct) //Remove duplicates
-                  .map(downloadDirectory => vscode.Uri.parse(`${localDir}/${downloadDirectory}`)); //Create local Uri path
+                  .map(downloadDirectory => vscode.Uri.parse((path.posix.join(localDir, downloadDirectory)))); //Create local Uri path
 
                 for (const downloadPath of downloadDirectories) {                  
                   try {
@@ -489,7 +489,7 @@ export namespace CompileTools {
                 // Then we download the files that is specified.
                 const downloads = chosenAction.postDownload.map(
                   async (downloadPath) => {
-                    const localPath = vscode.Uri.parse(`${localDir}/${downloadPath}`).fsPath;
+                    const localPath = vscode.Uri.parse(path.posix.join(localDir, downloadPath)).fsPath;
                     const remotePath = path.posix.join(remoteDir, downloadPath);
                     const isDirectoryCall = (await connection.sendCommand({
                       command: `cd ${remotePath}`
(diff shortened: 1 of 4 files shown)


src/api/CompileTools.ts
```diff
@@ -450,26 +450,46 @@ export namespace CompileTools {
...
+                      console.log(error);
+                    }
+                  }
                 }
 
                 // Then we download the files that is specified.
                 const downloads = chosenAction.postDownload.map(
                   async (downloadPath) => {
-                    const localPath = path.join(localDir, downloadPath);
+                    const localPath = vscode.Uri.parse(`${localDir}/${downloadPath}`).fsPath;
```
worksofliam, line 492: The `.parse` looks right, but shouldn't we still be using `path.join(localDir, downloadPath)` here?

src/api/CompileTools.ts
```diff
@@ -450,26 +450,46 @@ export namespace CompileTools {
...
-                    return vscode.workspace.fs.createDirectory(vscode.Uri.parse(path.join(localDir, pathInfo.dir || pathInfo.base)));
-                  });
-
-                  await Promise.all(directories);
-                } catch (e) {
-                  // We don't really care if it errors. The directories might already exist.
+                const downloadDirectories = chosenAction.postDownload.map(path.parse)
+                  .map(pathInfo => pathInfo.dir || pathInfo.base) //Get directories or files' parent directory
+                  .filter(Tools.distinct) //Remove duplicates
+                  .map(downloadDirectory => vscode.Uri.parse(`${localDir}/${downloadDirectory}`)); //Create local Uri path
```
worksofliam, line 461: Same note about `path.join` here.

worksofliam (changes requested):
Added a note where I think `path.join` should still be used, but beautiful nonetheless!

src/api/CompileTools.ts
sebjulliand: We are building a URI here so, we can't use `path.join`, as it will use `\` to join the paths on Windows, creating an invalid Uri.
What we could use is `path.posix.join`, that will always join using `/`, but it basically what I do here with string concatenation.
@worksofliam : I don't mind using `path.posix.join` here if you feel it's less confusing.

worksofliam (approved):
Working as expected. Tested with this custom Action:

```json
  {
    "name": "Build and deploy with make 🔨",
    "command": "/QOpenSys/pkgs/bin/gmake BIN_LIB=&CURLIB OPT=*EVENTF",
    "environment": "pase",
    "deployFirst": true,
    "extensions": [
      "GLOBAL"
    ],
    "postDownload": [
      ".logs/",
      ".evfevent/woop.txt"
    ]
  }
```

And this makefile content to ensure the postDownload data existed on the server:

```makefile
all:
	mkdir -p .logs
	echo "hello world" > .logs/hello.txt
	mkdir -p .evfevent
	echo "woops" > .evfevent/woop.txt
```

Issue #1026, opened 2023-01-26
Deploy loses executable permission


edmundreinhardt:
When deploying a file that has executable mode, these are lost in the resultant files in the IFS

**To Reproduce**
1. Clone https://github.com/IBM/ibmi-bob
2. Deploy it to an IFS directory
3. ls -al /deploy/dir/ibmi-bob/bin/makei 
4. Note the executable mode is missing

**Expected behavior**
The deployed file is still executable

**Screenshots**
```
reinhard@Edmunds-MBP ibmi-bob % ls -al bin
total 16
drwxr-xr-x@  4 reinhard  staff   128 25 Jan 17:35 .
drwxr-xr-x  23 reinhard  staff   736 26 Jan 12:53 ..
-rw-r--r--   1 reinhard  staff  1716 25 Jan 17:35 crtfrmstmf
-rwxr-xr-x   1 reinhard  staff  1720 25 Jan 17:35 makei
reinhard@Edmunds-MBP ibmi-bob % ssh p8adt03.rch.stglabs.ibm.com
reinhard@p8adt03.rch.stglabs.ibm.com's password: 
-bash-5.1$ ls -al ibmi-bob/bin
total 48
drwxr-sr-x 2 reinhard 0  8192 Jan 26 21:28 .
drwxr-sr-x 8 reinhard 0 24576 Jan 26 21:31 ..
-rw-r--r-- 1 reinhard 0  1716 Jan 26 21:28 crtfrmstmf
-rw-r--r-- 1 reinhard 0  1720 Jan 26 21:28 makei
```

**Environment (please complete the following information):**
 - Extension version [e.g. 1.6.12]
 - IBM i OS version [e.g. 7.3]

sebjulliand:
@worksofliam : I've got an idea to fix this. I'll try something 😊

PR #1265, opened 2023-04-30
Use tarball to transfer files to deploy

sebjulliand:
### Changes
Files to be deployed are put in a tarball (.tar archive) before being sent.
The benefits are:
- Faster deployment times, especially when deploying a lot of files
- Simpler deployment process (directory creations is handled by the `tar` command)
- File attributes and access permissions are kept during the transfer (if deploying from a POSIX os; not applicable for Windows)

Will fix https://github.com/halcyon-tech/vscode-ibmi/issues/1026

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/local/deployment.ts
+++ b/src/api/local/deployment.ts
@@ -536,7 +536,10 @@ export namespace Deployment {
 
       progress?.report({ message: `extracting deployment tarball to ${parameters.remotePath}...` });
       //Extract and remove tar's PaxHeader metadata folder
-      await connection.sendCommand({ command: `tar -xf ${remoteTarball} && rm -rf PaxHeader`, directory: parameters.remotePath });
+      const result = await connection.sendCommand({ command: `${connection.remoteFeatures.tar} -xf ${remoteTarball} && rm -rf PaxHeader`, directory: parameters.remotePath });
+      if(result.code !== 0){
+        throw new Error(`Tarball extraction failed: ${result.stderr}`)
+      }
 
       const entries: string[] = [];
       tar.t({ sync: true, file: localTarball.name, onentry: entry => entries.push(entry.path) });
--- a/src/api/local/deployment.ts
+++ b/src/api/local/deployment.ts
@@ -1,4 +1,5 @@
-
+import tar from 'tar';
+import tmp from 'tmp';
 import path, { basename, dirname, posix } from 'path';
 import vscode, { WorkspaceFolder } from 'vscode';
 import { getLocalActions } from './actions';
@@ -7,7 +8,7 @@ import { LocalLanguageActions } from '../../schemas/LocalLanguageActions';
 import { instance } from '../../instantiate';
 import ignore from 'ignore'
 import { NodeSSH } from 'node-ssh';
-import { readFileSync } from 'fs';
+import { createWriteStream, readFileSync } from 'fs';
 import Crypto from 'crypto';
 import IBMi from '../IBMi';
 import { DeploymentMethod, DeploymentParameters } from '../../typings';
@@ -205,7 +206,7 @@ export namespace Deployment {
           }
         }
       } else {
-        if(await vscode.window.showErrorMessage(`Chosen location (${folder.uri.fsPath}) is not configured for deployment.`, 'Set deploy location')){          
+        if (await vscode.window.showErrorMessage(`Chosen location (${folder.uri.fsPath}) is not configured for deployment.`, 'Set deploy location')) {
           setDeployLocation(undefined, folder, buildPossibleDeploymentDirectory(folder));
         }
       }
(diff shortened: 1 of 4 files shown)


src/api/local/deployment.ts
```diff
@@ -605,15 +510,46 @@ export namespace Deployment {
...
+      tar.create({ cwd: parameters.workspaceFolder.uri.fsPath, sync: true, file: localTarball.name }, toSend);
+      deploymentLog.appendLine(`Created deployment tarball ${localTarball.name}`);
+
+      progress?.report({ message: `sending deployment tarball...` });
+      await connection.client.putFile(localTarball.name, remoteTarball);
+      deploymentLog.appendLine(`Uploaded deployment tarball as ${remoteTarball}`);
+
+      progress?.report({ message: `extracting deployment tarball to ${parameters.remotePath}...` });
+      //Extract and remove tar's PaxHeader metadata folder
+      await connection.sendCommand({ command: `tar -xf ${remoteTarball} && rm -rf PaxHeader`, directory: parameters.remotePath });
```
worksofliam, line 539: Which version of tar are we expecting to use here? More notes in final review.

worksofliam (changes requested):
Before I approve, I think we need to be more explicit about which version of `tar` we should be using. There is the system shipped version, but also the version available in the IBM yum repo.

```
-bash-5.1$ /usr/bin/tar
Usage: tar -{c|r|t|u|x} [ -BdDEFhilmopRsUvwZ ] [ -Number ] [ -f TarFile ]
           [ -b Blocks ] [ -S [ Feet ] | [ Feet@Density ] | [ Blocksb ] ]
           [ -L InputList ] [-X ExcludeFile] [ -N Blocks ] [ -C Directory ] File ...
Usage: tar {c|r|t|u|x} [ bBdDEfFhilLXmNopRsSUvwZ[0-9] ]
           [ Blocks ] [ TarFile ] [ InputList ] [ ExcludeFile ] 
           [ [ Feet ] | [ Feet@Density ] | [ Blocksb ] ] [-C Directory ] File ...
-bash-5.1$ /QOpenSys/pkgs/bin/tar
/QOpenSys/pkgs/bin/tar: You must specify one of the '-Acdtrux', '--delete' or '--test-label' options
```

I think we need to update our internal API to ensure that `tar` is installed, which means adding `tar` to the `remoteApps` under `/usr/bin`. See `IBMi.ts:25`.

Thoughts?

worksofliam (commented):
I've just done some functional testing and....... I am blown AWAY lol. This is so amazing.

Just to confirm, Seb - you are Windows right? I am mac and this is working very well.

src/api/local/deployment.ts
```diff
@@ -605,15 +510,46 @@ export namespace Deployment {
...
+
+      const entries: string[] = [];
+      tar.t({ sync: true, file: localTarball.name, onentry: entry => entries.push(entry.path) });
+      deploymentLog.appendLine(`${entries.length} file(s) uploaded to ${parameters.remotePath}`);
+      entries.sort().map(e => `\t${e}`).forEach(deploymentLog.appendLine);
+    }
+    finally {
+      deploymentLog.appendLine('');
+      await connection.sendCommand({ command: `rm ${remoteTarball}` })
```
worksofliam, line 548: I see you are deleting the remote file - this is great. Should we also delete the local `.tar` as well?

sebjulliand: We do! This line below takes care of it:
`localTarball.removeCallback();`

src/api/local/deployment.ts
sebjulliand: That's the beauty of it: it doesn't mater.

sebjulliand:
> Before I approve, I think we need to be more explicit about which version of `tar` we should be using. There is the system shipped version, but also the version available in the IBM yum repo.
> 
> ```
> -bash-5.1$ /usr/bin/tar
> Usage: tar -{c|r|t|u|x} [ -BdDEFhilmopRsUvwZ ] [ -Number ] [ -f TarFile ]
>            [ -b Blocks ] [ -S [ Feet ] | [ Feet@Density ] | [ Blocksb ] ]
>            [ -L InputList ] [-X ExcludeFile] [ -N Blocks ] [ -C Directory ] File ...
> Usage: tar {c|r|t|u|x} [ bBdDEfFhilLXmNopRsSUvwZ[0-9] ]
>            [ Blocks ] [ TarFile ] [ InputList ] [ ExcludeFile ] 
>            [ [ Feet ] | [ Feet@Density ] | [ Blocksb ] ] [-C Directory ] File ...
> -bash-5.1$ /QOpenSys/pkgs/bin/tar
> /QOpenSys/pkgs/bin/tar: You must specify one of the '-Acdtrux', '--delete' or '--test-label' options
> ```
> 
> I think we need to update our internal API to ensure that `tar` is installed, which means adding `tar` to the `remoteApps` under `/usr/bin`. See `IBMi.ts:25`.
> 
> Thoughts?

I doesn't mater which one is found and used. Both are supporting the POSIX flags `x` and `f`. And I think tar is always available out of the box. But it wouldn't hurt adding to the `remoteApps` 😊

sebjulliand:
> I've just done some functional testing and....... I am blown AWAY lol. This is so amazing.
> 
> Just to confirm, Seb - you are Windows right? I am mac and this is working very well.

Thanks! Yes, I'm on Windows and I tested it on Linux too; it works very well in both cases.
Thankfully, the JS `tar` implementation is os independent.

sebjulliand:
@worksofliam I added tar to the APi's `remoteFeatures`, it will look in both possible locations.
I tested both just to be sure and they work the same, as expected. I'll let you give it a try!

worksofliam (approved):
I am happy with my mac testing. Tried every deploy option and it's all working smoothly. If you are happy that it works on Windows Seb, then let's merge.

Thanks for your great work here!

p.s. I tried with dumbo_app and the improvement is major. You should compare for yourself. You might get a chuckle from it.

Issue #1255, opened 2023-04-25
Code for IBM i object browser filter error


bigsdavid:
I've installed and completely wiped/reinstalled VS Code 1.77.3 + Code for IBM i, and after creating a filter for a source file under "object browser",  expanding the object browser filter returns "Cannot set properties of undefined (setting 'from')"

worksofliam:
@bigsdavid please share your VS Code settings JSON.

worksofliam:
@bigsdavid Please also share what version of Code for IBM i you are using?

bigsdavid:
Code for IBM i version 1.8.0

User settings JSON:  Problem is in the code-for-ibmi.connectionSettings for "DEV1". Opening the source code filter in the Object Browser I created throws the "Cannot set properties of undefined (setting 'from')" error when I attempt to open the filter. The source code filter for the PUB400 connection works - does not throw this error.

{
    "zowe.ds.history": {
        "persistence": true,
        "favorites": [],
        "history": [],
        "sessions": [
            "zxplore2022"
        ],
        "fileHistory": [],
        "searchHistory": []
    },
    "zowe.uss.history": {
        "persistence": true,
        "favorites": [],
        "history": [],
        "sessions": [
            "zxplore2022"
        ],
        "fileHistory": []
    },
    "zowe.jobs.history": {
        "persistence": true,
        "favorites": [],
        "history": [],
        "sessions": [
            "zxplore2022"
        ],
        "fileHistory": []
    },
    "code-for-ibmi.connections": [
        {
            "name": "DEV1",
            "host": "10.6.0.2",
            "port": 22,
            "username": "d_wilson",
            "privateKey": ""
        },
        {
            "name": "PUB400",
            "host": "pub400.com",
            "port": 2222,
            "username": "wilsond",
            "privateKey": "",
            "buttons": "submitButton",
            "password": "Billet#quandry_411",
            "keepaliveInterval": 35000
        }
    ],
    "code-for-ibmi.connectionSettings": [
        {
            "name": "DEV1",
            "host": "",
            "objectFilters": [
                {
                    "name": "Development source",
                    "library": "SOURCEDEV",
                    "object": "Q*",
                    "types": [
                        "*SRCPF"
                    ],
                    "member": "*",
                    "memberType": "*",
                    "protected": false,
                    "buttons": "SAVE"
                }
            ],
            "libraryList": [
                "GPLIB",
                "QGPL",
                "VLDLSVR",
[...]

PR #1270, opened 2023-05-04
Add check for additional user content in shell output

chrjorgensen:
### Changes

This PR will add a check for additional user content in the output from shell commands during connection, thus eliminating issues like #1255 and others with similar root cause.

If the user has a shell start script with `echo` or any other command, that sends output to stdout, then Code will not be able to interpret the output correctly, and the user will see some mysterious errors - see #1255. This new check will ensure no such additional user content is present, else an error will be issued and the connection attempt aborted.

billede

billede

billede

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -183,7 +183,7 @@ export default class IBMi {
         });
         if (checkShellResult.stderr || checkShellResult.stdout.split(`\n`)[0] !== checkShellText) {
           await vscode.window.showErrorMessage(`Error in shell configuration!`, {
-            detail: `This extension will not function properly, since the output from shell commands have additional content. This can be caused by running commands like "echo" or other\ncommands creating output in your shell start script.\n\nCode for IBM i cannot function correctly and will exit.`,
+            detail: `This extension can not work with the shell configured on ${this.currentConnectionName},\nsince the output from shell commands have additional content.\nThis can be caused by running commands like "echo" or other\ncommands creating output in your shell start script.\n\nThe connection to ${this.currentConnectionName} will be aborted.`,
             modal: true
             });
           throw(`Shell config error, connection aborted.`);
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -172,6 +172,23 @@ export default class IBMi {
         // Reload server settings?
         const quickConnect = (this.config.quickConnect === true && reloadServerSettings === false);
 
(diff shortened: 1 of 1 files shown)


chrjorgensen:
When this PR is merged into master, the check will be included in Quick Connect in an upcoming PR `Quick Connect part II`... 😉

worksofliam:
@chrjorgensen I like it a lot! How can I test this? Is it as simple as writing an `echo` in my `.profile`?

chrjorgensen:
@worksofliam I inserted an echo as the first line in my .bashrc file...

worksofliam:
@chrjorgensen Awesome. This catches that great! We've had many issues, and you yourself just helped someone solve this recently.

My only thought is that I don't know if I want it to disconnect the user. Even if this error occurs, we can still open streamfiles in the IFS to fix the issue.

Perhaps instead of disconnecting them, we can warn the user that until the issue is fixed, Code for IBM i may not function correctly. We do this when a home directory isn't setup correctly.

chrjorgensen:
@worksofliam 

> My only thought is that I don't know if I want it to disconnect the user. Even if this error occurs, we can still open streamfiles in the IFS to fix the issue.

I understand, but I was not sure IFS files could be opened with this configuration and wanted to avoid any mysterious errors in Code. But I will change it to just issue a warning "that everything may break"... 😆

sebjulliand (approved):
Change is good and works as expected, I'll approve it.
That's a clever check @chrjorgensen , it will probably help avoiding a lot of common issues 😅

chrjorgensen:
> @worksofliam
> 
> > My only thought is that I don't know if I want it to disconnect the user. Even if this error occurs, we can still open streamfiles in the IFS to fix the issue.
> 
> I understand, but I was not sure IFS files could be opened with this configuration and wanted to avoid any mysterious errors in Code. But I will change it to just issue a warning "that everything may break"... 😆

@worksofliam I just did some testing, having Code continue with configuration error - and it is not possible to edit the IFS streamfile(s) with this config! The IFS browser shows "Error loading objects" and sends this error:

billede

And Ctrl-Alt-P returns this:

billede

So I think that a disconnect is the only sensible option...

**If you agree, will you merge this PR?**

worksofliam:
I also tested this yesterday and forgot to reply here. You are completely right! Let's merge.

PR #1272, opened 2023-05-05
Exclude src/, types/ and tsconfig.json from vsix

sebjulliand:
### Changes
Updated .vscodeignore to exclude the following folders/files from the vsix build:
- `src/`
- `types/`
- `tsconfig.json`

This reduces the size of the vsix from ~400KB to ~200KB.

### Checklist
* [x] have tested my change


worksofliam (approved):
A small but welcome change. Thank you @sebjulliand!

Issue #1299, opened 2023-05-23
Deleting a file with a SYMLNK causes IFS Browser to fail


SJLennon:
If I delete a file with a symbolic link the IFS Browser cannot display the filter.
To recreate:
```
touch file1
ln -s file1 symlnk
```
![image](https://github.com/halcyon-tech/vscode-ibmi/assets/67484051/77f247ad-ac0b-49b4-9edc-16b25993548b)
Delete the file:
![image](https://github.com/halcyon-tech/vscode-ibmi/assets/67484051/a729172e-26e7-4380-8ad5-9949c4c6f4e8)
Result:  File deleted, SYMLNK remains and IFS browser is no longer functional.
![image](https://github.com/halcyon-tech/vscode-ibmi/assets/67484051/2ae64355-6b7a-44c9-bd9a-e5f5e0125573)
If I delete the SYMLNK external to VS Code, then the IFS browser returns to normal.

* QCCSID: 37
* Features:
   * git: true
   * grep: true
   * tn5250: true
   * setccsid: true
   * md5sum: true
   * bash: true
   * chsh: true
   * GENCMDXML.PGM: true
   * GETNEWLIBL.PGM: true
   * QZDFMDB2.PGM: true
   * startDebugService.sh: true
   * attr: true
   * iconv: true
* SQL enabled: true
* Source dates enabled: false

Variants
```json
{
  "american": "%23@$",
  "local": "%23@$"
}
```

Errors:
```json
[
  {
    "command": "system \"CRTLIB LIB(LENNONS2) TEXT('Code for i temporary objects. May be cleared.')\"",
    "code": 255,
    "stderr": "CPD0032: Not authorized to command CRTLIB in library *LIBL.\nCPF0006: Errors occurred in command.",
    "cwd": "/home/LENNONS"
  },
  {
    "command": "system \"DLTOBJ OBJ(LENNONS2/O_*) OBJTYPE(*FILE)\"",
    "code": 255,
    "stderr": "CPF2125: No objects deleted.",
    "cwd": "/home/LENNONS"
  }
]
```

sebjulliand:
@worksofliam , @chrjorgensen I'll fix that. However, I need your point of view so we can agree on the solution
That's the `--dereference` or `-L` flag making `stat` or `ls` fail. So I can think of three approaches to fix this:
1. Following symlinks isn't useful for the IFS browser, we take out the `--dereference` flag for both `stat` and `ls`: case closed
[...]

PR #1302, opened 2023-05-24
Do not prevent files listing if ls/stat has error(s)

sebjulliand:
### Changes
Fixes issue https://github.com/halcyon-tech/vscode-ibmi/issues/1299
This PR lets the `getFileList` method parse `ls` or `stat` output even if the call return code is not 0. Error output is reported as error notifications (one per error).

### Checklist
* [x] have tested my change


chrjorgensen (changes requested):
@sebjulliand It works almost as if I wrote it...! 🤣 

Very nice - just one minor thing: Can you make ignore the following error in `stat`:

billede

This appears when opening an empty directory. Maybe it could be filtered out of stderr?

Another minor issue, which may not be a new issue, is this error when opening a symlink to a directory using `ls`:

billede

billede

This is not a problem when using `stat`:

billede

sebjulliand:
Thanks for the feedback @chrjorgensen, right on point as always 😊
The error when stat runs on an empty directory is gone again.

And I figured out that the error raised when opening a symlink to a directory using `ls` was occurring with `/usr/bin/ls` but not with `/QOpenSys/pkgs/bin/ls`. So I added `ls` to the connections' Remote Feature, looking in both locations in the event it's not found in `/QOpenSys/pkgs/bin`. I also reordered the locations in the Remote Feature so that `/QOpenSys/pkgs/bin` are used in priority over `/usr/bin`'s.

Everything should be fine now. Let me know how it goes 😉

chrjorgensen (changes requested):
@sebjulliand Perfect - code works as requested. I especially like your ordering of the remote features, so the OSS tools are preferred over the PASE tools! 🚀 

But why the commit https://github.com/halcyon-tech/vscode-ibmi/pull/1302/commits/e25aff277f8c3871364ce07b7b8ca468b8c278fa? It does not seem related to the ls/stat issue...?

sebjulliand:
> @sebjulliand Perfect - code works as requested. I especially like your ordering of the remote features, so the OSS tools are preferred over the PASE tools! 🚀
> 
> But why the commit [e25aff2](https://github.com/halcyon-tech/vscode-ibmi/commit/e25aff277f8c3871364ce07b7b8ca468b8c278fa)? It does not seem related to the ls/stat issue...?

Ugh, I activated the `reorganize import on save` action and those two were reorganize...so I thought it wouldn't hurt to have them pushed as well 😅

chrjorgensen (changes requested):
@sebjulliand What happened to the commit https://github.com/halcyon-tech/vscode-ibmi/pull/1302/commits/ffc5b9240e66227fa02e8732e873b35c66122b54? Seems like it is not in the code anymore - did you do a merge/rebase on the master branch and forgot this?

sebjulliand:
> @sebjulliand What happened to the commit [ffc5b92](https://github.com/halcyon-tech/vscode-ibmi/commit/ffc5b9240e66227fa02e8732e873b35c66122b54)? Seems like it is not in the code anymore - did you do a merge/rebase on the master branch and forgot this?

Gasp, you're right! I did a rebase that went wrong apparently and I didn't merge this part properly. I restored it, it should be fine now.

Thank you and your keen eyes @chrjorgensen !

chrjorgensen (approved):
Perfect! 🎉

PR #1303, opened 2023-05-24
Fix clAPI module not found error

sebjulliand:
### Changes
The following error would show up if the extension needed to load the CL API:
`rejected promise not handled within 1 second: Error: Cannot find module './languages/clle/clAPI'`.

This PR fixes the import and the error.

### Checklist
* [ ] have tested my change


worksofliam:
@sebjulliand Good find! What is the reason I wasn't seeing this on my Macbook when developing and what caused you to find it?

sebjulliand:
> @sebjulliand Good find! What is the reason I wasn't seeing this on my Macbook when developing and what caused you to find it?

I still wonder why. So far I tested it on my desktop running on Windows, who doesn't care about file name case.
But today I'm working on my laptop running on Linux and the error popped. It should have crashed on MacOS for the same reason, unless your CL extension was not installed or active, in which case the line doesn't get executed.

worksofliam (approved):
Simple change and works as expected.

PR #1316, opened 2023-05-28
feat: Add versions and platform to issue

chrjorgensen:
### Changes

This PR will add the versions of this extension, VS Code and the platform running it to the issue details, when a user created an issue on GitHub from the Help panel.

This could be the start of having more debug information available. We often ask the same questions to users creating issues, and it would be time-saving to have this information available from the beginning.

What else comes to mind? I think of the following:
- the help view should maybe always be present and not just when connected - the user may need help or guidance before connecting to any system.
- the last errors shown are from the connection time. These should always be updated, so we have the errors from the situation when the user is having trouble.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -49,7 +49,7 @@ export async function activate(context: ExtensionContext): Promise<CodeForIBMi> 
     ),
     window.registerTreeDataProvider(
       `helpView`,
-      new HelpView()
+      new HelpView(context)
     ),
     window.registerTreeDataProvider(
       `libraryListView`,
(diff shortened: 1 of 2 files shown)


sebjulliand:
That's a very good enhancement; we definitely need more debug information/traces here an there.
I'll share what comes to my mind on this topic later 😉

worksofliam:
Beautiful! I didn't even know the package.json was accessible im through the context.

Thank you @chrjorgensen!

PR #1317, opened 2023-05-28
Fix member hover change timestamp when using SQL

chrjorgensen:
### Changes

This PR will change the member change date and time to show LAST_SOURCE_UPDATE_TIMESTAMP instead of LAST_CHANGE_TIMESTAMP when using SQL - this is the same value as shown when SQL is disabled.

Times are now shown in ISO format `YYYY-MM-DD HH:MM:SS`.
Member timestamps have no timestamp, but the streamfile timestamp will consider the timezone.

### Checklist

* [x] have tested my change


worksofliam (commented):
Tested and working as expected. Thank you!

worksofliam (approved):

PR #1318, opened 2023-05-29
Issue reporting enhancements

sebjulliand:
#### Issue reporting template reorganization
- The `Help & Support` view is now always displayed (not only when connected)
- Issue template's content depends on the extension connection status
- The template now organize information in collapsed sections to improve readability
image
- Some data are now listed in tables to improve readability

#### New information added to the report:
- Platform's architecture
- List of active extensions
- IBM i OS version and TR number
- Shell `env` output

### Checklist

* [x] have tested my change

### DIFF
--- a/src/views/helpView.ts
+++ b/src/views/helpView.ts
@@ -14,9 +14,9 @@ export class HelpView {
 
   public async getChildren(): Promise<HelpItem[]> {
     return [
-      new HelpOpenUrkItem(`book`, `Get started`, `https://halcyon-tech.github.io/vscode-ibmi/#/`),
-      new HelpOpenUrkItem(`output`, `Open official Forum`, `https://github.com/halcyon-tech/vscode-ibmi/discussions`),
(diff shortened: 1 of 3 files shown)


src/views/helpView.ts
```diff
@@ -67,3 +43,131 @@ class HelpItem extends vscode.TreeItem {
...
+    ``,
+    `<hr />`,
+    ``,
+    await getRemoteSection(),
+  ].join(`\n`);
+
+  let issueUrl = encodeURIComponent(issue);
+  if (issueUrl.length > 8130) {
+    //Empirically tested: issueUrl must not exceed 8130 characters
+    issueUrl = issueUrl.substring(0, 8130);
```
worksofliam, line 81: I understand this. I think it might be worth copying the body to the clipboard before the `substring` here and mentioning to the user that the issue data was also copied to the clipboard because it was too long. This can be in the future for sure.

worksofliam (approved):
A beautiful change. One that will absolutely make debugging easier for us.

Thank you!

src/views/helpView.ts
sebjulliand: Great idea!

PR #1319, opened 2023-05-29
Minor issue reports enhancements

sebjulliand:
### Changes
- Added warning to remove sensitive content
- Added notification to copy full data in clipboard if too long
- Added a few emojis

### Checklist
* [x] have tested my change

### DIFF
--- a/src/views/helpView.ts
+++ b/src/views/helpView.ts
@@ -167,9 +167,10 @@ function getRemoteFeatures(connection : IBMi) {
         features.set(featurePath.dir, featureDir);
       }
       featureDir.push(featurePath.base);
+      featureDir.sort();
     }
   });
-
+  
   const maxLine = Array.from(features.values()).map(e => e.length).sort().reverse()[0];
   const dirs = Array.from(features.keys());
   const rows = [];
--- a/src/views/helpView.ts
+++ b/src/views/helpView.ts
@@ -1,4 +1,5 @@
 
+import { parse } from 'path';
 import vscode from 'vscode';
 import IBMi from '../api/IBMi';
 import { instance } from '../instantiate';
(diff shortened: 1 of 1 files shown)


src/views/helpView.ts
```diff
@@ -78,6 +80,10 @@ async function openNewIssue() {
   let issueUrl = encodeURIComponent(issue);
   if (issueUrl.length > 8130) {
     //Empirically tested: issueUrl must not exceed 8130 characters
+    if(await vscode.window.showWarningMessage("Issue data is too long. It will be truncated.", "Copy full data to clipboard")){
```
chrjorgensen, line 83: I did not test this - don't know how to create an URL longer than 8130 chars... 😄

chrjorgensen (changes requested):
@sebjulliand Looking good - you are being creative with emojis! 😆 

Could we have the text `QCCSID` changed to `CCSID`? I know this has been like that for a long time, but technically it is no longer the system value deciding the CCSID for Code for iBM i, instead it's from the user profile...

src/views/helpView.ts
sebjulliand: It shouldn't happen for the moment...it's future proofing 😋
I tested it by changing 8130 to 10 in the condition to trigger it. But for now, in "production", it should not happen.

chrjorgensen: He he...

chrjorgensen (changes requested):
@sebjulliand I'm so sorry, but I have an additional request:

You changed some of the remote features to look in different places (e.g. `ls` and `stat`).
It would be nice to have the qualified remote feature in the debug info...

Other than that, it looks perfect! And I promise: I'll stop after this! 😉 💐

sebjulliand:
How about this @chrjorgensen ?
image

chrjorgensen:
@sebjulliand Nicely arranged in a table - I get more than I ask for... 😆 

But there seems to be an error:

I have the following remote features on a system:

billede

but I get the following content in the issue body:

billede

chrjorgensen:
`maxLine` contains the value `3`:

billede

chrjorgensen:
It's probably because I have 10 values in one array, and your `sort` is not by numeric value?

billede

Should it be

```javascript
..sort( (a, b) => {return a - b} )
```

chrjorgensen (approved):
@sebjulliand Perfect - this will help us a lot when debugging user error reports. 😍

Issue #1325, opened 2023-05-30
IFS Shortcut to non-existant path still shows objects


SJLennon:
To re-create
Create an IFS shortcut to a directory that doesn't exist:
![image](https://github.com/halcyon-tech/vscode-ibmi/assets/67484051/9308d895-22c6-46f7-a4b0-97a5c17e9d74)

Shortcut is created:
![image](https://github.com/halcyon-tech/vscode-ibmi/assets/67484051/31c95006-8b4f-448d-a6e0-ce81f59a7e15)

Expand the shortcut and see directories and files that don't exist
![image](https://github.com/halcyon-tech/vscode-ibmi/assets/67484051/39bf3e51-ad15-4a1e-b386-b3178bbd43f4)

Keep expanding and you can do so (seemingly) indefinitely.
![image](https://github.com/halcyon-tech/vscode-ibmi/assets/67484051/fc09de59-5078-4e27-bbea-9cf6b51d2ebd)

The files that turn up appear to be those in my home directory:
![image](https://github.com/halcyon-tech/vscode-ibmi/assets/67484051/d3cba381-f571-4e78-9b03-4114ab15a394)

<hr />

|Context|Version|
|-|-|
|Code for IBM i version|1.9.0|
|Visual Studio Code version|1.78.2|
|Operating System|win32_x64|

<details>
<summary>Active extensions</summary>

```
Code for IBM i Walkthroughs (vscode-ibmi-walkthroughs): 0.3.1
Db2 for IBM i (vscode-db2i): 0.2.11
Emmet (emmet): 1.0.0
Error Lens (errorlens): 3.11.0
Git (git): 1.0.0
Git Base (git-base): 1.0.0
GitHub (github): 0.0.1
GitHub Authentication (github-authentication): 0.0.2
GitLens — Git supercharged (gitlens): 13.6.0
IBM i Notebooks (vscode-ibmi-notebooks): 0.0.5
JSON Language Features (json-language-features): 1.0.0
Merge Conflict (merge-conflict): 1.0.0
Microsoft Account (microsoft-authentication): 0.0.1
Node Debug Auto-attach (debug-auto-launch): 1.0.0
Overtype (overtype): 0.5.0
TypeScript and JavaScript Language Features (typescript-language-features): 1.0.0
WSL: Recommender (remote-wsl-recommender): 0.0.19
```
</details>

<hr />

<details>
<summary>Remote system</summary>

|Setting|Value|
|-|-|
[...]

PR #1327, opened 2023-05-31
Fix display of empty and non-existing directories

chrjorgensen:
### Changes

This PR will fix the display of non-existing and empty directories in the IFS browser. Now errors will be caught for `stat` output:

billede

And the list of files will only be generated when the directory has content (`stdout` is not blank).

billede

The PR was created to fix issue #1325.

Additional tests and fixes:
- Test for getting non-existing IFS file
- Fix test getFileList for QSYS.LIB (path was not valid, owner was added)
- Fix test of runCommand API
- Allow lowercase member and extension in getMemberList (previously only library and sourcefile were uppercased)
- Fix test of getMemberList (SQL, no filter) (number of members in QSYSINC/MIH varied across releases)
- Pad member dates and times with zeros (non-SQL) (made timestamp creation fail)
- Test for valid directory when creating IFS shortcut (as suggested by @sebjulliand)

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -455,8 +455,8 @@ export default class IBMiContent {
 
     const library = lib.toUpperCase();
     const sourceFile = spf.toUpperCase();
-    let member = (mbr !== `*` ? mbr : null);
-    let memberExt = (ext !== `*` ? ext : null);
+    let member = (mbr !== `*` ? mbr.toUpperCase() : null);
+    let memberExt = (ext !== `*` ? ext.toUpperCase() : null);
 
     let results: Tools.DB2Row[];
 
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -749,7 +749,7 @@ export default class IBMiContent {
    */
   getDspfdDate(century: string = `0`, YYMMDD: string = `010101`, HHMMSS: string = `000000`): Date {
     let year: string, month: string, day: string, hours: string, minutes: string, seconds: string;
-    let dateString: string = (century === `1` ? `20` : `19`).concat(YYMMDD).concat(HHMMSS);
+    let dateString: string = (century === `1` ? `20` : `19`).concat(YYMMDD.padStart(6, `0`)).concat(HHMMSS.padStart(6, `0`));
     [, year, month, day, hours, minutes, seconds] = /(\d{4})(\d{2})(\d{2})(\d{2})(\d{2})(\d{2})/.exec(dateString) || [];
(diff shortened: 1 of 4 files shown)


worksofliam:
@chrjorgensen What are the chances for adding a test to ensure the API fails when passed an invalid directory?

sebjulliand (approved):
Code looks good and works as expected. Good one!

I'd have one suggestion though: how about adding a test in the `code-for-ibmi.addIFSShortcut` command, to check if the path exist and is an actual directory (by calling `IBMiContent.isDirectrory` method)?

chrjorgensen:
@worksofliam @sebjulliand Adding test for non-existing IFS streamfile/directory revealed several issues and errors, which has now been fixed. Please review... :eyes:

worksofliam:
@chrjorgensen The tests are working great!

I hate to be such a burden, but I noticed one of our internal APIs is using `showErrorMessage`, which I don't believe is something our internal API should be doing and should be handled by the views calling the APIs instead:

https://github.com/chrjorgensen/code-for-ibmi/blob/fix/non-existing-dir-in-IFS-browser/src/api/IBMiContent.ts#L636

Is there any chance this PR could clean it up and throw an error instead of logging them to the user like this?

chrjorgensen:
@worksofliam
I agree that an internal API should not write to the UI - but I'm not sure how to fix this: There could be many errors (non-existing symlinks, authorities etc) and the only this could be passed back to the caller is to change the return structure (e.g. combine array IFSFile and the error information in a new structure). This could be a breaking change and should probably wait for version 2.0.0, right?

@sebjulliand
This is beyond (above?) my TypeScript knowledge, I think. You're our TypeScript guru... 😉  ...what do you think - any ideas?

worksofliam (approved):
I am happy with the test coverage (and fixes, thank you!!) and the fix you have made for the bug.

sebjulliand:
> @sebjulliand This is beyond (above?) my TypeScript knowledge, I think. You're our TypeScript guru... 😉 ...what do you think - any ideas?

Oh you...don't say it too much or I'll start to believe it 😁
It's actually simple: the is a `isDirectory()` method to test if a path is a directory. I was thinking we could call it when adding a shortcut to check if the shortcut points to a directory in the first place. But I can make a PR for that if you wish, once this one gets merged.

worksofliam:
@sebjulliand 

> But I can make a PR for that if you wish, once this one gets merged.

Let's do that! I'd appreciate it. 

I am going to merge this and get a new fix release out.

Issue #1337, opened 2023-06-02
Shell config message showing if home directory doesn't exist


worksofliam:
Signing in with Code for IBM i will show this message if the home directory doesn't exist:

<img width="332" alt="image" src="https://github.com/halcyon-tech/vscode-ibmi/assets/3708366/04706e9b-b65a-4d99-9f09-fa036c3054ea">

It should not do this, as we have other logic to validate the home directory after this stage.

Tested with:

```cl
CHGUSRPRF USRPRF(XXX) HOMEDIR('/home/BADDDDDD')
```

worksofliam:
I also see an error when I login via SSH:

```
barrymac:~ barry$ ssh xxx@oss73dev
Could not chdir to home directory /home/BADDDDDD: No such file or directory
-bash-5.1$ 
```

PR #1338, opened 2023-06-02
Shell checker: Show error and read more button

worksofliam:
### Changes

Handles #1337.

* Adds error messages (`stderr`) to shell error modal
* Show 'Read more' button to open docs page with more info regarding this error

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -175,18 +175,16 @@ export default class IBMi {
 
         const checkShellText = `This should be the only text!`;
         const checkShellResult = await this.sendCommand({
-          command: `echo "${checkShellText}"`,
-          directory: `/` // Will use root as the working directory for this command
+          command: `echo "${checkShellText}"`
         });
-        if (checkShellResult.stderr || checkShellResult.stdout.split(`\n`)[0] !== checkShellText) {
+        if (checkShellResult.stdout.split(`\n`)[0] !== checkShellText) {
           const chosen = await vscode.window.showErrorMessage(`Error in shell configuration!`, {
             detail: [
               `This extension can not work with the shell configured on ${this.currentConnectionName},`,
               `since the output from shell commands have additional content.`,
               `This can be caused by running commands like "echo" or other`, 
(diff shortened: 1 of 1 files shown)


src/api/IBMi.ts
```diff
@@ -180,16 +176,33 @@ export default class IBMi {
         const checkShellText = `This should be the only text!`;
         const checkShellResult = await this.sendCommand({
           command: `echo "${checkShellText}"`,
-          directory: `.` // Will use whatever the user profile default is
+          directory: `/` // Will use root as the working directory for this command
```
chrjorgensen, line 179: The `directory` has no effect and should be removed - the SSH server is changing directory to the non-existing user home directory before running the command and will return `Could not chdir to home directory /home/xxxxxxx: No such file or directory`

```suggestion
          command: `echo "${checkShellText}"`
        });
        if (checkShellResult.code === 0 && checkShellResult.stdout.split(`\n`)[0] !== checkShellText) {
```

chrjorgensen (changes requested):
@worksofliam Your change was not enough - the SSH server is the one trying to change to the users home directory at login and setting an error message in `stderr` - and error code 0...!!

I have made a suggestion for a fix, that will work. It basically says to only check the shell output in `stdout` if no errors were found.

worksofliam:
@chrjorgensen 

Your suggestion is more fitting and I've made a change so the shell check only looks at the the `stdout` (as it should). We already have logic to warn the user if their home directory is bad! 

image

angelorpa:
another possibility could be used an SQL query to get user profile info, something like this:

`
SELECT USER_NAME, HOME_DIRECTORY, PASE_SHELL_PATH 
FROM QSYS2.USER_INFO
WHERE USER_NAME = 'USER_NAME'
`

if the HOME_DIRECTORY field is not empty so check the directory path value exist

worksofliam:
@angelorpa That's a little bit out of context for this PR, but thanks for the idea nonetheless!

chrjorgensen (approved):
@worksofliam Looking good!

PR #1339, opened 2023-06-02
Check if member can be written before edition

sebjulliand:
### Changes
This PR addresses what's been discussed here: https://github.com/orgs/halcyon-tech/discussions/1331

- Source file write access is now checked when it's expanded in the Object Browser. If it cannot be written to, all members will be opened in read-only mode (and a lock will be shown next to the members).
image
- Write access is also checked when opening a member using the `Go to File` command. The member is open in read-only mode if its source file cannot be written to.

Checking the write access while saving was not implemented since it would show the misleading "Failed to save" dialog.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -5,7 +5,7 @@ import tmp from 'tmp';
 import util from 'util';
 import vscode from "vscode";
 import { ObjectTypes } from '../schemas/Objects';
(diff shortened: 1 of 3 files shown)


chrjorgensen (approved):
@sebjulliand Works perfectly - as expected. Beautiful! 🚀

Issue #1348, opened 2023-06-06
IFS Browser - unable to browse a folder containing a space


jkdavew:
While using the IFS Browser and navigating to a folder containing a space, VSCode shows a couple of errors and the contents of the folder are not shown.

Error
1 error occurred while listing files.
Source: Code for IBM i (Extension).

Error
bsh: cd: too many arguments
Source: Code for IBM i (Extension).

Side note: displaying a file which has a space in the name works.

chrjorgensen:
@jkdavew Please update to the latest version if not already there, and then open the Help view and click `Report an issue` and copy-paste the issue content here...

sebjulliand:
I think that should take care of it: https://github.com/halcyon-tech/vscode-ibmi/pull/1350

jkdavew:
Thank you for the fix!  Now that the repo is updated, what is the deploy schedule to update the plugin in the VSCode Marketplace?  Thx

sebjulliand:
> Thank you for the fix! Now that the repo is updated, what is the deploy schedule to update the plugin in the VSCode Marketplace? Thx

It's being published at this very moment. It should be available in a few minutes.
This is @chrjorgensen first release, too 😎

chrjorgensen:
> This is @chrjorgensen first release, too 😎

Shhhh! Don't tell!! 😉

sebjulliand:
> > This is @chrjorgensen first release, too 😎
> 
> Shhhh! Don't tell!! 😉

This calls for a celebration!

jkdavew:
Amazing first release - worked like a charm!!  Thx team!

PR #1350, opened 2023-06-07
Fixed cd failing when path has whitespaces

sebjulliand:
### Changes
Fixes https://github.com/halcyon-tech/vscode-ibmi/issues/1348

When using `stat` to list a folder's content, the `cd` command run before failed if the target path had whitespaces. Puting quotes around the path fixes the issue.

### Checklist
* [x] have tested my change


chrjorgensen (approved):
@sebjulliand Thanks for the fix!

PR #1351, opened 2023-06-07
Add test for getFileList with special chars

chrjorgensen:
### Changes

Create a test for calling `getFileList` to get files with special characters in their names.

This is to avoid errors like the one encountered in issue #1348.

### Checklist

* [x] have tested my change


sebjulliand (changes requested):
It works fine, good job!
I just have a suggestion to clean things up even if test go wrong.

sebjulliand (approved):
Perfect 😀

PR #1353, opened 2023-06-07
Always start a new terminal for Open Terminal Here

chrjorgensen:
### Changes

When an IBM i PASE terminal has been started previously and exited, the Open Terminal Here function does not work anymore because it selects the exited terminal.

This PR changes the behavior for Open Terminal Here to always start a new IBM i PASE terminal.

It is still possible to use a path in an existing terminal by selecting `Copy Path` and switch to the existing terminal and enter `cd` and paste the path into the terminal.

### Checklist

* [x] have tested my change


sebjulliand (approved):
Simple and efficient change. Let's merge!

Issue #1364, opened 2023-06-10
'Report an Issue' not working


SJLennon:
Save us time! Click 'Report an Issue' from the Help and Support panel inside of VS Code after you have connected to a system.

I tried this and it flashes "Gathering issue details..." and then nothing.   I'm pretty sure I did this in the last couple of weeks and it opened up Firefox and copied in all the useful info you need into a new issue.  I'm on 1.9.4 running on Windows 10. connected to PUB400.COM. 

I've exited VSCode and reconnected and the problem still exists. Hope I'm not overlooking something or doing something stupid.

**Describe the bug**
(A clear and concise description of what the bug is. We will not accept your bug and close it if this is not filled out correctly.)

Los problemas en espanol estan bien.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**
 - Extension version [e.g. 0.x.x]
 - IBM i OS version [e.g. 7.x]

**Additional context**

sebjulliand:
Good catch!
Turns out some of the information we get from the system are not accessible to everybody (OS version and TR level).
This PR will fix this: https://github.com/halcyon-tech/vscode-ibmi/pull/1365

PR #1365, opened 2023-06-11
Fix "Report issue" crashing when not authorized to WRKPTFGRP

sebjulliand:
### Changes
Fixes https://github.com/halcyon-tech/vscode-ibmi/issues/1364
`Report Issue` would crash if the user was not authorized to `WRKPTFGRP` (like many of us on PUB400).
The related SQL query is now surround by `try/catch` to avoid it from crashing the whole process. If the user is not authorized, we'll try to read the `QUSRSYS/QSS1MRI` data area to get the OS version.
If it fails, `OS Version` and `TR level` will simply show `n/a`.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/views/helpView.ts
+++ b/src/views/helpView.ts
@@ -131,11 +131,12 @@ async function getRemoteSection() {
       catch (error) {
         console.log(`Couldn't run QSYS2.GROUP_PTF_INFO: ${error}`);
         try {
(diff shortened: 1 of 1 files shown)


chrjorgensen (approved):
@sebjulliand Looking good to me! :+1:

PR #1379, opened 2023-06-20
Fix Quick Connect remoteFeatures

chrjorgensen:
### Changes

This PR will fix Quick Connect remote features test - `ls` was missing in the `remoteFeatures`, so all remote features were checked at each connection.

I have also created a note at `remoteApps` to remember adding any new name to `remoteFeatures`.

### Checklist

* [x] have tested my change


sebjulliand (approved):
Perfect; that comment will save us a lot of trouble.

Issue #1290, opened 2023-05-17
Compile with multiple actions


jkdavew:
Type: <b>Bug</b>

Line 1 of action has a prompted command (with leading ? symbol)
Line 2 can be any other statemnt

Compile a program and the prompted command will show all lines but should only be prompting the line for which the ? was on.

Extension version: 1.8.0
VS Code version: Code 1.78.2 (b3e4e68a0bc097f0ae7907b217c1119af9e03435, 2023-05-10T14:39:26.248Z)
OS version: Windows_NT x64 10.0.19044
Modes:
Sandboxed: No

jkdavew:
![image](https://github.com/halcyon-tech/vscode-ibmi/assets/15818305/261f92f9-0a03-4727-8670-441379988358)

jkdavew:
@worksofliam thx for having a look at this!

sebjulliand:
This will be fixed with the next release.

PR #1380, opened 2023-06-20
Fix issue #1290

chrjorgensen:
### Changes

This PR will fix issue #1290 by looping through all commands in an action and prompt the ones with `?`.

It will also add a `Cancel` button when prompting for custom inputs.

NOTE: eslint is complaining: Property 'buttons' does not exist on type '{}'.
...and I don't know how to fix that. Help from one of our TS gurus is needed...!

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/CompileTools.ts
+++ b/src/api/CompileTools.ts
@@ -557,10 +557,19 @@ export namespace CompileTools {
         getDefaultVariables(instance, ileSetup)
       );
 
-      if (commandString.startsWith(`?`)) {
-        commandString = await vscode.window.showInputBox({ prompt: `Run Command`, value: commandString.substring(1) }) || '';
-      } else {
-        commandString = await showCustomInputs(`Run Command`, commandString, title);
+      if (commandString) {
+        const commands = commandString.split(`\n`).filter(command => command.trim().length > 0);
+        let promptedCommands = [];
+        for (let command of commands) {
+          if (command.startsWith(`?`)) {
+            command = await vscode.window.showInputBox({ prompt: `Run Command`, value: command.substring(1) }) || '';
+          } else {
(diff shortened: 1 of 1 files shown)


src/api/CompileTools.ts
```diff
@@ -710,12 +719,12 @@ export namespace CompileTools {
         }
       }
 
-      commandUI.addButtons({ id: `execute`, label: `Execute` });
+      commandUI.addButtons({ id: `execute`, label: `Execute` }, { id: `cancel`, label: `Cancel` });
 
       const page = await commandUI.loadPage(name);
```
sebjulliand, line 724: This will fix eslint's complain
```suggestion
      const page = await commandUI.loadPage<any>(name);
```

src/api/CompileTools.ts
```diff
@@ -557,10 +557,19 @@ export namespace CompileTools {
         getDefaultVariables(instance, ileSetup)
       );
 
-      if (commandString.startsWith(`?`)) {
-        commandString = await vscode.window.showInputBox({ prompt: `Run Command`, value: commandString.substring(1) }) || '';
-      } else {
-        commandString = await showCustomInputs(`Run Command`, commandString, title);
+      if (commandString) {
+        const commands = commandString.split(`\n`).filter(command => command.trim().length > 0);
+        let promptedCommands = [];
```
sebjulliand, line 562: Make this explicitely immutable
```suggestion
        const promptedCommands = [];
```

sebjulliand (changes requested):
Code looks good so far.
Only the fix for eslint and a minor change are in order ☺

sebjulliand (approved):
Looks good, let's merge this!

Issue #1235, opened 2023-04-12
Source member sometimes truncates text after column 80


richardm90:
Follow on from issue #1214 

Code for IBM i v1.7.12. I have source dates enabled with a tracking mode of diff.

I connect to my IBM i server in VS Code.

I then create a new source member in QRPGLESRC, which has a record length of 112.

```
**free

snd-msg 'The SND-MSG operation sends an informational message or an exception message.';

*inlr = *on;
return;
```

I don't save the source, I don't try to compile. I now simply close VS Code.

When I open VS Code again I see the tab for my new member with the following message.

![image](https://user-images.githubusercontent.com/7809440/231564252-5bfecb9f-8ce2-41dd-b0fb-cd6bba9d7c3e.png)

I don't click Try Again yet.

I reconnect to my IBM i server and then click Try Again.

I then compile, which does an auto save, and it fails.

![image](https://user-images.githubusercontent.com/7809440/231564287-c48dc497-80d4-40f5-b526-d611ce1dfce5.png)

If I look in PDM I can see the line has been truncated.

![image](https://user-images.githubusercontent.com/7809440/231564328-c2818a1a-f2a2-4817-9895-a71de3d9f5ab.png)

I'm not sure this is the only way to replicate the problem but I wonder if you don't open a member via the object browser then maybe the record length isn't being retrieved.

worksofliam:
This is where it is happening: https://github.com/halcyon-tech/vscode-ibmi/blob/master/src/filesystems/qsys/extendedContent.ts#L121

I suspect because the document remains open unsaved between instances, it doesn't have to open the member again on the second instance, therefore we don't have the record length and we default to 80.

richardm90:
I'm working on a fix for this, I think I'm pretty close to a solution.

richardm90:
This issue has been fixed and addressed as part of pull request #1389.

PR #1389, opened 2023-06-23
Fix source member lines from being truncated

richardm90:
### Changes

The record length of the source member was not always being respected and there were circumstances when the default record length of 80 characters was being assumed, losing characters between position 81 and 100.

See issue #1235 for full details.

When saving a source member if the record length has not been established then it will be determined before the save. The logic to establish the record length has been extracted into a private function called `getRecordLength` within `extendedContest.ts`, as this is used in two places.

### Checklist

* [x] have tested my change


src/filesystems/qsys/extendedContent.ts
```diff
@@ -46,17 +46,8 @@ export class ExtendedIBMiContent {
       } catch (e) { }
 
       if (!this.sourceDateHandler.recordLengths.has(alias)) {
```
worksofliam, line 48: It looks like this is **always** returning false, even if the alias does exist in it.

worksofliam: Ok. My mistake, looks like we are internally deleting the stored source length in `sourceDateHandler.ts:124`.

worksofliam:
@richardm90 Looks good and my testing looks good as well. Fancy fixing this conflict, and then we can merge!

richardm90:
@worksofliam, I've fixed the conflict.

worksofliam (approved):
Thanks @richardm90! Thanks so much for the PR!

PR #1403, opened 2023-06-30
Use temp member to save new file

worksofliam:
### Changes

* Attempts to insert new file into temporary table first. This is good because we can check the file will be saved without clearing out the original member first.
* Seems to have seriously improved the performance of member saving

### Checklist

* [x] have tested my change

### DIFF
--- a/src/filesystems/qsys/extendedContent.ts
+++ b/src/filesystems/qsys/extendedContent.ts
@@ -144,7 +144,10 @@ export class ExtendedIBMiContent {
         }
 
         //We assume the alias still exists....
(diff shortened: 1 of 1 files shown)


chrjorgensen (approved):
Looking good and works perfectly!

chrjorgensen:
I did not feel any speed improve - which would also not be logical, since two more steps is introduced to the save operation.
However, that was not the point either - the point being to have a safer save, not saving half a member. And the additional steps are almost not noticeable and a good compromise for the safer save.

PR #1409, opened 2023-07-03
Fix error when no files are select for upload

chrjorgensen:
### Changes

This PR will fix the error occurring when no files were selected for upload:

image

### Checklist

* [x] have tested my change


sebjulliand (approved):
You could have fixed it by just adding a `?`

```
chosenFiles?.forEach 
```

But a good old `if` is fine too! :)

chrjorgensen:
@sebjulliand `?` was my first thought - but then thought it only applied to TypeScript... :roll_eyes:

PR #1410, opened 2023-07-03
Fix checking for directory in safe delete mode

chrjorgensen:
### Changes

This PR will fix checking if the node is a directory when `Safe Delete Mode` is active:
Check for node context `directory` instead if node path ending in `/`.

### Checklist

* [x] have tested my change


sebjulliand (approved):
Nice change. Simple and makes the process safer. Let's merge!

PR #1417, opened 2023-07-06
Fix parsing of `getTable` result

SanjulaGanepola:
### Changes

When testing the `getLibraryList` API, I noticed that for a library (with a text description of `*BLANK`), the `text` field was `''` when SQL is enabled, but `' '` when disabled. When SQL is disabled, the result which is parsed in `getTable` seems to have `' '` when there is no value. 

image

This PR makes the result consistent regardless of if SQL is enabled.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/testing/content.ts
+++ b/src/testing/content.ts
@@ -5,6 +5,7 @@ import { instance } from "../instantiate";
 import util from 'util';
 import tmp from 'tmp';
 import { CommandResult } from "../typings";
+import { Tools } from "../api/Tools";
 
 export const ContentSuite: TestSuite = {
   name: `Content API tests`,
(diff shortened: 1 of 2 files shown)


src/testing/content.ts
```diff
@@ -179,6 +180,32 @@ export const ContentSuite: TestSuite = {
...
+
+        config!.enableSQL = false;
+
+        // Then we fetch the table without SQL
+        const tableB = await content?.getTable(tempLib, TempName, TempName, true);
+
+        // Reset the config
+        config!.enableSQL = true;
+
+        assert.notDeepStrictEqual(tableA, tableB);
```
worksofliam, line 205: Should this not be `deepStrictEqual`? Don't we want `tableA` and `tableB` to match?

SanjulaGanepola: This is what I thought as well initially, but it seems that certain values in `tableA` are strings whereas in `tableB` they are numbers:

image

SanjulaGanepola: I briefly took a look at the `db2Parse` function and I see that there is some logic to convert to numbers, but it is only being done if the value starts with `' '`. Is there a reason for this?

https://github.com/halcyon-tech/vscode-ibmi/blob/master/src/api/Tools.ts#L103

worksofliam: Yes. Numbers in the result set for the `runSQL` API are right-adjusted and is an easy way for us to determine if the field is a number or not.

worksofliam (approved):
Code is simple and tests continue to pass. Thank you @SanjulaGanepola!

PR #1420, opened 2023-07-07
Feature/disconnectOnDeactivate

sebjulliand:
### Changes
This change will call the `code-for-ibmi.disconnect` command when the extension gets deactivated (i.e. when VSCode closes).
It never hurts to clean behind us right? 😅

The only way I could test this was to put a breakpoint in `disconnect` and close VSCode. But it seems to work though!

### Checklist
* [x] have tested my change


chrjorgensen (approved):
@sebjulliand Looking good and working AFAICT - and cleanup is always good! 👍

Issue #1431, opened 2023-07-12
Protected source members


mprovvedi:
Last week while i was opening source members in VSCode a lock icon appeared next to them (as shown below), without changing any configuration.

![image](https://github.com/halcyon-tech/vscode-ibmi/assets/107923967/2e0f24be-b841-4cb8-a394-cd3db2e1bfed)

I tried to check both the filter settings (the "Protected" checkbox is unchecked) and the connection settings (the "Read only mode" in the "Source Code" tab is unchecked).
This morning i tried to create a new connection and when i was switching between connections to emulate all the settings the same thing happened.
I also tried to create new filters for other source files but the members are still locked.

Finally, i found out that this happens when you put a non exisisting library in the library list on VSCode. So i think that this happened last week because we had to delete one library in the library list.

Is this a feature to prevent to compile with the wrong library list or is this a bug? If this is a feauture, is it possible to show some error messages to enlight the problem?

Thank you!
<hr />

|Context|Version|
|-|-|
|Code for IBM i version|2.0.0|
|Visual Studio Code version|1.80.0|
|Operating System|win32_x64|

<details>
<summary>Active extensions</summary>

```
Account Microsoft (microsoft-authentication): 0.0.1
Autenticazione GitHub (github-authentication): 0.0.2
Code Coverage for IBM i (code-coverage-ibmi): 0.1.7
Code for IBM i Walkthroughs (vscode-ibmi-walkthroughs): 0.3.1
Collegamento automatico per debug di Node (debug-auto-launch): 1.0.0
Creazione estensione (extension-editing): 1.0.0
Db2 for IBM i (vscode-db2i): 0.2.11
Emmet (emmet): 1.0.0
Error Lens (errorlens): 3.11.1
Esegui merge del conflitto (merge-conflict): 1.0.0
Funzionalità dei linguaggi TypeScript e JavaScript (typescript-language-features): 1.0.0
Funzionalità del linguaggio JSON (json-language-features): 1.0.0
GIT (git): 1.0.0
Git Base (git-base): 1.0.0
Git Graph (git-graph): 1.30.0
GitHub (github): 0.0.1
IBM i Notebooks (vscode-ibmi-notebooks): 0.0.5
Material Icon Theme (material-icon-theme): 4.28.0
Monokai Pro (theme-monokai-pro-vscode): 1.2.1
Overtype (overtype): 0.5.0
Path Intellisense (path-intellisense): 2.8.4
[...]

PR #1433, opened 2023-07-12
Validate library list when connecting

chrjorgensen:
### Changes

This PR was created to partially fix issue #1431 by validating the library list when connecting to a system.
If any libraries in the list does not exist, the user will be prompted to remove them from the list:

billede

billede

Library list validation is not done on Quick Connect.

### Checklist

* [x] have tested my change


chrjorgensen:
I know this will not fix the issue of operations failing in the extension if a library in the library list is deleted while connected. The configuration can contain many library lists, e.g. (Command) Profiles, that can cause errors in the extension if not valid.

Ideas of whether or how we should handle this are welcome. 😃

worksofliam (changes requested):
It's working as expected, but I'd like to see a `progress.report` thrown in similar to the other validation operations.

Thanks!

chrjorgensen:
@worksofliam I added the progress message anyway - please review and approve when convenient. 😃

worksofliam (approved):
Tested again to be safe and works as expected.

PR #1435, opened 2023-07-13
Expose `Deployment` namespace in `CodeForIBMi` API

SanjulaGanepola:
### Changes

Instead of just exposing the `deploy` function in the `CodeForIBMi` API, this PR exposes the entire `Deployment` namespace to give access to the other functions. In this PR, I also exported the 4 functions to retrieve files based on the deployment method so they can be reused by other extensions.

### Checklist

* [x] have tested my change


sebjulliand (changes requested):
The change is good and works fine. I tested it from another extension and using the Deployment namespace works.
However, now the entire namespace gets exported, I'd move the `initialize` method out of it since it's not meant to be exported and called.

SanjulaGanepola:
That is a good point. However how can we go about moving the `initialize` function out because it access other functions/variables within the namespace which are not exported.

SanjulaGanepola:
Yeah that can work. Since the other functions are still related to Deployment, I think it would still be a good idea to have them in their own namespace. I think we keep the `Deployment` namespace to contain deployment related variables/helper functions and create a new namespace called `DeployTools` that will contain the functions we would like to have in the exported API.

@worksofliam @sebjulliand Any issues with this or can I go ahead and make this change?

sebjulliand:
> Yeah that can work. Since the other functions are still related to Deployment, I think it would still be a good idea to have them in their own namespace. I think we keep the `Deployment` namespace to contain deployment related variables/helper functions and create a new namespace called `DeployTools` that will contain the functions we would like to have in the exported API.
> 
> @worksofliam @sebjulliand Any issues with this or can I go ahead and make this change?

Sounds good to me!

SanjulaGanepola:
PR is ready for review.

**Changes:**
- `Deployment` namespace contains the deployment related variables/helper functions
- `DeployTools` namespace contains the APIs to be exposed for other extensions to use. I also included some more functions here compared to before as other extensions may want to use them
- Updated the `CodeForIBMi` to include `DeployTools`
- Renamed Deployment tests to Deploy Tools

sebjulliand (approved):
Perfect! Looks good to me and tests are still OK.

PR #1460, opened 2023-07-30
Ability to create home directory

### LINKED ISSUE #1205: Better information or automation relating to home directory
When the user profile used doesn't have a home directory, a notification pops up that says the extension might not work properly without a home directory. That part is fine. Then it suggests assigning a home directory with the CHGUSRPRF command. That is also fine. But if the actual directory in the IFS does not exist, the message doesn't provide an explanation or remedy. It is especially confusing if the user has already done the suggested step of issuing CHGUSRPRF.

So the minimum I'm suggesting is to also include, in the message, the command to create the home directory.

But what would be even better, if feasible, is to do those things automatically (change the user profile and/or create the home directory), after prompting the user and getting their OK. If they decline, then of course don't do those things. If they would like to have those things done automatically but don't have sufficient system privileges, then they should be informed of that, with the suggestion that they contact their administrator to do those things for them, or grant them the required authority.

* QCCSID: 65535
* Features:
   * git: true
   * grep: true
   * tn5250: true
   * setccsid: true
   * md5sum: true
   * GENCMDXML.PGM: false
   * QZDFMDB2.PGM: true
   * startDebugService.sh: false
   * iconv: true
   * attr: true
* SQL enabled: false
* Source dates enabled: false

Variants
```json
{
  "american": "%23@$",
  "local": "%23@$"
}
```

Errors:
```json
[
  {
    "command": "system \"CHKOBJ OBJ(QSYS/QCPFRMIMPF) OBJTYPE(*DTAARA)\"",
    "code": 255,
    "stderr": "Could not chdir to home directory /home/JYEUNG: No such file or directory\nCPF9801: Object QCPFRMIMPF in library QSYS not found.",
    "cwd": "."
  },
  {
    "command": "ls -p /QIBM/ProdData/IBMiDebugService/bin/startDebugService.sh",
    "code": 2,
    "stderr": "Could not chdir to home directory /home/JYEUNG: No such file or directory\n/QIBM/ProdData/IBMiDebugService/bin/startDebugService.sh not found",
    "cwd": "."
  },
  {
    "command": "ls -p /QSYS.lib/ILEDITOR.lib/GENCMDXML.PGM",
    "code": 2,
    "stderr": "Could not chdir to home directory /home/JYEUNG: No such file or directory\n/QSYS.lib/ILEDITOR.lib/GENCMDXML.PGM not found",
    "cwd": "."
  }
]
```

ThePrez:
### Changes

Fixes #1205 

Changes:
- Check if home directory exists but is not writeable
- Allow for automatic creation of home directory (after asking the user)
- Added a `console.log` message when creating the home directory (runs counter to PR checklist requirements)
- Remove guidance about giving the user a home directory via `CHGUSRPRF` as I believe it to be incorrect

### Checklist

* [x] have tested my change
* ~~[ ] updated relevant documentation~~ **N/A**
* ~~[ ] Remove any/all `console.log`s I added~~ **(See note above in change list)**

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -213,14 +213,42 @@ export default class IBMi {
 
         let defaultHomeDir;
 
-        const commandResult = await this.sendCommand({
-          command: `pwd`,
+        const echoHomeResult = await this.sendCommand({
+          command: `echo $HOME`,
           directory: `.`
         });
-        if (commandResult.stderr) {
-          defaultHomeDir = undefined;
+        // If the home directory does not exist, stderr contains 'Could not chdir to home directory /home/________: No such file or directory'
+        if (echoHomeResult.stderr) {
+          // even if the home directory doesn't exist, the echo command gives us our real home in stdout
+          let actualHomeDir = echoHomeResult.stdout.trim();
+
+          // we _could_ just assume the home directory doesn't exist but maybe there's something more going on, namely mucked-up permissions
+          let doesHomeExist = (0 === (await this.sendCommand({ command: `test -w $HOME` })).code);
+          if (doesHomeExist) {
+            // home directory exists but we don't have write permission to it
+            await vscode.window.showWarningMessage(`Your home directory (${actualHomeDir}) is unusable. Code for IBM i may not function correctly. Please contact your system administrator`, { modal: !reconnecting });
+          }
+          else if (reconnecting) {
+            vscode.window.showWarningMessage(`Your home directory (${actualHomeDir}) does not exist. Code for IBM i may not function correctly.`, { modal: false });
+          }
+          else if (await vscode.window.showWarningMessage(`Home directory does not exist`, {
+            modal: true,
+            detail: `Your home directory (${actualHomeDir}) does not exist, so Code for IBM i may not function correctly. Would you like to create this directory now?`,
+          }, `Yes`)) {
+            console.log(`creating home directory ${actualHomeDir}`);
+            let mkHomeCmd = `mkdir -p ${actualHomeDir} && chown ${connectionObject.username.toLowerCase()} ${actualHomeDir} && chmod 0755 ${actualHomeDir}`;
+            let mkHomeResult = await this.sendCommand({ command: mkHomeCmd });
+            if (0 === mkHomeResult.code) {
+              defaultHomeDir = actualHomeDir;
+            } else {
+              let mkHomeErrs = mkHomeResult.stderr;
+              // We still get 'Could not chdir to home directory' in stderr so we need to hackily gut that out, as well as the bashisms that are a side effect of our API
+              mkHomeErrs = mkHomeErrs.substring(1 + mkHomeErrs.indexOf(`\n`)).replace(`bash: line 1: `, ``);
+              await vscode.window.showWarningMessage(`Error creating home directory (${actualHomeDir}):\n${mkHomeErrs}.\n\n Code for IBM i may not function correctly. Please contact your system administrator`, { modal: true });
+            }
+          }
         } else {
-          defaultHomeDir = commandResult.stdout.trim();
+          defaultHomeDir = echoHomeResult.stdout.trim();
         }
 
         //Get home directory if one isn't set
@@ -716,7 +744,7 @@ export default class IBMi {
               });
           }
         } else {
-          vscode.window.showWarningMessage(`Code for IBM i may not function correctly until your user has a home directory. Please set a home directory using CHGUSRPRF USRPRF(${connectionObject.username.toUpperCase()}) HOMEDIR('/home/${connectionObject.username.toLowerCase()}')`);
+          vscode.window.showWarningMessage(`Code for IBM i may not function correctly until your user has a home directory.`);
         }
 
         if (!reconnecting) {
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -225,13 +225,8 @@ export default class IBMi {
         //   - The 'cd' command causes an error if the home directory does not exist or otherwise can't be cd'ed into
         //   - The 'test' command causes an error if the home directory is not writable (one can cd into a non-writable directory)
         let isHomeUsable = (0 == echoHomeResult.code);
-        if (!isHomeUsable && quickConnect) {
-          // If quick connect is enabled, just show a non-modal dialog without details and continue
-          let actualHomeDir = echoHomeResult.stdout.trim();
-          vscode.window.showWarningMessage(`Your home directory (${actualHomeDir}) is unusable. Code for IBM i may not function correctly. Please contact your system administrator`, { modal: false });
-        }
-        else if (!isHomeUsable) {
-          // If quick connect is not enabled, let's try to provide more valuable information to the user about why their home directory
+        if (!isHomeUsable) {
+          // Let's try to provide more valuable information to the user about why their home directory
           // is bad and maybe even provide the opportunity to create the home directory
 
           let actualHomeDir = echoHomeResult.stdout.trim();
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -276,9 +276,7 @@ export default class IBMi {
 
         // Check to see if we need to store a new value for the home directory
         if (defaultHomeDir) {
-          if (this.config.homeDirectory === defaultHomeDir) {
-            progress.report({message: `Configured home directory (${defaultHomeDir}) appears usable.`});
-          } else {
+          if (this.config.homeDirectory !== defaultHomeDir) {
             this.config.homeDirectory = defaultHomeDir;
             progress.report({message: `Configured home directory reset to ${defaultHomeDir}.`});
           }
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -225,7 +225,9 @@ export default class IBMi {
         //   - The 'cd' command causes an error if the home directory does not exist or otherwise can't be cd'ed into
         //   - The 'test' command causes an error if the home directory is not writable (one can cd into a non-writable directory)
         let isHomeUsable = (0 == echoHomeResult.code);
-        if (!isHomeUsable) {
+        if (isHomeUsable) {
+          defaultHomeDir = echoHomeResult.stdout.trim();
+        } else {
           // Let's try to provide more valuable information to the user about why their home directory
           // is bad and maybe even provide the opportunity to create the home directory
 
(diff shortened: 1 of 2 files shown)


chrjorgensen:
@ThePrez Thanks for the PR, great enhancement of the home directory validation. I will test this asap. :smiley:

chrjorgensen (changes requested):
@ThePrez Using `~` instead of `$HOME` does not fully work in the default shell `/QOpenSys/usr/bin/bsh`:

```
cd ~
~: No such file or directory
```

We can't assume the users are running `bash` as shell, however much we would like that.
Also, for IBM i users not fully familiar with IFS and shells, I think it's better to have the absolute path from `$HOME` in the IFS browser and not the tilde `~`. And also for us as developers when debugging issues etc.

WDYT?

ThePrez:
> @ThePrez Using `~` instead of `$HOME` does not fully work in the default shell `/QOpenSys/usr/bin/bsh`:
> 
> ```
> cd ~
> ~: No such file or directory
> ```
> 
> We can't assume the users are running `bash` as shell, however much we would like that.
> Also, for IBM i users not fully familiar with IFS and shells, I think it's better to have the absolute path from `$HOME` in the IFS browser and not the tilde `~`. And also for us as developers when debugging issues etc.
> 
> WDYT?

Thank you. Good catch!
I will revert https://github.com/halcyon-tech/vscode-ibmi/commit/7c43e664af2700dc6c1d079e18be7b5137cef0fc

src/api/IBMi.ts
```diff
@@ -213,14 +213,42 @@ export default class IBMi {
...
         });
-        if (commandResult.stderr) {
-          defaultHomeDir = undefined;
+        // If the home directory does not exist, stderr contains 'Could not chdir to home directory /home/________: No such file or directory'
+        if (echoHomeResult.stderr) {
+          // even if the home directory doesn't exist, the echo command gives us our real home in stdout
+          let actualHomeDir = echoHomeResult.stdout.trim();
+
+          // we _could_ just assume the home directory doesn't exist but maybe there's something more going on, namely mucked-up permissions
+          let doesHomeExist = (0 === (await this.sendCommand({ command: `test -w $HOME` })).code);
```
chrjorgensen, line 226: @ThePrez How should the test for write permission to the home directory work? The test is not performed if the `echo` command completes without error - and it never will...

Also, just to check that $HOME points to a directory, maybe your `test` should be changed to `test -d $HOME && test -w $HOME`?

ThePrez: @chrjorgensen, we don't check whether the echo command completes without error, only whether there's something in stdout. Your comment got me rethinking the design, and I'll be pushing up a commit shortly for your opinion. 

Good call on including `test -d`!

ThePrez: Actually I don't think we want the `test -d`. See commit comments

ThePrez: Marking resolved and asking for re-review based on latest commit

src/api/IBMi.ts
```diff
@@ -213,14 +213,60 @@ export default class IBMi {
...
+          // we _could_ just assume the home directory doesn't exist but maybe there's something more going on, namely mucked-up permissions
+          let doesHomeExist = (0 === (await this.sendCommand({ command: `test -e ${actualHomeDir}` })).code);
+          if (doesHomeExist) {
+            // Note: this logic might look backward because we fall into this (failure) leg on what looks like success (home dir exists).
+            //       But, remember, but we only got here if 'cd $HOME' failed.
+            //       Let's try to figure out why....
+            if (0 !== (await this.sendCommand({ command: `test -d ${actualHomeDir}` })).code) {
+              await vscode.window.showWarningMessage(`Your home directory (${actualHomeDir}) is not a directory! Code for IBM i may not function correctly. Please contact your system administrator`, { modal: !reconnecting });
+            }
+            else if (0 !== (await this.sendCommand({ command: `test -w ${actualHomeDir}` })).code) {
```
chrjorgensen, line 238: @ThePrez I'm sorry, but I still have an issue with the test for writable home directory: My test user has a non-writable home directory, but no error shows up!

The writable test is only done if the `cd` command fails - but `cd` into a non-writable directory does not fail...

Should the test for writable be included in the initial test? Like this:
`echo $HOME && cd && test -w $HOME`

???

This is not as simple as one could expect... 😉

chrjorgensen (changes requested):

chrjorgensen (changes requested):

src/api/IBMi.ts
ThePrez: That makes sense. Does https://github.com/halcyon-tech/vscode-ibmi/pull/1460/commits/287f7c02255dc1327f1414d1689781dc53495518 look right?

chrjorgensen: Yep, this commit is correct.

src/api/IBMi.ts
```diff
@@ -213,14 +213,65 @@ export default class IBMi {
...
+          }
+          else if (reconnecting) {
+            vscode.window.showWarningMessage(`Your home directory (${actualHomeDir}) does not exist. Code for IBM i may not function correctly.`, { modal: false });
+          }
+          else if (await vscode.window.showWarningMessage(`Home directory does not exist`, {
+            modal: true,
+            detail: `Your home directory (${actualHomeDir}) does not exist, so Code for IBM i may not function correctly. Would you like to create this directory now?`,
+          }, `Yes`)) {
+            console.log(`creating home directory ${actualHomeDir}`);
+            let mkHomeCmd = `mkdir -p ${actualHomeDir} && chown ${connectionObject.username.toLowerCase()} ${actualHomeDir} && chmod 0755 ${actualHomeDir}`;
```
chrjorgensen, line 262: @ThePrez One issue here: Creation of home directory fails if a previous connection has been made and the home directory set in the configuration.

Maybe the operation should set the working directory to `.` like when you check the directory?

Currently it will run the `mkdir` after changing the directory to the home directory - which does not exists and make the operation fail...

chrjorgensen (changes requested):

src/api/IBMi.ts
ThePrez: > Maybe the operation should set the working directory to `.` like when you check the directory?
> 
> Currently it will run the `mkdir` after changing the directory to the home directory - which does not exists and make the operation fail...

I don't believe the operation would fail due to the missing home directory. There should be indications in stdout but the exit code should be 0. @chrjorgensen, are you observing that behavior? Based on #1477 I'm guessing so (due to different shell ? permissions?) 

Adding `.` is simple and harmless. Committed at https://github.com/halcyon-tech/vscode-ibmi/pull/1460/commits/7384684c6085a89b1058a7b980a38996de7bab98 . Thanks for thinking through these use cases!

ThePrez: > @ThePrez One issue here: Creation of home directory fails if a previous connection has been made and the home directory set in the configuration. This could happen if the home directory was deleted after previous connection.

This is an interesting use case. A previously-stored home directory might not exist, or the user profile's home directory may have changed.

I think it's debatable what the correct behavior is. Options:
1. Have VSCode respect the prior home directory setting
2. Save the new value and use it in the config accordingly. 

I think option 1 is the obvious choice, and is more diligently respected with commit at d452c5937fb49e7b574b8d3784f8dcc431906294 which cleans up the logic around storing the home directory into the configuration

chrjorgensen: > I don't believe the operation would fail due to the missing home directory.

The `sendCommand` function always sends a directory (either the specified or the config home directory) to the NodeSSH `execCommand`, which then does a `cd` into that directory before running the specified command. If the directory does not exist, the `cd` operation will fail (and the command never executed) and return error code `1` and stderr `______: No such file or directory`.

I got bitten by the same error, as you saw in #1477... 😄 

NodeSSH `execCommand` can have a directory of `null` and will then not change the directory before running the command - but C4i `sendCommand` does not use this feature and will always force a directory change. The only way to avoid this is specifying `.` as the directory for `sendCommand`.

Btw, the stderr `Could not chdir to home directory /home/________: No such file or directory` is from the SSH login - and will be returned for all following commands. @worksofliam Is this something we need to look at?
The initial working directory in such a situation will be the root `/`.

src/api/IBMi.ts
```diff
@@ -276,9 +276,7 @@ export default class IBMi {
 
         // Check to see if we need to store a new value for the home directory
         if (defaultHomeDir) {
-          if (this.config.homeDirectory === defaultHomeDir) {
-            progress.report({message: `Configured home directory (${defaultHomeDir}) appears usable.`});
-          } else {
+          if (this.config.homeDirectory !== defaultHomeDir) {
             this.config.homeDirectory = defaultHomeDir;
             progress.report({message: `Configured home directory reset to ${defaultHomeDir}.`});
```
chrjorgensen, line 281: I did not see this info message (must have been overwritten by the next progress message) - if you want the user to be notified about this change, maybe it should be written using the `showInformationMessage` function?

src/api/IBMi.ts
```diff
@@ -213,35 +213,79 @@ export default class IBMi {
...
-              command: `pwd`
-            });
-            if (pwdResult.stderr) {
-              //If it doesn't exist, reset it
-              this.config.homeDirectory = defaultHomeDir;
-              progress.report({
-                message: `Configured home directory reset to ${defaultHomeDir}.`
-              });
-            }
+            progress.report({message: `Configured home directory reset to ${defaultHomeDir}.`});
```
chrjorgensen, line 281: I did not see this info message (must have been overwritten by the next progress message) - if you want the user to be notified about this change, maybe it should be written using the `showInformationMessage` function?

chrjorgensen (commented):
@ThePrez Only one question left, then we can approve and merge... 😃

src/api/IBMi.ts
worksofliam: > Could not chdir to home directory /home/________: No such file or directory

I am not sure what we can do in this regard, though correct me if I’m wrong. This message will always appear if the user has a home directory configured on the profile that does not exist.

chrjorgensen: @worksofliam 
> I am not sure what we can do in this regard, though correct me if I’m wrong. This message will always appear if the user has a home directory configured on the profile that does not exist.

I was just surprised that the error message kept appearing in stderr for all commands executed - I would expect it to appear only once. Seems like the stderr is not emptied - an error in NodeSSH?

src/api/IBMi.ts
ThePrez: Sounds reasonable

ThePrez: Resolved at https://github.com/halcyon-tech/vscode-ibmi/pull/1460/commits/fe003eb5e075c947500d2dcabbb544309dde22e8

chrjorgensen (approved):
@ThePrez Perfect - approved and merged.
Welcome aboard, glad to have you as a contributor! 😃

PR #1477, opened 2023-08-09
Fix directory for shell check

chrjorgensen:
### Changes

This PR will set the directory when checking for shell output.
Previously if the home directory did not exists, the check would fail (discovered while reviewing PR #1460)

### Checklist

* [x] have tested my change


worksofliam (approved):
Working as expected! There is no impact to the user.

PR #1482, opened 2023-08-11
Wrap paths to resolve with double quote

worksofliam:
### Changes

This will fix a bug that will allow the vscode-rpgle extension to resolve streamfiles with spaces in them (though... what crazy person would do that?!)

Tests are continuing to pass.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -696,7 +696,7 @@ export default class IBMiContent {
   }
 
   async streamfileResolve(names: string[], directories: string[]): Promise<string | undefined> {
-    const command = `for f in ${directories.flatMap(dir => names.map(name => `"${path.posix.join(dir, name)}"`)).join(` `)}; do if [ -f $f ]; then echo $f; break; fi; done`;
+    const command = `for f in ${directories.flatMap(dir => names.map(name => `"${path.posix.join(dir, name)}"`)).join(` `)}; do if [ -f "$f" ]; then echo $f; break; fi; done`;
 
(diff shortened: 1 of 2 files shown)


src/api/IBMiContent.ts
```diff
@@ -696,7 +696,7 @@ export default class IBMiContent {
   }
 
   async streamfileResolve(names: string[], directories: string[]): Promise<string | undefined> {
-    const command = `for f in ${directories.flatMap(dir => names.map(name => path.posix.join(dir, name))).join(` `)}; do if [ -f $f ]; then echo $f; break; fi; done`;
+    const command = `for f in ${directories.flatMap(dir => names.map(name => `"${path.posix.join(dir, name)}"`)).join(` `)}; do if [ -f $f ]; then echo $f; break; fi; done`;
```
chrjorgensen, line 699: @worksofliam I added a test for the `streamfileResolve` function and found an error: `-bash: [: too many arguments`

You need to encapsulate $f in the [] test like this:

```suggestion
    const command = `for f in ${directories.flatMap(dir => names.map(name => `"${path.posix.join(dir, name)}"`)).join(` `)}; do if [ -f "$f" ]; then echo $f; break; fi; done`;
```

chrjorgensen (changes requested):

worksofliam:
New test is passing - thank you @chrjorgensen!

chrjorgensen (approved):
@worksofliam All tests pass now - will approve and merge! 👍

PR #1484, opened 2023-08-11
Add support for custom actions

SanjulaGanepola:
### Changes

With the removal of `iproj.json` actions in PR #1481, this PR adds the ability to run custom actions with custom deployment method. Also, updates `launchActionSetup` to accept a workspace folder.

### Checklist

* [x] have tested my change


worksofliam:
@halcyon-tech/core I need some additional eyes on this PR. I am happy with the code and have reviewed with Sanjula, but I'd like a second opinion. I'd also like some additional testing on this for different areas where Actions can be run.

* Running an action from active editors
   * Members
   * IFS
   * Local
* Running actions from views
   * Object Browser
   * IFS Browser

Thank you!

src/instantiate.ts
```diff
@@ -214,7 +214,7 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
       quickPick.onDidHide(() => quickPick.dispose());
       quickPick.show();
     }),
-    vscode.commands.registerCommand(`code-for-ibmi.runAction`, async (node) => {
+    vscode.commands.registerCommand(`code-for-ibmi.runAction`, async (node: any, group?: any, action?: Action, method?: DeploymentMethod) => {
```
sebjulliand, line 217: `group` seems to be unused here.
```suggestion
    vscode.commands.registerCommand(`code-for-ibmi.runAction`, async (node: any, action?: Action, method?: DeploymentMethod) => {
```

src/instantiate.ts
```diff
@@ -245,7 +245,7 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
           }
 
           if (canRun && [`member`, `streamfile`, `file`].includes(uri.scheme)) {
-            CompileTools.runAction(instance, uri);
+            CompileTools.runAction(instance, uri, action, method);
```
sebjulliand, line 248: We should `await` here if the caller of the command want to actually wait for it to end.
```suggestion
            await CompileTools.runAction(instance, uri, action, method);
```

sebjulliand (changes requested):
Code looks good. I just suggested minor changes.

I tested the following cases:
- Run action from current editor (IFS, Member and Local)
- Run action from IFS and Object browsers

Every action ran as expected.

A few other changes to consider though:
- Maybe the `extensions` field in `Action` should be optional now.
- Add a test for the `code-for-ibmi.runAction`, to be on the safe side (with an `Action` defined on the fly).

src/instantiate.ts
SanjulaGanepola: Although this parameter is unused, we still need it because when `code-for-ibmi.runAction` is invoked from the editor's context menu, VS Code passes in this parameter (it has to do with the editor tab group).

image

src/instantiate.ts
SanjulaGanepola: Good point. Update that: cf3a4a5

src/instantiate.ts
sebjulliand: Ha right, good catch 😅

SanjulaGanepola:
I made the following changes to the tests:
- Instead of testing `CompileTools.runAction` directly, I updated the tests to call the `code-for-ibmi.runAction` command as this gives us better coverage
- Added 2 additional tests for custom actions against members and streamfiles

image

SanjulaGanepola:
@sebjulliand As you suggested, I updated the `Actions` interface so that the `extensions` property is now optional -> [98a99f8](https://github.com/halcyon-tech/vscode-ibmi/pull/1484/commits/98a99f8392207e4db03ba0d5066f0ee260d1958c)

However, the actions.json schema was not changed.

worksofliam (approved):
I have tested the following scenarios:

* Running user defined actions for members
* Running user defined actions for local files
* Test cases are passing

I'd also like an additional approval from someone in @halcyon-tech/core before this gets merged. Thanks!

sebjulliand (approved):
Looks good and work as expected!

PR #1486, opened 2023-08-14
Initialize `code-for-ibmi:connected` context value when extension is activated

SanjulaGanepola:
### Changes

Whenever the user changes the first workspace folder or goes from a single-folder workspace to a multi-root workspace, the extension is restarted which is expected in VS Code as mentioned [here](https://github.com/Microsoft/vscode/issues/46048). However, right now this leaves Code4i's views as unusable (as shown below).

This seems to be because the value of the `code-for-ibmi:connected` context value is kept despite the extension restarting. Thus, this PR sets it to be false when we first activate the extension. Although this will require uses to have to reconnect, at least they can do this instead of having to restart the entire extension again.

image

### Checklist

* [x] have tested my change

### DIFF
--- a/src/instantiate.ts
+++ b/src/instantiate.ts
@@ -82,16 +82,19 @@ export async function disconnect(): Promise<boolean> {
 }
 
 export async function loadAllofExtension(context: vscode.ExtensionContext) {
+  // No connection when the extension is first activated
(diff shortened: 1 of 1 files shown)


sebjulliand (approved):
Simple and efficient change. Thanks!

PR #1496, opened 2023-08-22
Feature/task_view

worksofliam:
### Changes

This PR removes the Output Channel from Code for IBM i and makes all Actions use a traditional VS Code task.

### Checklist

* [x] have tested my change


chrjorgensen:
@worksofliam Looking good so far! 😍

worksofliam:
@chrjorgensen This is ready for a formal review. Cleaned up the last bits and it's good to go!

chrjorgensen:
@worksofliam I just tested the PR, and while I still like the speed and immediate show-up of the output from IBM i, I also miss the ability to see previous commands run. It may be because I'm an IBM i guy and used to (job)logs, but whatever the reason I like to have a log and be able to go back in history and see what happened - and this is not possible with the new task view...

Could it be an option for the user to activate this - giving the user a choice of the previous and new behavior?
Or could the IBM i command and output history be stored in a view - IBM i job log?

I may not be the only one having this issue... especially IBM i developers still working on source located on the server (source members or streamfiles) and used to compile from command line or RDi and having the output from previous commands available...

worksofliam:
@chrjorgensen Per your idea, I added `clearOutputEveryTime` back and now works the same way it did before.

chrjorgensen:
@worksofliam 

> @chrjorgensen Per your idea, I added `clearOutputEveryTime` back and now works the same way it did before.

Thanks - but probably not enough, because the task is removed at any press of a key if in focus, and there is no way of getting it back... 😢 

There are pros and cons for having it in Task instead of Output Channel - and Task may great for some while others may prefer Output Channel... I've not yet decided which I prefer... WDYT?

worksofliam:
@chrjorgensen Personally I kind of like it. It removes clutter when I am done with the task. I don't know if there is anyway to stop that functionality from happening... 😞

chrjorgensen (approved):
@worksofliam All good, no problems found. Approving... 👌

SJLennon:
@worksofliam 
I am somewhat confused with this change.  Where do I find the compile listing?  It used to be under IBM i Output.  Some times you need to look at the listing and/or the job log. I have `clearOutputEveryTime` checked (forever...)

worksofliam:
@SJLennon it's under the Terminals view now!

SJLennon:
@worksofliam 
That's the last place I've have thought of looking, since I virtually never use the Terminals view.  But maybe I'm an exception.

(There seem to be quite a few places in the documentation where IBM I Output turns up.)

PR #1521, opened 2023-09-06
Wait actual end of CompileTools.runAction

sebjulliand:
### Changes
Since `CompileTools.runAction` now uses task, it returns right after the task is created and started.
This PR makes it wait for the actual end of the action before returning and also lets the `code-for-ibmi.runAction` command returns the result of the action as a `boolean`.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/languages/clle/clRunner.ts
+++ b/src/languages/clle/clRunner.ts
@@ -23,7 +23,7 @@ export function initialise(context: ExtensionContext) {
               );
             }
             
-            return await CompileTools.runAction(instance, editor.document.uri, {
+            CompileTools.runAction(instance, editor.document.uri, {
               command: selectedCommand.content,
               environment: `ile`,
               name: `CL command`,
@@ -31,8 +31,6 @@ export function initialise(context: ExtensionContext) {
           }
         }
       }
-
-      return false;
     })
   )
 }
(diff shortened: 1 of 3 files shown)


src/languages/clle/clRunner.ts
```diff
@@ -25,14 +23,16 @@ export function initialise(context: ExtensionContext) {
               );
             }
             
-            CompileTools.runAction(instance, editor.document.uri, {
+            return await CompileTools.runAction(instance, editor.document.uri, {
```
worksofliam, line 26: Is there a reason why the `return` statements were added here? Generally don't want this API to used like that, and actually will eventually be moved into the CL extension.

worksofliam (dismissed):
Works as expected, just check out my comment about the CL runner above.

src/languages/clle/clRunner.ts
sebjulliand: Mainly for consistency and in the event it gets called by another extension.
But if it's not intended to be used externally or even this way, I'll be happy to roll it back 😉

worksofliam: I think I'd prefer it if you did! Don't want to let people get comfy with it and then have me remove it down the line!~

sebjulliand:
@worksofliam there you go!

worksofliam (approved):
A nice functional change. Thanks @sebjulliand!

PR #1539, opened 2023-09-14
Automatically bump versions on release

sebjulliand:
### Changes
This PR adds two actions to the release process:
- It automatically change set the version number in the extension's `package.json` and in the types' `package.json`, based on the release name.
- It commits and push this change on the master branch

This would allow contributors to create a release without the need to create a PR for bumping the versions.
It also ensures that both the extension and the types version are always aligned when publishing.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -5,7 +5,13 @@ on:
 jobs:
   release:
     name: Release and publish
+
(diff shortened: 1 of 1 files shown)


worksofliam (approved):
This is awesome. Let's merge! I am excited to give this one a try.

PR #1540, opened 2023-09-14
Missing call to translation function

sebjulliand:
### Changes
Added a missing call to `t()` in the Object Browser to fix this translation:
image

### Checklist
* [x] have tested my change


chrjorgensen (approved):
@sebjulliand Good find! What an effect three missing characters can have... 😆

Issue #1542, opened 2023-09-15
Reconnecting with private key fails with ENAMETOOLONG


edmundreinhardt:
After a timeout disconnect, Code for IBM i prompts correctly to reconnect.
![image](https://github.com/halcyon-tech/vscode-ibmi/assets/11184579/3d5194d1-1e55-452d-ba57-6e53f957f709)

But when attempting to do so, the connection attempt fails with
![image](https://github.com/halcyon-tech/vscode-ibmi/assets/11184579/29a6e57c-f7c3-4d06-bad9-d5dc17edb1bc)
...
<img width="528" alt="image" src="https://github.com/halcyon-tech/vscode-ibmi/assets/11184579/23d8ef3e-a0af-4bfb-b711-6b9ab1bcddbe">

Pressing Yes to reconnect simply causes the same  error to reappear 

However when I Cancel and reconnect it succeeds
![image](https://github.com/halcyon-tech/vscode-ibmi/assets/11184579/d4175667-ed6c-4ce5-a28e-6a0ad97dfc59)

Note the Login Settings specify the Private Key
<img width="853" alt="image" src="https://github.com/halcyon-tech/vscode-ibmi/assets/11184579/a7078476-bd7e-4fb7-abc2-79c9a67ef315">

<hr />

⚠️ **REMOVE THIS LINE AND ANY SENSITIVE INFORMATION BELOW!** ⚠️

|Context|Version|
|-|-|
|Code for IBM i version|2.2.0|
|Visual Studio Code version|1.82.0|
|Operating System|darwin_arm64|

<details>
<summary>Active extensions</summary>

```
Code for IBM i Walkthroughs (vscode-ibmi-walkthroughs): 0.3.1
Db2 for IBM i (vscode-db2i): 0.4.1
ESLint (vscode-eslint): 2.4.2
Emmet (emmet): 1.0.0
Extension Authoring (extension-editing): 1.0.0
Git (git): 1.0.0
Git Base (git-base): 1.0.0
GitHub (github): 0.0.1
GitHub Authentication (github-authentication): 0.0.2
GitLens — Git supercharged (gitlens): 14.3.0
IBM i Project Explorer (vscode-ibmi-projectexplorer): 1.2.2
JSON Language Features (json-language-features): 1.0.0
Live Share (vsliveshare): 1.0.5883
Merge Conflict (merge-conflict): 1.0.0
Microsoft Account (microsoft-authentication): 0.0.1
[...]

PR #1544, opened 2023-09-18
Renamed connection's privateKey to privateKeyPath

sebjulliand:
### Changes
- Use the correct field used by node-ssh Config object
- Added a method called on activation to fix existing configurations (can be removed in the future)
- Fixes https://github.com/halcyon-tech/vscode-ibmi/issues/1542

The private key file content was loaded into the `privateKey` field, which is not wrong, but it caused an issue when the `ConnectionData` object was reused while reconnecting a dropped connection.

This PR renames `ConnectionData` `privateKey` to `privateKeyPath` which is an actual field used by node-ssh when connecting. This way we use the correct field for the job and leave it up to node-ssh to read the file's content, too.

A small migration method is called when the extension activates to update the existing configurations.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -3,7 +3,6 @@ import * as node_ssh from "node-ssh";
 import * as vscode from "vscode";
 import { ConnectionConfiguration } from "./Configuration";
 
-import { readFileSync } from "fs";
 import path from 'path';
 import { instance } from "../instantiate";
 import { CommandData, CommandResult, ConnectionData, IBMiMember, RemoteCommand, StandardIO } from "../typings";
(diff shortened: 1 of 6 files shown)


worksofliam:
@edmundreinhardt Are you able to test this PR?

edmundreinhardt:
I have tested the PR and it works

worksofliam:
Thank you @edmundreinhardt!

worksofliam (approved):
Code looks good to me. Thanks @sebjulliand!

I bet in the future we rename `fixPrivateKeys` to something more generic and use it to fix more config changes in the future. 

Great work!

PR #1546, opened 2023-09-18
Remove check for excessive authority

forstie:
removed the dunning msg

### Changes

Description of change here.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -341,18 +341,6 @@ export default class IBMi {
           }
         }
 
-        if (!reconnecting) {
-          progress.report({
-            message: `Checking for excessive authority.`
-          });
-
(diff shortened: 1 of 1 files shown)


forstie:
@worksofliam

worksofliam:
@chrjorgensen I have no issues with this personally. What do you think? :D

chrjorgensen (approved):
@forstie @worksofliam I agree with this...

When the check was first introduced, I was hesitating since I would be hit by this message.
Then I grew to like it, because I was reminded about having more authority than needed.
But now I agree about removing the check because the message is not clear and often just gets scrolled out of view by other messages at connection time.

Perhaps the check could be re-introduced later with a more thorough check for not needed authorities - as a separate function, so the user can check at will...

Approved! 😉

PR #1548, opened 2023-09-19
Fixed ProviderResult import in typings

sebjulliand:
### Changes
The `ProviderResult` type imported in the typings was imported from the Git API types instead of `vscode`'s.

### Checklist
* [x] have tested my change


sebjulliand:
It blew up to my face this afternoon 😛

Issue #1565, opened 2023-10-03
Copy library fails trying to use CPYOBJ command


edmundreinhardt:
It appears that the action for copying a library is using CPYOBJ which is failing as *LIB is not vaid for parameter OBJTYPE
The message CPD0084 is shown.
We probably need a unique action for copying a library when showing libraries in the QSYS Object Browser

![image](https://github.com/halcyon-tech/vscode-ibmi/assets/11184579/2dbb72b9-0833-42a9-b51f-aa3f3cae3f26)

<hr />

⚠️ **REMOVE THIS LINE AND ANY SENSITIVE INFORMATION BELOW!** ⚠️

|Context|Version|
|-|-|
|Code for IBM i version|2.3.1|
|Visual Studio Code version|1.82.2|
|Operating System|darwin_arm64|

<details>
<summary>Active extensions</summary>

```
Code for IBM i Walkthroughs (vscode-ibmi-walkthroughs): 0.3.2
Db2 for IBM i (vscode-db2i): 0.5.0
ESLint (vscode-eslint): 2.4.2
Emmet (emmet): 1.0.0
Extension Authoring (extension-editing): 1.0.0
Git (git): 1.0.0
Git Base (git-base): 1.0.0
GitHub (github): 0.0.1
GitHub Authentication (github-authentication): 0.0.2
GitLens — Git supercharged (gitlens): 14.3.0
IBM i Project Explorer (vscode-ibmi-projectexplorer): 1.2.2
JSON Language Features (json-language-features): 1.0.0
Live Share (vsliveshare): 1.0.5883
Merge Conflict (merge-conflict): 1.0.0
Microsoft Account (microsoft-authentication): 0.0.1
NPM support for VS Code (npm): 1.0.1
Node Debug Auto-attach (debug-auto-launch): 1.0.0
RPGLE (vscode-rpgle): 0.21.4
TypeScript and JavaScript Language Features (typescript-language-features): 1.0.0
undefined (vscode-sourceorbit): 1.1.0
```
</details>

<hr />

<details>
<summary>Remote system</summary>

|Setting|Value|
|-|-|
|IBM i OS|n/a|
|Tech Refresh|n/a|
|CCSID|65535|
|SQL|Disabled
|Source dates|Disabled

### Enabled features

[...]

PR #1566, opened 2023-10-03
Fixed copyObject crashing when copying library

sebjulliand:
### Changes
Fixes https://github.com/halcyon-tech/vscode-ibmi/issues/1565

The object type test made in the `code-for-ibmi.copyObject` action was not changed after the Object Browser was converted to TypeScript. This PR fixes the test to compare the selected object's type with `*lib`.

### Checklist
* [x] have tested my change


worksofliam (approved):
Good find!

PR #1585, opened 2023-10-12
Remove URI query part when prompting for member comparison

sebjulliand:
### Changes
Fixes https://github.com/codefori/vscode-ibmi/issues/1584
This PR removes the query part of the URI shown when prompting for a source member to compare to.

### Checklist
* [x] have tested my change


worksofliam (approved):
No kidding, very small! Thanks @sebjulliand !!

PR #1596, opened 2023-10-18
Use OS dependant function to get relative path

sebjulliand:
### Changes
Fixes https://github.com/codefori/vscode-ibmi/issues/1586

On Windows, there is sometimes a case difference in the path's drive letter. For example:
image

In my case, I could reproduce this by having two workspace folders in my workspace folders in my workspace and running a local action after opening a file from the Search results (like described in https://github.com/codefori/vscode-ibmi/issues/1586).

This PR simply checks the target platform and run the `path.win32.relative` function to get the relative path for local actions if running on Windows.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/CompileTools.ts
+++ b/src/api/CompileTools.ts
@@ -224,8 +224,15 @@ export namespace CompileTools {
 
                   if (fromWorkspace) {
                     baseDir = fromWorkspace.uri.path;
+                    
+                    if(process.platform === 'win32'){
+                      relativePath = path.win32.relative(baseDir, uri.path);
+                    }
+                    else{
+                      relativePath = path.posix.relative(baseDir, uri.path); 
+                    }
+                    relativePath = relativePath.split(path.sep).join(path.posix.sep);
 
-                    relativePath = path.posix.relative(baseDir, uri.path).split(path.sep).join(path.posix.sep);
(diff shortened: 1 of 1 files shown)


worksofliam:
@sebjulliand I can pull my Windows machine up tomorrow and give this a try.

worksofliam:
@sebjulliand Works on my Windows machine, but just a though:

Before, we were using `path.posix.join` to create a posix style path, when actually we needed a Windows style path - thus your PR.

But I don't think you even need to check if it's Windows to use `path.win32.join` - can't you just use `path.join` and let the runtime figure out the correct path style?

worksofliam (dismissed):
Works but see my comment!

sebjulliand:
> Works but see my comment!

You're absolutely right! It is simpler to let `path` functions handle the OS specifics. See my changes, it simplified it all (while still working 😅).

worksofliam (approved):

worksofliam (approved):
I haven't been able to test on my Windows machine (I have broken it.. long story), but it does continue to work on my macbook.

I trust that if you feel it works on Windows, and I know it works on Mac, I think a merge would be ok.

sebjulliand:
A merge it is then!

PR #1612, opened 2023-10-24
Fixed sort action in Object Browser

sebjulliand:
### Changes
Sorting in the Object Browser was not working properly because:
- Nodes' `refresh` method was not calling the correct command
- The `sortBy` method at member level was not bound correctly

### Checklist
* [x] have tested my change


chrjorgensen (approved):
@sebjulliand Good catch! 🎣

Issue #1475, opened 2023-08-05
$HOME environment variable being set to the workspace deploy directory


richardm90:
This may have always been the case and I've only just noticed whilst recently trying out the local development functionality but when I deploy my local workspace and build my application (using gmake) I've noticed that my home directory is being set to the workspace deploy directory.

This means I can't reference my true home directory (using the $HOME environment variable) in a Makefile or a bash script that I want to execute through VS Code.

This doesn't seem right to me.

<hr />

|Context|Version|
|-|-|
|Code for IBM i version|2.0.2|
|Visual Studio Code version|1.81.0|
|Operating System|linux_x64|

<details>
<summary>Active extensions</summary>

```
COBOL (cobol): 9.7.23
Code for IBM i Walkthroughs (vscode-ibmi-walkthroughs): 0.3.1
Db2 for IBM i (vscode-db2i): 0.3.3
Emmet (emmet): 1.0.0
Error Lens (errorlens): 3.12.0
Extension Authoring (extension-editing): 1.0.0
Git (git): 1.0.0
Git Base (git-base): 1.0.0
GitHub (github): 0.0.1
GitHub Authentication (github-authentication): 0.0.2
GitLens — Git supercharged (gitlens): 14.2.0
IBM i Notebooks (vscode-ibmi-notebooks): 0.0.6
JSON Language Features (json-language-features): 1.0.0
Makefile Tools (makefile-tools): 0.7.0
Merge Conflict (merge-conflict): 1.0.0
Microsoft Account (microsoft-authentication): 0.0.1
NPM support for VS Code (npm): 1.0.1
Node Debug Auto-attach (debug-auto-launch): 1.0.0
TODO Highlight (vscode-todo-highlight): 1.0.5
```
</details>

<hr />

**_Not connected_** 🔌

worksofliam:
Yep, we do the set working directory as the deploy directory - this is because most build tools expect that.

It looks like [`getDefaultVariables`](https://github.com/halcyon-tech/vscode-ibmi/blob/master/src/api/CompileTools.ts#L185) is setting `&HOME` to the working directory (which we refer to internally as `homeDirectory` even though it is not), which is then being set to `&HOME` [here](https://github.com/halcyon-tech/vscode-ibmi/blob/master/src/api/CompileTools.ts#L608).

Easy fix is to take it out of `getDefaultVariables`.

richardm90:
Hi @worksofliam , let me give that a try.

PR #1621, opened 2023-10-31
1475 fix home envvar

richardm90:
### Changes

This PR relates to issue #1475.

I have modified `CompileTools.ts` so that it no longer sets the `HOME` variable, which goes on to set the PASE `HOME` environment variable. I have also added a new variable called `WORKDIR`, as a replacement for `HOME`.

In doing so the `HOME` variable is no longer available for use in actions.

**IMPORTANT** Is there a way to advise users that their actions will need updating?

I have also added another new variable called `FILEDIR` that gives you the remote directory that the file exists. This is extremely useful when you have multiple `Makefile`s in your project. You can have an action command such as `cd &FILEDIR && /QOpenSys/pkgs/bin/gmake`.

I have also updated the `varinfo.ts` and appropriate locale files with the action variable description changes. I used Google Translate on the French and Danish versions so if someone could check these that would be useful.

### Checklist

* [x] have tested my change


src/api/CompileTools.ts
```diff
@@ -61,7 +61,7 @@ export namespace CompileTools {
       variables.set(`{usrprf}`, connection.currentUser);
       variables.set(`&HOST`, connection.currentHost);
       variables.set(`{host}`, connection.currentHost);
-      variables.set(`&HOME`, config.homeDirectory);
```
worksofliam, line 64: I don't think we should remove `&HOME` for now since a lot of people are using this today and we don't have a way to fix where it's used automatically.

worksofliam: I think adding the `&HOME` variable back in, but removing it from the docs is a good idea. I think moving to `&WORKDIR` is the right idea for sure.

worksofliam: @richardm90 I am going to write a review.

Let's keep `&FILEDIR` for sure, but let's also keep both `&WORKDIR` plus `&HOME` for compatibility reasons.

worksofliam (changes requested):
Let's keep `&FILEDIR` for sure, but let's also keep both `&WORKDIR` plus `&HOME` for compatibility reasons.

When you have added `&HOME` back in (you don't need to add it back to the variable list UI) then I will review again.

richardm90:
Hey @worksofliam was my update OK?

richardm90:
@worksofliam , took me a little while as I was getting a problem when testing which must have been sorted in a recent commit as pulling down the latest has resolved my problem. I've updated the PR as suggested.

worksofliam:
@richardm90 I am sorry - every time I come around to look at this there are new conflicts. Any chance you can fix those and I can get another review in?

richardm90:
@richardm90 , no problem, I'll get to it this week along with the comments you left on the other pr

richardm90 (commented):
@worksofliam I've applied the latest changes

worksofliam (approved):
Looks good to me!

If you want to update this page for local development to include these changes, you're welcome: https://github.com/codefori/docs/blob/main/docs/pages/developing/local/actions.md

PR #1632, opened 2023-11-06
Remove duplicates from the library list

worksofliam:
## Changes

When `.env` is used, or when the `runCommand` API is used, it's possible for the user or code to pass in duplicate libraries, which the IBM i doesn't like too much. This is a simple change to remove duplicates from the end of the library list.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/CompileTools.ts
+++ b/src/api/CompileTools.ts
@@ -516,7 +516,7 @@ export namespace CompileTools {
         if (curlib) ileSetup.currentLibrary = curlib;
       }
 
-      // Remove any duplicates fromr the library list
+      // Remove any duplicates from the library list
       ileSetup.libraryList = ileSetup.libraryList.filter(Tools.distinct); 
 
       let commandString = replaceValues(


src/api/CompileTools.ts
```diff
@@ -516,6 +516,12 @@ export namespace CompileTools {
         if (curlib) ileSetup.currentLibrary = curlib;
       }
 
+      // Remove any duplicates fromr the library list
+      ileSetup.libraryList = ileSetup.libraryList
+        .filter((item, pos, self) => {
+          return self.indexOf(item) == pos;
+        })
```
sebjulliand, line 523: For the sake of reusing the Tools at hand :)
```suggestion
      ileSetup.libraryList = ileSetup.libraryList.filter(Tools.distinct);
```

sebjulliand (changes requested):
Just me nitpicking ;)

src/api/CompileTools.ts
worksofliam: Oh, I did not even know we had that. Fab!!

sebjulliand (approved):
Nice! Let's merge!

PR #1643, opened 2023-11-09
Branch library variable

worksofliam:
### Changes

* Sets branch new `&BRANCHLIB` variable for running an Action out of a branch. ([Docs need additional logic](https://github.com/codefori/docs/pull/19) for this when merged)
* Adds logic in to recursively resolve variables in variables. For example: `CURLIB=&BRANCHLIB`

### Checklist

* [x] have tested my change

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,17 +1,18 @@
 {
 	"name": "code-for-ibmi",
-	"version": "2.3.1",
+	"version": "2.4.0",
 	"lockfileVersion": 2,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "code-for-ibmi",
-			"version": "2.3.1",
+			"version": "2.4.0",
 			"license": "MIT",
 			"dependencies": {
 				"@bendera/vscode-webview-elements": "^0.12.0",
 				"@ibm/ibmi-eventf-parser": "^1.0.2",
 				"@types/tmp": "^0.2.3",
+				"crc-32": "https://cdn.sheetjs.com/crc-32-latest/crc-32-latest.tgz",
 				"csv": "^6.2.1",
 				"ignore": "^5.1.9",
 				"node-ssh": "^13.1.0",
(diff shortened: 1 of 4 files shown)


sebjulliand (changes requested):
I'm fine with the `replaceValues` enhancement, that's great!
I'm not so sure about the proposed way to get the library name out of the branche name. It seems risky, especially if a number find it's way as the first character 😅

Allow me to expose another approach I have in mind: I'd advise we use a hash algorithm to compute the library name out of the branch name. I looked for an algorithm that would produce <= 10 characters long hash and went with CRC32 and its [JS implementation](https://github.com/SheetJS/js-crc32). It would produce a 8 characters long hex string we concatenate to a 2 characters constant (e.g `VS`).

I guess this would give something like this:
```js
import { str } from "crc-32/crc32c";

export function getBranchLibraryName(currentBranch: string) {
   return `VS${(str(currentBranch, 0)>>>0).toString(16).toUpperCase()}`;
}
```
It may not be ideal since there will always be a risk of collision (even though it should be low in our context), but I feel it may be a safer approach since we don't assume anything about the branch name.

And the library could be created from VSCode when needed (i.e. if the action about to be run uses `$BRANCHLIB`, then create it beforehand if it doesn't exist), so we could put the branch name in the library's description (so people know which library is linked to which branch).

What do you guys think?

chrjorgensen:
Just a note from an experienced (or is it just old?) IBM i system administrator:
**What about the cleanup of these temporary libraries?**

I'm all for working in temp libraries, but there needs to be some way to clean them up (i.e. remove them from the system), or the system will be cluttered with obsolete libraries, which no one will be able to identify as obsolete...

😉

worksofliam:
Long term, I am thinking of additional action props (for local actions) that would be triggered when a branch checkout happens or when a branch is deleted (although no one ever deletes local branches).

@chrjorgensen I believe those issues are real, but likely a separate PR will solve those D:

chrjorgensen:
> I believe those issues are real, but likely a separate PR will solve those D:

@worksofliam Accepted, but I will keep an 👁️ for it! 😜

worksofliam:
@sebjulliand You're right, that crc32 implementation for a deterministic name is much better than my hack job, esp. considering we can put the branch name into the object text.

I am also the same logic for Source Orbit, so I put some tests for this over there: https://github.com/IBM/sourceorbit/pull/12/files#diff-72619d6ae4762f96f7a669264ea989c507a8373a79dfc09e5e9fb52b327616b2

The only downside about the crc32 package you shared is that it's not available on npm and only sheetjs. Hopefully that won't be an issue in the future!

I will work on getting that added to this PR soon.

sebjulliand:
> @sebjulliand crc32 is perfect for this. I've been using it a bunch over the last few days and it is really nice. Also it's really nice seeing branch libraries with `WRKLIB VS*` - thank! I will work on adding it to this PR soon.

Awesome! Glad I could help 😁

worksofliam:
@sebjulliand I have added crc32 to generate the library name as you suggested. Thanks again! I have requested your review again.

PR #1651, opened 2023-11-15
Look for member on SYSBAS if config.sourceASP is set

sebjulliand:
### Changes
This fixes an issue that can happen when `downloadMemberContent` is called to open a member located on `*SYSBAS` and an iASP is specified in the Connection Settings (i.e. `config.sourceASP`).

When `config.sourceASP` is set, `downloadMemberContent` will always assume the member content to fetch is located on that IASP. If the member is actually on `*SYSBAS`, it can never be read.

This PR makes `downloadMemberContent` tries twice to fetch a member if an IASP is set in the settings:
- First it assumes the member is on the IASP (as it used to)
- If not found, it will try again without the IASP in the path
- If still not found, then it crashes

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -121,9 +121,9 @@ export default class IBMiContent {
     const client = this.ibmi.client;
 
     let retry = false;
+    let path = Tools.qualifyPath(library, sourceFile, member, asp);
+    const tempRmt = this.getTempRemote(path);
     while (true) {
-      const path = Tools.qualifyPath(library, sourceFile, member, asp);
-      const tempRmt = this.getTempRemote(path);
       try {
         await this.ibmi.remoteCommand(
           `CPYTOSTMF FROMMBR('${path}') TOSTMF('${tempRmt}') STMFOPT(*REPLACE) STMFCCSID(1208) DBFCCSID(${this.config.sourceFileCCSID})`, `.`
@@ -147,7 +147,7 @@ export default class IBMiContent {
             case "CPFA0A9":
               //The member may be located on SYSBAS
               if (asp) {
-                asp = undefined;
+                path = Tools.qualifyPath(library, sourceFile, member);
                 retry = true;
               }
               break;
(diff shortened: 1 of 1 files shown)


worksofliam:
Nice. Having a look at the code, it looks like it's falling back to *SYSBAS if it's not in the ASP the user defines first, so it looks in this order:

1. If the user has an ASP defined, look there. Return it if found.
2. Otherwise, look in *SYSBAS (`undefined` in the `asp` variable).

worksofliam (changes requested):
Looks like it worked for reading files.. but not saving!

I don't have any ASPs to test with, but it did allow me to test with an invalid name to test the fall back: 

image

In the output:

```
.: system "CPYTOSTMF FROMMBR('/boop/QSYS.lib/LIAMA.lib/QRPGLESRC.file/JP.mbr') TOSTMF('/tmp/vscodetemp-O_3z3bFCsE') STMFOPT(*REPLACE) STMFCCSID(1208) DBFCCSID(*FILE)"
{
    "code": 255,
    "signal": null,
    "stdout": "",
    "stderr": "CPFA0A9: Object not found.  Object is /boop/QSYS.lib/LIAMA.lib/QRPGLESRC.file/JP.mbr.\nCPFA097: Object not copied.  Object is /boop/QSYS.lib/LIAMA.lib/QRPGLESRC.file/JP.mbr."
}

.: system "CPYTOSTMF FROMMBR('/QSYS.lib/LIAMA.lib/QRPGLESRC.file/JP.mbr') TOSTMF('/tmp/vscodetemp-O_yubtaBfe') STMFOPT(*REPLACE) STMFCCSID(1208) DBFCCSID(*FILE)"
{
    "code": 0,
    "signal": null,
    "stdout": "CPCA082: Object copied.",
    "stderr": ""
}
```

When trying to save this member:

image

sebjulliand:
Ugh, yeah, I forgot that there is a download AND an upload memberContent 😅
I fixed the upload too. Got for it!

worksofliam (approved):
Yep, now it works! This is a totally nice UX improvement. 

@sebjulliand Merge when you are ready!!

Issue #1381, opened 2023-06-22
Ability to set the default deploy option


worksofliam:
Chosen from the connection settings

markhull1990:
Hello!

I'm looking to make this my first contribution (I'm new to JS and TS). I've added a "Default Deployment Location" in the connection settings on the features tab, but I'm struggling to come up with the logic of how it should behave in Deployment.ts.  Does anyone have any thoughts on how best to implement this?

![image](https://github.com/halcyon-tech/vscode-ibmi/assets/137716546/6b7c5b36-6d6b-48b8-b92d-f1d891cf1bdf)

worksofliam:
Hey mark

We are looking for 'default deploy **option**', not location 😄 

Ideally would be a dropdown with 6 options, the first being 'no default' and the remaining items being the options available when selecting which deploy method to use.

Also, I think this belongs in the Source Code tab.

Hope this helps!

PR #1652, opened 2023-11-15
Add Ability to set Default Deployment Method Issue #1381

beckhamrryyaann:
### Changes

This PR addresses issue #1381 .

The user is now able to set the default deploy method from the "Source" tab in the Connection Settings web view.  All methods are available to choose.  Once set, when user runs action that involves deploying, this PR will validate the default method is a usable option based on users current instance.  If all checks pass then the project will deploy without user needing to select the deploy method from the quick pick.  If there is an issue (default method is not usable) then the code will gracefully show the quick pick menu.

### Checklist

* [X] have tested my change


src/webviews/settings/index.ts
```diff
@@ -89,8 +89,46 @@ export class SettingsUI {
               text: `Track changes using the diff mechanism. Before the document is saved, it is compared to the original state to determine the changed lines. (Test enhancement)`,
             },
           ], `Determine which method should be used to track changes while editing source members.`)
+          .addSelect(`defaultDeploymentMethod`, `Default Deployment Method`, [
```
worksofliam, line 92: This select makes sense, but when deploying we usually only show the user the options that are available to them. I am guessing we should do the same here.

https://github.com/codefori/vscode-ibmi/blob/2ae070eaff3d978e34063551fc01e26d382f7621/src/api/local/deployTools.ts#L58-L75

worksofliam (changes requested):
Nice change! I left a comment about only showing options that are available. This might be tricky because we won't know what options are available when the user is not connected - this is because the connection settings can be edited offline.

What do you think @sebjulliand?

src/webviews/settings/index.ts
beckhamrryyaann: I do not know what is available until the user connects.  It's like the which comes first ...egg or chicken parable!  :-)  What I decided to do was when the actual deploy action takes place I verify that the default method is a valid choice.  If not, it gracefully falls back as if the user has no default selected and shows the quick pick.  I could add a informational message with something like "default deploy method is not usable in your environment.  You will need to update your default deployment to a valid method."
https://github.com/beckhamrryyaann/vscode-ibmi/blob/e5a0a1bba5ae593fbcb6c700dd039e5e20482d1c/src/api/local/deployTools.ts#L75

But I will do whatever you and your team suggests...just let me know how you and @sebjulliand want me to proceed when you get chance.

worksofliam: I actually think what you have now makes total sense. Show them all in the settings. but then complain when they deploy if the method is not available.

src/api/Configuration.ts
```diff
@@ -56,6 +56,7 @@ export namespace ConnectionConfiguration {
     debugEnableDebugTracing: boolean;
     readOnlyMode: boolean;
     quickConnect: boolean;
+    defaultDeploymentMethod: string;
```
worksofliam, line 59: Instead of type `string` can you use `DeploymentMethod | ''`?

src/api/local/deployTools.ts
```diff
@@ -72,9 +72,20 @@ export namespace DeployTools {
...
+          // the default deploy method must be an available option for this connection.
+          // if not available show list of quick pick options else skip quick pick and use default
+          let defaultDeploymentMethod = instance.getConfig()?.defaultDeploymentMethod as DeploymentMethod
+
+          if (methods.find((element) => element.method === defaultDeploymentMethod &&
+              defaultDeploymentMethod as String !== `` )) {
+            method = defaultDeploymentMethod
+          } else {
+
+            method = (await vscode.window.showQuickPick(methods,
```
worksofliam, line 85: Perhaps when `defaultDeploymentMethod` is not blank, we need to show a warning here if the `defaultDeploymentMethod` is not available?

worksofliam (changes requested):
Add some nit picks, but other than that it looks pretty good.

@sebjulliand you might want to check this one out too.

src/api/Configuration.ts
beckhamrryyaann: done.

src/api/local/deployTools.ts
beckhamrryyaann: warning is added.

sebjulliand:
@worksofliam I love it.
@beckhamrryyaann great job, I'll just have a few remarks.

src/api/local/deployTools.ts
```diff
@@ -57,7 +57,7 @@ export namespace DeployTools {
         if (!method) {
           const methods = [];
           if (Deployment.getConnection().remoteFeatures.md5sum) {
-            methods.push({ method: "compare" as DeploymentMethod, label: `Compare`, description: `Synchronizes using MD5 hash comparison` });
+            //methods.push({ method: "compare" as DeploymentMethod, label: `Compare`, description: `Synchronizes using MD5 hash comparison` });
```
sebjulliand, line 60: I guess this wasn't meant to be commited?
```suggestion
            methods.push({ method: "compare" as DeploymentMethod, label: `Compare`, description: `Synchronizes using MD5 hash comparison` });
```

src/api/local/deployTools.ts
```diff
@@ -72,9 +72,24 @@ export namespace DeployTools {
 
           methods.push({ method: "all" as DeploymentMethod, label: `All`, description: `Every file in the local workspace` });
 
-          method = (await vscode.window.showQuickPick(methods,
-            { placeHolder: `Select deployment method to ${remotePath}` }
-          ))?.method;
+          let defaultDeploymentMethod = instance.getConfig()?.defaultDeploymentMethod as DeploymentMethod
```
sebjulliand, line 75: You can (and should) use `const` here as `defaultDeploymentMethod` isn't changed after being set.
```suggestion
          const defaultDeploymentMethod = instance.getConfig()?.defaultDeploymentMethod as DeploymentMethod
```

src/api/local/deployTools.ts
```diff
@@ -72,9 +72,24 @@ export namespace DeployTools {
...
+
+          if (methods.find((element) => element.method === defaultDeploymentMethod)) { // default deploy method is usable
+            method = defaultDeploymentMethod
+          
+          } else if (defaultDeploymentMethod as string !== '' && defaultDeploymentMethod !== undefined) { // default deploy method is set but not usable.
```
sebjulliand, line 80: You can you reduce your test to this since you're testing if `defaultDeploymentMethod` is not `falsy`.
```suggestion
          } else if (defaultDeploymentMethod) { // default deploy method is set but not usable.
```

sebjulliand (approved):
It's a nice change!
Apart from a few style suggestions, I'd love to see the `compare` method back 🤣

beckhamrryyaann:
@sebjulliand oh my!! I sure am glad you caught that.  Some people might miss that!  I have put your fixes in place.  Thanks for your review.

sebjulliand (dismissed):
Looks good to me! That's very nice, thanks @beckhamrryyaann !
@worksofliam I'll let you have a last look and merge.

src/api/local/deployTools.ts
```diff
@@ -72,9 +72,24 @@ export namespace DeployTools {
...
-          ))?.method;
+          const defaultDeploymentMethod = instance.getConfig()?.defaultDeploymentMethod as DeploymentMethod
+
+          if (methods.find((element) => element.method === defaultDeploymentMethod)) { // default deploy method is usable
+            method = defaultDeploymentMethod
+          
+          } else if (defaultDeploymentMethod) { // default deploy method is set but not usable.
+            vscode.window.showWarningMessage('Default deployment method is set but not usable in your environment.')
+
+            method = (await vscode.window.showQuickPick(methods,
```
worksofliam, line 83: I don't really like that this assignment is duplicated in two else blocks. Perhaps only one else block is needed, with an additional if inside of it to check the valididity of `defaultDeploymentMethod`?

worksofliam (changes requested):

worksofliam (changes requested):
Looking really good! Left one comment about some duplicate code which I think can be cleaner. 

Other than that, nice work!

src/api/local/deployTools.ts
beckhamrryyaann: much cleaner now.

worksofliam:
Code looks good! I will get a review in today!

worksofliam (approved):

worksofliam (approved):
Works good to me! I am never taking it off of compare again haha. Thanks for a great PR!

PR #1653, opened 2023-11-15
Fix variable assignment order in CompileTools.ts

worksofliam:
### Changes

Because `BRANCH` was before `BRANCHLIB`, I was getting the incorrect value in `BRANCHLIB` as `&BRANCHLIB` was evaluating to `some/random/branchLIB`

### Checklist

* [x] have tested my change


sebjulliand (approved):
It took me a few minutes to understand why the order was important 😅

PR #1654, opened 2023-11-17
Additional privateKey prop to allow private key strings

worksofliam:
### Changes

This will allow a `privateKey` property on `ConnectionData`, which allows the content of the private key to be passed in, instead of solely just the path to the key.

No UI changes needed as this is a requirement for the connection API.

### Checklist

* [ ] have tested my change


william-xiang:
@worksofliam Tested in Merlin, it works now with the private key content in the connection data.

PR #1655, opened 2023-11-17
Add code to escape member names when doing resolve

worksofliam:
### Changes

Fixes issue where member resolves with dollar signs in the name (which sadly is very common) was not working because `$` is a special character in the shell.

Unit tests for `memberResolve` are continung to pass.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -652,9 +652,13 @@ export default class IBMiContent {
 
   async memberResolve(member: string, files: QsysPath[]): Promise<IBMiMember | undefined> {
     // Escape names for shell
-    const fixed = files.map(f => ({name: f.name.replace(/([$\\])/g,'\\$1'), library: f.library.replace(/([$\\])/g,'\\$1')}));
-    const command = `for f in ${fixed.map(file => `/QSYS.LIB/${file.library.toUpperCase()}.LIB/${file.name.toUpperCase()}.FILE/${member.toUpperCase()}.MBR`).join(` `)}; do if [ -f $f ]; then echo $f; break; fi; done`;
+    const pathList = files
+      .map(file => `/QSYS.LIB/${file.library}.LIB/${file.name}.FILE/${member}.MBR`)
+      .join(` `)
+      .replace(/([$\\])/g,'\\$1')
+      .toUpperCase();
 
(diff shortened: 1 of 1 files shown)


src/api/IBMiContent.ts
```diff
@@ -651,7 +651,9 @@ export default class IBMiContent {
   }
 
   async memberResolve(member: string, files: QsysPath[]): Promise<IBMiMember | undefined> {
-    const command = `for f in ${files.map(file => `/QSYS.LIB/${file.library.toUpperCase()}.LIB/${file.name.toUpperCase()}.FILE/${member.toUpperCase()}.MBR`).join(` `)}; do if [ -f $f ]; then echo $f; break; fi; done`;
+    // Escape names for shell
+    const fixed = files.map(f => ({name: f.name.replace(/([$\\])/g,'\\$1'), library: f.library.replace(/([$\\])/g,'\\$1')}));
+    const command = `for f in ${fixed.map(file => `/QSYS.LIB/${file.library.toUpperCase()}.LIB/${file.name.toUpperCase()}.FILE/${member.toUpperCase()}.MBR`).join(` `)}; do if [ -f $f ]; then echo $f; break; fi; done`;
```
sebjulliand, line 656: Isn't `member` going to be a problem if it contains a `$` as well?
In which case it may be faster to escape it everywhere at once.
```suggestion
    const command = `for f in ${files.map(file => `/QSYS.LIB/${file.library}.LIB/${file.name}.FILE/${member}.MBR`.replace(/([$\\])/g,'\\$1').toUpperCase()).join(` `)}; do if [ -f $f ]; then echo $f; break; fi; done`;
```

sebjulliand (changes requested):
Shouldn't the member name be escaped as well?

src/api/IBMiContent.ts
worksofliam: Oh yeah... I am totally blind. Good spot

worksofliam: Moved the file list into a new variable, purely for readability.

worksofliam:
@sebjulliand Take another look when you can - thanks!

sebjulliand (approved):
Perfect; code is clear and it works fine! Let's merge.

PR #1659, opened 2023-11-18
Ignore stat error when no hidden files in directory

chrjorgensen:
### Changes

This PR will fix the old error of `stat` complaining `cannot stat '.*'` by adding a filter for it - like the already filtered error `cannot stat '*'`.

### Checklist

* [x] have tested my change


sebjulliand (approved):
Looks good to me 😄

chrjorgensen:
Thanks @sebjulliand !

PR #1664, opened 2023-11-19
Add "Run on protected" option on Actions

sebjulliand:
### Changes
Resolves https://github.com/codefori/vscode-ibmi/issues/1400

This PR adds a new `Run on protected` attribute to Actions:
Capture d’écran du 2023-11-19 16-04-49

If this attribute is enabled, then the Action will be allowed to be selected and to run on read-only members or from protected filters.

Before this PR, no action was allowed to run on protected or read-only targets.

### Checklist
* [x] have tested my change


package.json
```diff
@@ -571,6 +571,10 @@
 									"filter",
 									"browser"
 								]								
+							},
+							"runOnProtected": {
+								"type": "boolean",
+								"description": "If enabled, this Action will be allowed to run on protected/read-only targets."
```
worksofliam, line 577: I think we have used 'readonly' in labels, instead of 'read-only'

src/locale/ids/da.ts
```diff
@@ -269,5 +269,7 @@ export const da: Locale = {
   'actions.workAction.refresh.filter':'Filter',
   'actions.workAction.refresh.filter.description':'Det aktuelle filter bliver opfrisket',
   'actions.workAction.refresh.browser':'Vindue',
-  'actions.workAction.refresh.browser.description':'Hele vinduet bliver opfrisket'
+  'actions.workAction.refresh.browser.description':'Hele vinduet bliver opfrisket',
+  'actions.workAction.runOnProtected':'Run on protected/read-only',
```
worksofliam, line 273: Same about 'readonly' here.

worksofliam (changes requested):
Other than some label changes, code looks ok to me. Re-request my review when it's ready! This is a neat change.

sebjulliand:
You're right @worksofliam ; the label we used was actually `read only` elsewhere.
@chrjorgensen may I request your expertise for the Danish translation? 😊

worksofliam (approved):
Works good for me! This is another nice UX change. You can see in this image that I have a locked member open and don't have any Actions available (as told by the error message).

image

In this screenshot you can see that I have one or more actions enabled in protected mode, and the list appears with them in correctly.

image

@sebjulliand Perhaps we should wait for Christian to add the correct translation and also let him test too before this is merged?

Thanks and great stuff!

worksofliam (approved):
@chrjorgensen Thank you!!

@sebjulliand Over to you for a merge whenever you like 😄

chrjorgensen (approved):
@sebjulliand Beautiful feature - works in my tests. 😍 

I made the Danish translation (and fixed some typos). Approved! Merge, Captain! 😃

sebjulliand:
Thank you guys! You've been brilliant, as always!

m-tyler:
Thanks a million for this enhancement.  Makes my life easier when I don't have `*ALLOBJ` on my user profile.
A list of my actions I marked to run in protected mode.  I don't really care if the standard CRT* actions are marked up as we have a CMS tool that should create objects outside of the developer environment. 

image

Issue #1672, opened 2023-11-20
QSH Compilation Command Fails When Library Name Contains '#'


bpmcpherson:
Hello all,

In the shell commands generated by the compilation actions, I noticed that no matter what I did, they seemed to fail with no output and little to no traceability. As a result, I did everything from check all possible SSH settings, file permissions, updates and reinstallations to no avail. 

liblist -d MYLIBCMD MYLIB LEGACY QGPL PRJPRDDTA PRJPRDOBJ #MENUI3 && liblist -c PRJPRDOBJ && liblist -a BMCPHERSON MYLIB PRJPRDOBJ QGPL QTEMP #MENUI3 && system "CRTSQLRPGI OBJ(PRJPRDOBJ/RPG101) SRCFILE(PRJPRDOBJ/QRPGLESRC) SRCMBR(RPG101) OPTION(*EVENTF) DBGVIEW(*SOURCE) OBJTYPE(*PGM) TGTRLS(*CURRENT)"; if [[ $? -ne 0 ]]; then exit 1; fi

Run independently, I could see that the command would last mention interacting with "PRJPRDOBJ", but not #MENUI3. When I removed all instances of it from the command, it ran exactly as expected. Additionally, surrounding the library name with single quotes allowed it to function. 

Is it be possible in the buildLiblistCommands under the CompileTools.ts file to include an additional replacement using a regex expression to hunt down any of the library names which contain characters bash will not like and surround them with quotes? 

Thanks!

<hr />

|Context|Version|
|-|-|
|Code for IBM i version|2.4.0|
|Visual Studio Code version|1.84.2|
|Operating System|win32_x64|

<details>
<summary>Active extensions</summary>

```
COBOL (cobol): 9.11.10
Code for IBM i Walkthroughs (vscode-ibmi-walkthroughs): 0.4.0
Db2 for IBM i (vscode-db2i): 0.5.2
Emmet (emmet): 1.0.0
Error Lens (errorlens): 3.15.0
Git (git): 1.0.0
Git Base (git-base): 1.0.0
GitHub (github): 0.0.1
GitHub Authentication (github-authentication): 0.0.2
JSON Language Features (json-language-features): 1.0.0
Merge Conflict (merge-conflict): 1.0.0
Node Debug Auto-attach (debug-auto-launch): 1.0.0
TODO Highlight (vscode-todo-highlight): 1.0.5
TypeScript and JavaScript Language Features (typescript-language-features): 1.0.0
[...]

PR #1673, opened 2023-11-20
Sanitize library names before running liblist command

chrjorgensen:
### Changes

This PR will fix the error described in issue #1672.
It will also fix errors in other library list commands:
- show an error when changing current library, if the new current library name starts with `#`
- don't allow non-existing libraries with name starting with `#` when changing the library list

### Checklist

* [x] have tested my change

### DIFF
--- a/src/views/LibraryListView.ts
+++ b/src/views/LibraryListView.ts
@@ -3,7 +3,6 @@ import { ConnectionConfiguration, GlobalConfiguration } from "../api/Configurati
 import { instance } from "../instantiate";
 import { t } from "../locale";
 import { Library as LibraryListEntry } from "../typings";
-import { Tools } from "../api/Tools";
 
 export class LibraryListProvider implements vscode.TreeDataProvider<LibraryListNode>{
   private readonly _emitter: vscode.EventEmitter<LibraryListNode | undefined | null | void> = new vscode.EventEmitter();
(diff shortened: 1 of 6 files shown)


src/views/LibraryListView.ts
```diff
@@ -3,6 +3,7 @@ import { ConnectionConfiguration, GlobalConfiguration } from "../api/Configurati
 import { instance } from "../instantiate";
 import { t } from "../locale";
 import { Library as LibraryListEntry } from "../typings";
+import { Tools } from "../api/Tools";
```
worksofliam, line 6: Unused import was added.

worksofliam (changes requested):

worksofliam (changes requested):
I'd consider this a high impact change since it is used by a lot of our extension (and others!).

The tests are passing and things are working as normal.

Other than the unused import, the code and functionality works good.

Optionally, if you felt inclined, I'd love to see a new test case to show that it works with libraries that start with `#`.

src/views/LibraryListView.ts
chrjorgensen: Ups - a leftover from when I thought I should sanitize library names in this module. Fixed!

chrjorgensen:
@worksofliam I'm currently working on some tests - which revealed some defects with the handling of '$'!
We have IT auditing (💩) at work tomorrow, but I hope to have some fixes ready tomorrow evening or soon after...

chrjorgensen:
@worksofliam I fixed errors for sanitizing names with '$' and validating library list with sanitized names - and finally added some tests to verify library name sanitizing. Please have another go...

worksofliam (approved):
Looking good @chrjorgensen! Thanks for your work here.

PR #1680, opened 2023-11-23
feat(Go to file): Autotype suggestions typing path

chrjorgensen:
### Changes

This pull request is a squashed version of PR #1242 and must be used instead to make a cleaner git history.
The original author @sebCIL is kept on the commit.

### DIFF
--- a/src/instantiate.ts
+++ b/src/instantiate.ts
@@ -241,10 +241,6 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
           const lastPart = selectionSplit[selectionSplit.length - 1];
           const filterText = lastPart.substring(0, lastPart.indexOf(`*`));
 
-          // If the user has provided a value, then we can increase the max items.
-          // Otherwise, we'll limit it to 30 to prevent the query from taking too long.
-          const maxItems = (filterText ? 10000 : 30);
-
           let resultSet: Tools.DB2Row[] = [];
 
           switch (selectionSplit.length) {
@@ -317,6 +313,9 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
                   description: 'Searching members..',
                 },
               ]
+              // If the user has provided a value, then we can increase the max items.
+              // Otherwise, we'll limit it to 30 to prevent the query from taking too long.
+              const maxItems = (filterText ? 10000 : 30);
 
               resultSet = await content!.runSQL(`
                   SELECT cast(TABLE_PARTITION as char(10) for bit data) TABLE_PARTITION, 
@@ -327,7 +326,7 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
                     AND table_name = '${selectionSplit[1]}'
                     ${filterText ? `AND TABLE_PARTITION like '${filterText}%'` : ``}
                   ORDER BY 1
-                  LIMIT ${filterText ? 10000 : 30}
+                  LIMIT ${maxItems}
                 `);
 
               const listMember = resultSet.map(row => ({
@@ -358,7 +357,7 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
 
           // We remove the asterisk from the value so that the user can continue typing
           quickPick.value = quickPick.value.substring(0, quickPick.value.indexOf(`*`));
-          
+
         } else {
 
           if (filteredItems.length > 0) {
--- a/src/instantiate.ts
+++ b/src/instantiate.ts
@@ -224,14 +224,16 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
             ]
           }
         });
-
-
       }
+
+      let filteredItems: vscode.QuickPickItem[] = [];
 
       quickPick.onDidChangeValue(async () => {
 
         // INJECT user values into proposed values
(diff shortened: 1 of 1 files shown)


worksofliam (changes requested):
I think there is a lot of logic being handled in the wrong file.

I am thinking `code-for-ibmi.goToFile` (and `code-for-ibmi.goToFileReadOnly`) should likely go into a separate file.

Secondly, I think we need to be careful about using `onDidChangeValue` because a fast typing using is resulting in spamming the server. We need to look into using some kind of timeout to slow down the requests to the server. 

Something like this perhaps?

```ts
quickPick.onDidChangeValue(async () => {
  clearTimeout(timeout);
  timeout = setTimeout(() => {
    // Do something with quickPick.value
  }, 500);
});
```

chrjorgensen:
@worksofliam 

>Secondly, I think we need to be careful about using `onDidChangeValue` because a fast typing user is resulting in spamming the server. We need to look into using some kind of timeout to slow down the requests to the server.

Ah - I didn't understand your comment until I realized that `onDidChangeValue` would call the server at any time there was a `*` in the typed value. This was an error, it was only supposed to call the server when the typed value had a '*' in the end. I have fixed this in commit [574b2bd](https://github.com/codefori/vscode-ibmi/pull/1680/commits/574b2bdc1542c01b9e3a7f7f31f5486f728293fc).

I have also fixed some other small issues - can I make you test it one more time and see if it's better now?

I'm fine with the code being extracted into a separate file - but can we do that in 2.4.1? I will make an issue about this, if you are okay with this.

src/instantiate.ts
```diff
@@ -185,26 +193,174 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
...
+
+                filterText = selectionSplit[2].substring(0, selectionSplit[2].indexOf(`*`));
+                  
+                resultSet = await content!.runSQL(`SELECT cast(TABLE_PARTITION as char(10) for bit data) TABLE_PARTITION, 
+                  ifnull(PARTITION_TEXT, '') PARTITION_TEXT, 
+                  lower(ifnull(SOURCE_TYPE, '')) SOURCE_TYPE
+                FROM qsys2.SYSPARTITIONSTAT
+                WHERE TABLE_SCHEMA = '${selectionSplit[0]}'
+                  AND table_name = '${selectionSplit[1]}'
+                  AND TABLE_PARTITION like '${filterText}%'
```
worksofliam, line 310: Consider not adding this condition against `TABLE_PARTITION` when `filterText` is blank. It's evaluating to: `TABLE_PARITION like '%'`.

```
QSYSINC/H/*
```

src/instantiate.ts
```diff
@@ -185,26 +193,174 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
...
+
+                filterText = selectionSplit[1].substring(0, selectionSplit[1].indexOf(`*`));
+
+                resultSet = await content!.runSQL(`SELECT 
+                  ifnull(cast(system_table_name as char(10) for bit data), '') AS SYSTEM_TABLE_NAME, 
+                  ifnull(TABLE_TEXT, '') TABLE_TEXT 
+                FROM QSYS2.SYSTABLES 
+                WHERE TABLE_SCHEMA = '${selectionSplit[0]}' 
+                  AND FILE_TYPE = 'S' 
+                  AND SYSTEM_TABLE_NAME like '${filterText}%'
```
worksofliam, line 268: Same here about `filterText` being blank.

worksofliam (changes requested):
Functionally works good. I left some comments about the `where` clause being used for the filter. I also think we need an improved UX to show that it is loading data, perhaps with something like this before the SQL statements are executed?

```
                quickPick.items = [
                  {
                    label: 'Searching',
                    alwaysShow: true,
                    description: 'Please wait..',
                  },
                ]
```

worksofliam:
@chrjorgensen Here is a summary of my changes, which are quite a lot.

* You can not longer select the 'Loading' indicator 
* Stopped passing in a clause for `filterValue` if it is empty and limit the results based on the `filterValue` variable (see `maxItems` constant)
* Removed `forEach` and `push` combination in place of `map` on all result sets.
* Removed the previous caching mechanism due to the fact that I couldn't backtrack my filter, for example:
   1. If I filtered against `QSYSINC/MIH/*`, it would show me the members for that source file - great!
   2. In the same quick pick, if I change the value to `QSYSINC/H/*` then nothing would load.
* Keep the limit to 30 if the `filterValue` is empty, otherwise the limit is very high as the user is being more precise.
* After the filter items have loaded, remove the `*` so the user can filter the filtered items (filter-caption?!?!)

This is obviously a lot of changes, but think I did what you would think is right. If not, I am happy to back out my changes.

chrjorgensen:
@worksofliam Your changes worked great - but there were some old errors and other stuff, that I changed:

- Searching for names with variant characters did not work - fixed.
- I removed the limit on the SQL (`maxItems` constant), since it's fast enough even without the limit. 14000+ members were retrieved in approx. 5 seconds on my small test system.
- The typed value should not be added to the list (a leftover from copying from the `Change Current Library` function, I guess).
- I changed the search instruction to be more clear.

Please have a go. If you're pleased with it, let's make the release - we can always fix any small error later... 😃

rbeckham:
just got to try this new feature.  This is so awesome!!!  this was my biggest complaint with Code for i, such a great enhancements.  Thanks to all who contributed!

PR #1695, opened 2023-12-07
Add /QOpenSys/pkgs/bin to .bashrc if needed

sebjulliand:
### Changes
Resolves https://github.com/codefori/vscode-ibmi/issues/1691

This PR enhances the connection process and make it look for `/QOpenSys/pkgs/bin` in the `PATH` environment variable.
If it's not there, it will offer to either create `~/.bashrc` file or modify it to append `/QOpenSys/pkgs/bin` to the `PATH` environment variable.

The creation/modification is delayed in the connection process, so it's only done once the connection is ready to do it.

The connection process will only check this once or if the user connects and reload the server settings, so we won't bother someone who would not like to modify its `PATH`.

Since .bashrc is only relevant for non-login shell, this shouldn't interfere with `.profile` or `.bash_profile`.

### Checklist
* [x] have tested my change


chrjorgensen:
@sebjulliand Questions:

1. Is your code always run - or only run when the user shell is `bash`?
2. There is a test in the connection for `bash` as shell - is this code linked to that check? If not, maybe it should?

sebjulliand:
> @sebjulliand Questions:
> 
>     1. Is your code always run - or only run when the user shell is `bash`?
> 
>     2. There is a test in the connection for `bash` as shell - is this code linked to that check? If not, maybe it should?

Good questions! The code runs whatever shell is used and do not run only if bash is the current shell. On the other hand, we expect the shell to be bash and offer to change it if it's not; that's why I didn't think about conditioning this action to bash only.

What do you think?

chrjorgensen:
> we expect the shell to be bash

Do we expect it? It's not a requirement (to my knowledge), although it could be argued that it should be... however, the users may use the extension against older or customer systems, where `bash` or OSS environment is not installed.

My thoughts are that we should keep the `bash` checking code together and not have code in different places changing user settings.

Additional comment about PATH: /QOpenSys/pkgs/bin should be before /usr/bin and /QOpenSys/usr/bin - but you don't check for that. Would it be easy to have this check too?

sebjulliand:
> > we expect the shell to be bash
> 
> Do we expect it? It's not a requirement (to my knowledge), although it could be argued that it should be... however, the users may use the extension against older or customer systems, where `bash` or OSS environment is not installed.
> 
> My thoughts are that we should keep the `bash` checking code together and not have code in different places changing user settings.

Fair enough! We'll only do this if the current shell is bash then.

> Additional comment about PATH: /QOpenSys/pkgs/bin should be before /usr/bin and /QOpenSys/usr/bin - but you don't check for that. Would it be easy to have this check too?

I don't check it but since I always add /QOpenSys/pkgs/bin in first place in the path, that should not be necessary 😄

chrjorgensen:
> I don't check it but since I always add /QOpenSys/pkgs/bin in first place in the path, that should not be necessary

But your check will accept a PATH having /QOpenSys/pkgs/bin after the other ones - which is not correct (maybe a user mistake). 😉

sebjulliand:
> > I don't check it but since I always add /QOpenSys/pkgs/bin in first place in the path, that should not be necessary
> 
> But your check will accept a PATH having /QOpenSys/pkgs/bin after the other ones - which is not correct (maybe a user mistake). 😉
> 
> (Sorry for being such a P... in the A..! 😆 )

No, no, you're absolutely right 😅

sebjulliand:
@chrjorgensen I followed your advice and:
- Only run this whole process is bash is the current shell
- Check if `/QOpenSys/pkgs/bin` appears before anything else in `PATH`

Hopefully this will receive the "Jørgensen Seal of Approval" 😁

src/api/IBMi.ts
```diff
@@ -691,47 +692,115 @@ export default class IBMi {
...
+                  }
+                  else if (currentPaths.indexOf("/QOpenSys/pkgs/bin") > 0){
+                    reason = "/QOpenSys/pkgs/bin is not in first position in your $PATH shell environment variable";
+                  }
+                  if (reason && await vscode.window.showWarningMessage(`/QOpenSys/pkgs/bin not found in $PATH`, {
+                    modal: true,
+                    detail: `${reason}, so Code for IBM i may not function correctly. Would you like to ${bashrcExists ? "update" : "create"} ${bashrcFile} to fix this now?`,
+                  }, `Yes`)) {
+                    delayedOperations.push(async () => {
+                      console.log(`${bashrcExists ? "update" : "create"} ${bashrcFile}`);
```
worksofliam, line 755: Perhaps this should log to our output channel instead?

src/api/IBMi.ts
```diff
@@ -691,47 +692,115 @@ export default class IBMi {
...
+                if ((!quickConnect || !cachedServerSettings?.pathChecked)) {
+                  const currentPaths = (await this.sendCommand({ command: "echo $PATH" })).stdout.split(":");
+                  const bashrcFile = `${defaultHomeDir}/.bashrc`;
+                  let bashrcExists = (await this.sendCommand({ command: `test -e ${bashrcFile}` })).code === 0;
+                  let reason;
+                  if(!currentPaths.includes("/QOpenSys/pkgs/bin")){
+                    reason = "Your $PATH shell environment variable does not include /QOpenSys/pkgs/bin";
+                  }
+                  else if (currentPaths.indexOf("/QOpenSys/pkgs/bin") > 0){
+                    reason = "/QOpenSys/pkgs/bin is not in first position in your $PATH shell environment variable";
```
worksofliam, line 748: Quick question. Why does it need to be in the first poisition? I think it'd be ok as long as it were on the `PATH` in general.

Perhaps:

```js
if (currentPaths.indexOf("/QOpenSys/pkgs/bin") === -1) {
  reason = "/QOpenSys/pkgs/bin is not in your $PATH shell environment variable.";
}
```

worksofliam (commented):
Left some comments. I think the `console.log` should be removed.

src/api/IBMi.ts
sebjulliand: Indeed!

src/api/IBMi.ts
sebjulliand: OK, I changed that test according to your comment.

worksofliam (approved):
Looks like it works correctly for me! 

image

image

(I had to connect and reload as expected to test the positional case)

PR #1700, opened 2023-12-07
IFS browser enhancements

chrjorgensen:
### Changes

This PR changes the IFS browser behavior in the following ways:
- When copying or moving an object, only the object name will be selected in the prompt for faster entering of new object name. Previously the full path was selected, which seemed a bit cumbersome.
- When creating a new directory or streamfile, the prompt is changed to not select the full path, but instead place the cursor after the path. This makes it faster to enter the name of the new directory or streamfile.
- A protection mechanism for vital IBM directories has been created. Now the following directories will not have the menu options for delete, copy or move:
  - /
  - /QOpenSys
  - /QSYS.LIB
  - /QDLS
  - /QOPT
  - /QNTC
  - /QFileSvr.400
  - /bin
  - /dev
  - /home
  - /tmp
  - /usr
  - /var

The PR will make it faster working with objects in the IFS and increase the security when working in the IFS.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/views/ifsBrowser.ts
+++ b/src/views/ifsBrowser.ts
@@ -12,6 +12,8 @@ import { instance, setSearchResults } from "../instantiate";
 import { t } from "../locale";
 import { BrowserItem, BrowserItemParameters, FocusOptions, IFSFile, WithPath } from "../typings";
 
+const Protected_dirs = /^(\/|\/QOpenSys|\/QSYS\.LIB|\/QDLS|\/QOPT|\/QNTC|\/QFileSvr\.400|\/bin|\/dev|\/home|\/tmp|\/usr|\/var)$/i;
+
 class IFSBrowser implements vscode.TreeDataProvider<BrowserItem> {
   private readonly emitter = new vscode.EventEmitter<BrowserItem | BrowserItem[] | undefined | null | void>();
   readonly onDidChangeTreeData = this.emitter.event;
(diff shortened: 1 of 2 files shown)


src/views/ifsBrowser.ts
```diff
@@ -12,6 +12,8 @@ import { instance, setSearchResults } from "../instantiate";
 import { t } from "../locale";
 import { BrowserItem, BrowserItemParameters, FocusOptions, IFSFile, WithPath } from "../typings";
 
+const Protected_dirs = /^(\/|\/QOpenSys|\/QSYS\.LIB|\/QDLS|\/QOPT|\/QNTC|\/QFileSvr\.400|\/bin|\/dev|\/home|\/tmp|\/usr|\/var)$/i;
```
worksofliam, line 15: My only qualm is that this variable name doesn't go very well with the rest of the code. Perhaps `ProtectedDirs` or `PROTECTED_DIRS` since it is a const.

worksofliam (approved):
Looks like it is working as expected on those protected directories. I love this UX change!

image

The highlighted text is a super nice change!!

image

worksofliam:
@chrjorgensen Merge whenever you feel comfy. Thanks!

src/views/ifsBrowser.ts
chrjorgensen: Agree - fixed

sebjulliand (approved):
Very nice UX change! 😍

PR #1701, opened 2023-12-07
Fix goto file quickpick

chrjorgensen:
### Changes

This PR will fix some issue with the new search function in Go To File:
- The list label sort is disabled, since the list is already sorted when sent to UI.
- A progress bar will be shown when searching the server.
- The list is now more dense by showing text as description instead of detail.
- The labels are now shown correctly
- The list is now correctly changed to cached files when the value is cleared.
- The `Clear list` will always be shown, now with a separator line above and an icon in front.

### Checklist

* [x] have tested my change


worksofliam (approved):
* I can see the new progress bar. That is awesome.
* I like the dense list a lot! 
* Line seperator for the 'clear list' item is a nice touch.

Merge whenever you're ready!

image

image

PR #1703, opened 2023-12-11
Fixes/deployment_working_directory

worksofliam:
### Changes

* Adds `&WORKDIR` for local actions, which is the deployment directory and is not the same as `&HOME` (as they can be different)
* Fixes issue where remote working directory was being changes by deployment. It no longer does this and instead uses the deploy directory when running commands (and we added `&WORKDIR`)
* Cleanup 'compare' deploy API to have seperation of delete login (this will require a type change in the API)

### Checklist

* [x] have tested my change


sebjulliand (approved):
LGTM; gave it another test round and it looks OK.
And I was here while you coded this 😅

Feel free to merge when you're OK with it.

PR #1706, opened 2023-12-11
ASP support member lookup

worksofliam:
### Changes

Similar to opening files, `memberResolve` will now look in the user defined ASP to find a member. Additionally, the API can accept an ASP and will use that instead.

Also solves a bug with the `qualifyPath` which wasn't handling `QSYS` correctly.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -681,14 +681,15 @@ export default class IBMiContent {
       .map(file => {
         const asp = file.asp || this.config.sourceASP;
         if (asp && asp.length > 0) {
-          return [`/${asp}/QSYS.LIB/${file.library}.LIB/${file.name}.FILE/${member}.MBR`, `/QSYS.LIB/${file.library}.LIB/${file.name}.FILE/${member}.MBR`].join(` `);
+          return [
+            Tools.qualifyPath(file.library, file.name, member, asp, true), 
+            Tools.qualifyPath(file.library, file.name, member, undefined, true)
+          ].join(` `);
         } else {
-          return `/QSYS.LIB/${file.library}.LIB/${file.name}.FILE/${member}.MBR`;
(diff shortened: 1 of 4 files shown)


src/api/IBMiContent.ts
```diff
@@ -678,7 +678,15 @@ export default class IBMiContent {
...
-      .map(file => `/QSYS.LIB/${file.library}.LIB/${file.name}.FILE/${member}.MBR`)
+      .map(file => {
+        const asp = file.asp || this.config.sourceASP;
+        if (asp && asp.length > 0) {
+          return [`/${asp}/QSYS.LIB/${file.library}.LIB/${file.name}.FILE/${member}.MBR`, `/QSYS.LIB/${file.library}.LIB/${file.name}.FILE/${member}.MBR`].join(` `);
+        } else {
+          return `/QSYS.LIB/${file.library}.LIB/${file.name}.FILE/${member}.MBR`;
+        }
+      })
+      .map(file => Tools.escapePath(file))
```
worksofliam, line 689: @chrjorgensen Is this the valid way to use escape path? The tests continue to pass and this should ensure that `$` can be used in the path itself.

chrjorgensen: @worksofliam LGTM - even though I haven't had anything to do with `Tools.escapePath`... you may be thinking of `Tools.sanitizeLibraryNames`?

Using `Tools.escapePath` should be okay, when the library is not preceded by a space (which caused trouble in the shell when the library name started with `#`).

worksofliam: Supposed I better use `Tools.sanitizeLibraryNames` too!

worksofliam: @chrjorgensen I actually used both in a new commit. Made a well needed improved to one of our existing APIs.

worksofliam:
Tests continue to pass.

Issue #1704, opened 2023-12-11
Possible loop


pabloto:
Hi, I attach my codefori output log, there's a possible issue.

we work in local, and we have an iasp, I don't know why in a certain moment codefori try to:

.: system "CPYTOSTMF FROMMBR('/QSYS.lib/AL400MNUV2.lib/QRPGLESRC.file/UMIFUNC_H.mbr') TOSTMF('/tmp/vscodetemp-O_iXrqBaLz') STMFOPT(*REPLACE) STMFCCSID(1208) DBFCCSID(*FILE)"
{
    "code": 255,
    "signal": null,
    "stdout": "",
    "stderr": "CPFA0A9: L'oggetto non è stato trovato. L'oggetto è /QSYS.lib/AL400MNUV2.lib/QRPGLESRC.file/UMIFUNC_H.mbr.\nCPFA097: Oggetto non copiato. L'oggetto è /QSYS.lib/AL400MNUV2.lib/QRPGLESRC.file/UMIFUNC_H.mbr."

this object really doesn't exists, it is on /DIRECTA1/QSYS.lib/AL400MNUV2.lib/QRPGLESRC.file/UMIFUNC_H.mbr

DIRECTA1 is the name of my iasp.

I try to downgrade to the previous version but the problem seems to be the same.

Could you check??

I try to clrlib vscode.

Today I have more then 160.000 jobs on my system, deleting the jobs QP0ZSPWT  it remains only 5.000. changing my qdftsrv jobd to log (4 00 *nolist) the joblog doesn't remain on the system.

Thank you
[Codefori_output.txt](https://github.com/codefori/vscode-ibmi/files/13638666/Codefori_output.txt)

<hr />

⚠️ **REMOVE THIS LINE AND ANY SENSITIVE INFORMATION BELOW!** ⚠️

|Context|Version|
|-|-|
|Code for IBM i version|2.5.0|
|Visual Studio Code version|1.85.0|
|Operating System|darwin_arm64|

<details>
<summary>Active extensions</summary>

```
COBOL (cobol): 9.11.22
Code Spell Checker (code-spell-checker): 3.0.1
Code for IBM i Walkthroughs (vscode-ibmi-walkthroughs): 0.4.0
Db2 for IBM i (vscode-db2i): 0.6.0
Dev Containers (remote-containers): 0.327.0
ESLint (vscode-eslint): 2.4.2
Emmet (emmet): 1.0.0
Error Lens (errorlens): 3.15.0
[...]

PR #1707, opened 2023-12-11
Fix missing throw in downloadMemberContent

worksofliam:
### Changes

I noticed when vscode-rpgle tries to fetch rpglint.json from the QSYS file system, it was spamming if it was not found. This was due to some invalid retry logic where it would just get caught in a loop. This is also possibly related to the issue @pabloto found in #1704. You would see this, including for all missing includes, in the output log.

```
.: system "CPYTOSTMF FROMMBR('/QSYS.lib/LIAMA.lib/VSCODE.file/RPGLINT.mbr') TOSTMF('/tmp/vscodetemp-O_7cyB9KET') STMFOPT(*REPLACE) STMFCCSID(1208) DBFCCSID(*FILE)"
{
    "code": 255,
    "signal": null,
    "stdout": "",
    "stderr": "CPFA0A9: Object not found.  Object is /QSYS.lib/LIAMA.lib/VSCODE.file/RPGLINT.mbr.\nCPFA097: Object not copied.  Object is /QSYS.lib/LIAMA.lib/VSCODE.file/RPGLINT.mbr."
}
```

#### Recreate

1. Check out to `master`
2. Have vscode-rpgle installed
3. Make sure the `VSCODE` source file exists, but make sure no `RPGLINT.JSON` member exists in it
4. Create a brand new SQL/RPGLE member in the same library: `source.rpgle`
5. See the spam (like above) in the output console

#### The fix

1. Checkout this PR and launch in debug
2. Have vscode-rpgle installed
3. Make sure the `VSCODE` source file exists, but make sure no `RPGLINT.JSON` member exists in it
4. Create a brand new SQL/RPGLE member in the same library: `source.rpgle`
5. See it (like above) written to the output log once and then stop.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -8,6 +8,7 @@ import { CommandResult, IBMiError, IBMiFile, IBMiMember, IBMiObject, IFSFile, Qs
 import { ConnectionConfiguration } from './Configuration';
 import { default as IBMi } from './IBMi';
 import { Tools } from './Tools';
+import { r } from 'tar';
 const tmpFile = util.promisify(tmp.file);
 const readFileAsync = util.promisify(fs.readFile);
 const writeFileAsync = util.promisify(fs.writeFile);
@@ -149,11 +150,13 @@ export default class IBMiContent {
               }
               break;
             default:
-              throw e;
+              retry = false;
+              break;
           }
         }
-        else {
-          throw e;
+
+        if (!retry) {
+          throw e
         }
       }
     }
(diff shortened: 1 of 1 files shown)


src/api/IBMiContent.ts
```diff
@@ -8,6 +8,7 @@ import { CommandResult, IBMiError, IBMiFile, IBMiMember, IBMiObject, IFSFile, Qs
 import { ConnectionConfiguration } from './Configuration';
 import { default as IBMi } from './IBMi';
 import { Tools } from './Tools';
+import { r } from 'tar';
```
sebjulliand, line 11: This should be removed.

sebjulliand (changes requested):
Works as expected! I could recreate the spamming before but not after.
Just remove that unused `import` to keep it clean and merge away.
Thanks!

PR #1708, opened 2023-12-12
Removed deprecated commands call and ignore library list

sebjulliand:
### Changes
This PR includes the following changes:
- It removes the deprecated methods `remoteCommand` and `paseCommand` from `IBMi`.
- It replaces the calls to these two methods with `sendCommand` or `runCommand` depending on the context.
- It adds a `noLibList` option to `runCommand`. If it is set to `true`, `runCommand` will not build the library list before running the command. This will fix issues we had with people seeing their members as read-only because of a faulty library list. Other command calls that should ignore the library list have been updated as well.

### DIFF
--- a/src/api/CompileTools.ts
+++ b/src/api/CompileTools.ts
@@ -536,17 +536,14 @@ export namespace CompileTools {
     if (config && connection) {
       const cwd = options.cwd;
 
-      let ileSetup: ILELibrarySettings = {
+      const ileSetup: ILELibrarySettings = {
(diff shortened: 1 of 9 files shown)


worksofliam:
@sebjulliand PRs like this is why I love having a test suite!

image

worksofliam (approved):

sebjulliand:
> @sebjulliand PRs like this is why I love having a test suite!
> image

I know, right? 😁

PR #1723, opened 2023-12-14
Feature/local_improvements_2

worksofliam:
### Features

* [x] New optional feature to automatically create a branch library when switching/creating a branch
    * Or, if the library already exists when they check out to branch, it will ask if they want to create a filter for it if they do not have one.
* [x] Variables now appear in `actions.json`, making it easier to understand the content

### Internal changes

Variable expansion was broken. If I had an `.env` like this before, it would not expand when building the library list. This is now resolved.

```properties
CURLIB=&BRANCHLIB
```

Secondly, we had some type mis-matches for variables. In `runAction` we were using `Map<string, string>`, but the `RemoteCommand` interface is using `Record<string, string>` (the same as `{[key: string]: string})`). This was a pain and because changing `RemoteCommand` risks breaking a lot of APIs (including external extensions using our API) I felt that a reasonable idea to convert `Variables` to be a `Record<string, string>`. Sorry @sebjulliand !! But, it does mean variable expansion of variables works in almost every scenario now.

## Checklist

* [x] have tested my change


package.json
```diff
@@ -448,6 +452,11 @@
 					"default": false,
 					"description": "When lines with ILE diagnostics are edited, the diagnostic will be removed. Requires reload/restart if changed."
 				},
+				"code-for-ibmi.createLibraryOnBranchChange": {
+					"type": "boolean",
+					"default": false,
+					"description": "Used in conjuction with git. When the branch changes, prompt the user if they would like a new branch library for this branch."
```
Wright4i, line 458: Would like to see this mention the VSxxxxxxxx generated name somehow.

src/api/local/git.ts
```diff
@@ -0,0 +1,117 @@
...
+                  });
+
+                  config.objectFilters = filters;
+                  ConnectionConfiguration.update(config).then(() => {
+                    commands.executeCommand(`code-for-ibmi.refreshObjectBrowser`);
+                  });
+                }
+              });
+            })
+            .catch(err => {
```
Wright4i, line 110: Fails to catch when you are not authorized to create the library and will prompt you to make a filter anyway, which then doesn't work because of the failed create command.
image

src/api/local/git.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+  }
+}
+
+export function setupGitEventHandler(context: ExtensionContext, workspaceFolders: WorkspaceFolder[]) {
+  const gitApi = Tools.getGitAPI();
+
+  if (gitApi) {
+    for (const workspaceFolder of workspaceFolders) {
+      const initRepo = gitApi.getRepository(workspaceFolder.uri);
```
Wright4i, line 26: I don't know if this is a bug or I am just stupid, but I had a hell of a time trying to get this to not return null for a git report I initialized in vscode as part of the review. I did a lot of googling, made sure it was actually a repo, had commits, folder authority was good, re-opened the repository, was definitely in my workspace, tried different file paths and stepped through on debug making sure the uri was good and it was. 

I was able to continue my testing though after cloning a git repo and opening that as my workspace. So I'm completely OK with calling the above user-error or a problem on my machine, but I'd like you to try making a new repo and see if you see the same issue.

Wright4i (changes requested):
@worksofliam see 3 pending code comments. Thanks!

src/api/local/git.ts
worksofliam: Resolved. I wasn't handling the error correctly.

src/api/local/git.ts
worksofliam: Yeah, that makes sense. `setupGitEventHandler` is called when the extension is loaded. When the extension is loaded, but it's not a git repo, then any action against it is ignored. So if you load the extension, and it's a git repo, we listen to it. For now, I think this is expected behaviour.

worksofliam:
@Wright4i I fixed two of the three issues you raised - thanks for a great review.

Sadly, we need #1741 to be merged into this branch for the fixes to take affect right now. I will re-request a review when that is done.

Wright4i (approved):
Fixes look good. Approved

worksofliam:
@Wright4i Requesting your review again! I fixed the issue you described above: we can now detect when a repo is initialized and handle those branch changes too!

Please try again and let me know what you think.

Wright4i (approved):
LGTM! Fix worked and prompted to create a library on a branch switch after creating a new git repo.
image

worksofliam:
@Wright4i Any chance I can bother you to retest this again after a fairly large merge? The same tests as before would be fab. Thank you.

Wright4i (approved):
@worksofliam No issues with my tests after the merge & typo fix. Code looks solid. Good to go 👍

PR #1726, opened 2023-12-14
Added Drag and Drop support to IFS Browser

sebjulliand:
### Changes
This PR enables drag and drop support in the IFS browser and multiple selection (using `ctrl + click` or `shift + click`)

#### Drag and drop from outside VSCode inside the IFS Browser
Dragging files/folders from an external explorer to the IFS browser will upload these files/folders. If the users drop their selection on a folder, that folder will be the target for the upload. If they drop on a file, the file's parent folder will be the target.
xB4fDZa7QK

#### Drag and drop inside the IFS Browser
Dragging one or more items from the IFS browser and dropping them in it will copy or move them depending on a new setting:
image
- **ask**: a modal dialog will appear to ask the user if they want to move or copy their selection
- **move**: the selection will be moved to the drop target
- **copy**: the selection will be copied to the drop target
Code_GxYgB3OLUR

#### Multi-selection actions
Most of the actions from the right-click menu in the IFS browser will be disabled, save for the `Delete object` action that now allows for deleting multiple items at once.

@chrjorgensen : I'll once again require your expertise for translating a few new sentences (+ one that was slightly changed).

Resolves https://github.com/codefori/vscode-ibmi/issues/1725 by adding a progress when uploading files.

### Checklist
* [x] have tested my change


chrjorgensen:
@sebjulliand Whoa! Can't wait to test this one!! 😍

chrjorgensen:
@sebjulliand Awesome - I've been wanting this for so long! 😍 

I gave it a spin, and it works as expected - of course! 😉 

Some thoughts from my side:
- I think it should work like the built-in file explorer as much as possible. In the file explorer the default is to move the items, and when pressing `Ctrl` the operation is a copy. Can this be done?
- In the file explorer there's a confirmation dialog before the move/copy operation. Maybe the `ask` option should be permanent?
- The file explorer shows the number of selected items at the cursor. Can we do this?
- Now the user (me) can upload files by dragging into the IFS browser view, he/she (I) would expect to be able to do the opposite - downloading by dragging out of the IFS browser view. Can this be done?

As always, don't give me anything, I just want more! 🤣 

I translated the new texts into Danish - and added checking for protected directories, including changing and translating the message text. Pardon my French! 😉

chrjorgensen (changes requested):
LGTM - see comment above...

sebjulliand:
Thanks @chrjorgensen, goods questions!

> Some thoughts from my side:
> 
>     * I think it should work like the built-in file explorer as much as possible. In the file explorer the default is to move the items, and when pressing `Ctrl` the operation is a copy. Can this be done?

The drag and drop API doesn't allow this. I tried to find a way to mimic this but didn't find any.

>     * In the file explorer there's a confirmation dialog before the move/copy operation. Maybe the `ask` option should be permanent and the dialog changed into this?

I don't have that confirmation. Could it be a settings too? If I drag and drop files around in the file explorer, it does it right away.

>     * The file explorer shows the number of selected items at the cursor. Can we do this?
It already does!
image

>     * Now the user (me) can upload files by dragging into the IFS browser view, he/she (I) would expect to be able to do the opposite - downloading by dragging out of the IFS browser view. Can this be done?
I failed to find a way to do this as well, and god knows I tried 😅 If it's possible, I didn't find how yet (maybe with a specific mime type?). I'll try harder...

chrjorgensen:
@sebjulliand 

> The drag and drop API doesn't allow this. I tried to find a way to mimic this but didn't find any.

Okay, just a wish that may come true later.. 😃 

> I don't have that confirmation. Could it be a settings too? If I drag and drop files around in the file explorer, it does it right away.

You have probably checked this one to not be disturbed! 😆 

image

> It already does!

Ah, sorry - only had one file selected when I checked for this. 👍 

> I failed to find a way to do this as well, and god knows I tried 😅 If it's possible, I didn't find how yet (maybe with a specific mime type?). I'll try harder...

Well, this is not a show-stopper, just a nice-to-have... 😃

chrjorgensen:
@sebjulliand Did you know you can drag a file into an open editor? Then it copies the link into the text:

image

Not a problem, though...

chrjorgensen (changes requested):
@sebjulliand I did some further testing on the PR and observed this:

When I select some files and move them to another folder. the next item below is selected:

image

image

Also, it's possible to drag a folder into itself, which should not be possible, instead ignored...

sebjulliand:
Thanks @chrjorgensen for your review!
I completely messed up my rebase and had to clean thing up and force a clean branch but here we are.

Here is what's been done, based on your feedback:
- Dragging a folder into itself will now do nothing
- After a successful move operation, the target folder gets focused
- Dragging files into the editor area will open them

chrjorgensen (approved):
@sebjulliand It's looking very good indeed - no issues found, and your fixes for my latest remarks works great! :rose: 

Can't wait to get this into C4i - really like how easy it is now to move/copy files and open them in the editor! :heart_eyes:

chrjorgensen:
@sebjulliand (Rebase and) merge when you're ready! :1st_place_medal:

sebjulliand:
You're the best @chrjorgensen, thank you for your review!
I'm quite happy with the result too! Let's merge 😁

PR #1728, opened 2023-12-15
Correct all places where it was expected runCommand to crash

worksofliam:
### Changes

Noticed when testing the new connection settings PR that when reloading or creating a new connection, I was always seeing this message about the old data area that we don't like, even though it didn't exist.

Previously, we relied on an API that was throwing an exception if the returned status code was not 0. We removed that API in place of an API that never crashes, but we didn't change the code around it. This is a pretty important fix and likely needs to be deployed ASAP.

Highest impact areas:

* Creating the connection
* Object Browser UI/UX

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/CompileTools.ts
+++ b/src/api/CompileTools.ts
@@ -552,7 +552,7 @@ export namespace CompileTools {
   /**
    * Execute a command
    */
-  export async function runCommand(instance: Instance, options: RemoteCommand, writeEvent?: EventEmitter<string>): Promise<CommandResult | null> {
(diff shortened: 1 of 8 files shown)


worksofliam:
Test cases are passing.

image

worksofliam:
My test cases are below. For each error I tested, I also tested they worked too.

<details>
<summary>Click me</summary>

Create error showing:

image

Rename error:

image

Switching from `master` to this PR branch shows that the error about the data area stops appearing when connecting and reloading the settings (or connecting for the first time)

</details>

Issue #1732, opened 2023-12-15
Is the new `Searching for members 🔍` feature replacing how `Go to File...` used to work?


chrjorgensen:
### Discussed in https://github.com/orgs/codefori/discussions/1730

<div type='discussions-op-text'>

<sup>Originally posted by **m-tyler** December 15, 2023</sup>
Since the feature update for `Go to File...` was implemented I have been unable to open a specific member reference.  If I type string like `lib/file/member.ext` nothing happens.  Pressing Enter will not cause the tool to open the member.  I have to remove the extension, type an asterisk (*) and wait for the query to finish (which on our 7.5 sometimes takes a long time).  Once the action of the search has completed for a particular source file the search is usually quicker to return results. 

So am I not able to specify an specific member path anymore?

Thanks, Matt</div>

PR #1733, opened 2023-12-15
Fix error opening a specific member reference in Go To File

chrjorgensen:
### Changes

This PR will fix issue #1732 by (re)adding the entered value to the list, thus allowing it to be selected (accepted).
This was a regression from the new member search in function Go To file.

Additional fixes are included:
- Validation of entered member reference - previously the user could enter any extension for the member, and the member would still open. Now there is a warning if the member does not exist or the member type is wrong.
- Search now works even if member and dot has been specified when entering '*'.
- Member type is shown in uppercase like the other parts in a list item.
- SQL statements and use of metadata columns are now consistent.
- The Go To file dialog are kept open when the list is cleared.
- Validation of entered streamfile before open - keeping the list available in case of error.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/instantiate.ts
+++ b/src/instantiate.ts
@@ -409,29 +409,27 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
               if (config && config.enableSQL && !selection.startsWith(`/`)) {
                 const lib = `${connection!.sysNameInAmerican(selectionSplit[0])}`;
                 const file = `${connection!.sysNameInAmerican(selectionSplit[1])}`;
-                const member = `${connection!.sysNameInAmerican(selectionSplit[2])}`;
-                let memberName, memberType = ``;
-                if (member.includes(`.`)) {
-                  memberName = member.slice(0, member.lastIndexOf(`.`));
-                  memberType = member.slice(member.lastIndexOf(`.`) + 1,);
-                } else {
-                  memberName = member;
-                };
+                const member = path.parse(`${connection!.sysNameInAmerican(selectionSplit[2])}`);
+                member.ext = member.ext.substring(1);
                 const fullMember = await content!.runSQL(`
                   select rtrim( cast( SYSTEM_TABLE_MEMBER as char( 10 ) for bit data ) ) as MEMBER
                        , rtrim( coalesce( SOURCE_TYPE, '' ) ) as TYPE
                     from QSYS2.SYSPARTITIONSTAT
-                   where ( SYSTEM_TABLE_SCHEMA, SYSTEM_TABLE_NAME, SYSTEM_TABLE_MEMBER ) = ( '${lib}', '${file}', '${memberName}' )
+                   where ( SYSTEM_TABLE_SCHEMA, SYSTEM_TABLE_NAME, SYSTEM_TABLE_MEMBER ) = ( '${lib}', '${file}', '${member.name}' )
                    limit 1
-                `).then((resultSet) => { return resultSet.length === 1 ? `${resultSet[0].MEMBER}.${resultSet[0].TYPE}` : `` });
+                `).then((resultSet) => { return resultSet.length !== 1 ? {} :
+                  { base: `${resultSet[0].MEMBER}.${resultSet[0].TYPE}`,
+                    name: `${resultSet[0].MEMBER}`,
+                    ext: `${resultSet[0].TYPE}`,
+                  }});
                 if (!fullMember) {
-                  vscode.window.showWarningMessage(`Member ${lib}/${file}/${member} does not exist.`);
+                  vscode.window.showWarningMessage(`Member ${lib}/${file}/${member.base} does not exist.`);
                   return;
-                } else if (fullMember !== `${memberName}.${memberType}`) {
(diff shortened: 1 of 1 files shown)


src/instantiate.ts
```diff
@@ -386,20 +392,55 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
...
+                const lib = `${connection!.sysNameInAmerican(selectionSplit[0])}`;
+                const file = `${connection!.sysNameInAmerican(selectionSplit[1])}`;
+                const member = `${connection!.sysNameInAmerican(selectionSplit[2])}`;
+                let memberName, memberType = ``;
+                if (member.includes(`.`)) {
+                  memberName = member.slice(0, member.lastIndexOf(`.`));
+                  memberType = member.slice(member.lastIndexOf(`.`) + 1,);
+                } else {
+                  memberName = member;
+                };
```
worksofliam, line 419: I am thinking this could be simpler than using `lastIndexOf` and `includes`.

I think we should use `path.parse(member)` here as it will take care of getting the name and extension out of the entered value correctly.

```
> member
'ABCD.RPGLE'
> path.parse(member)
{ root: '', dir: '', base: 'ABCD.RPGLE', ext: '.RPGLE', name: 'ABCD' }
> noext
'THEMBR'
> path.parse(noext)
{ root: '', dir: '', base: 'THEMBR', ext: '', name: 'THEMBR' }
```

worksofliam (changes requested):
Left one small suggestion. Code looks pretty good otherwise!

src/instantiate.ts
chrjorgensen: Now uses `path.parse` instead...

src/instantiate.ts
```diff
@@ -386,20 +392,53 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
...
+      quickPick.onDidAccept(async () => {
+        let selection = quickPick.selectedItems[0].label;
         if (selection && selection !== LOADING_LABEL) {
           if (selection === clearList) {
             storage!.setSourceList({});
+            quickPick.items = clearListArray;
             vscode.window.showInformationMessage(`Cleared list.`);
-            quickPick.hide()
           } else {
             const selectionSplit = selection.split('/')
```
worksofliam, line 407: Looks like `selectionSplit` doesn't account for case, so if a lowercase value is provided, the rest of this code fails because it expects members in uppercase for the SQL statement.

```suggestion
            const selectionSplit = selection.toUpperCase().split('/');
```

src/instantiate.ts
```diff
@@ -386,20 +392,53 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
...
             if (selectionSplit.length === 3 || selection.startsWith(`/`)) {
+              if (config && config.enableSQL && !selection.startsWith(`/`)) {
+                const lib = `${connection!.sysNameInAmerican(selectionSplit[0])}`;
+                const file = `${connection!.sysNameInAmerican(selectionSplit[1])}`;
+                const member = path.parse(`${connection!.sysNameInAmerican(selectionSplit[2])}`);
+                member.ext = member.ext.substring(1);
+                const fullMember = await content!.runSQL(`
```
worksofliam, line 414: SQL statement makes sense. I wonder if it would be better (or faster?) to use `CHKOBJ` with the `MBR` parameter here? What do you think?

We have the `checkObject` API, but might need a new parameter for member.

worksofliam: Also, makes sense if you want to leave it like this.

worksofliam (changes requested):
Only one real issue with casing, other than that it works as expected now!

src/instantiate.ts
chrjorgensen: The selection will always be in uppercase, if it's a member:
- if the member is entered by the user, the value is put in the list in uppercase
- if the member is from a search, it is in uppercase by SQL
So it should not be necessary to convert to uppercase. We also don't want to convert streamfiles to uppercase.

src/instantiate.ts
chrjorgensen: I prefer the SQL - is already in use by the search.
And SQL returns the type, which can then be validated.
If the user specifies the member without type, it will be opened if found - if the type is specified, both name and type must match.

src/instantiate.ts
worksofliam: Can you confirm this is user error then? I have all these lowercase listings in my quick pick from prior usage.

image

worksofliam: I also have this enabled, so it looks like it is adding lowercase names to the list! 
image

worksofliam: Looks like it's this line of code storing the values, including as lowercase when that option is enabled: https://github.com/chrjorgensen/code-for-ibmi/blob/d863157505d93b76388c9c9092fee9c398375394/src/views/objectBrowser.ts#L216

chrjorgensen: The uppercase issue has been fixed.

worksofliam (approved):
* My lowercase cache works now
* Search/filter still working as expected

Beautiful work @chrjorgensen!

PR #1741, opened 2023-12-21
Fix CRTLIB error checking when connecting + other error checkings

sebjulliand:
### Changes
Fixes https://github.com/codefori/vscode-ibmi/issues/1739

This PR adds a method that will parse commands `stderr` into `IBMiMessages` to facilitate error analysis.
This helps checking what IBM i messages are returned after a command call and fixes some checking that only looked for the first returned message ID.

Also fixes a very weird bug that the core team have only been able to recreate on pub400:

```
qsh: 001-0019 Error found searching for command [[. No such path or directory.
```

This new test would fail without the one line change I made in CompileTools#runCommand.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/CompileTools.ts
+++ b/src/api/CompileTools.ts
@@ -660,7 +660,7 @@ export namespace CompileTools {
               command: [
                 ...options.noLibList? [] : buildLiblistCommands(connection, ileSetup),
                 ...commands.map(command =>
-                  `${`system ${GlobalConfiguration.get(`logCompileOutput`) ? `` : `-s`} "${command.replace(/[$]/g, `\\$&`)}"; if [[ $? -ne 0 ]]; then exit 1; fi`}`,
+                  `${`system ${GlobalConfiguration.get(`logCompileOutput`) ? `` : `-s`} "${command.replace(/[$]/g, `\\$&`)}"; exit $?`}`,
                 )
               ].join(` && `),
               directory: cwd,
--- a/src/api/CompileTools.ts
+++ b/src/api/CompileTools.ts
@@ -660,7 +660,7 @@ export namespace CompileTools {
               command: [
                 ...options.noLibList? [] : buildLiblistCommands(connection, ileSetup),
                 ...commands.map(command =>
-                  `${`system ${GlobalConfiguration.get(`logCompileOutput`) ? `` : `-s`} "${command.replace(/[$]/g, `\\$&`)}"; exit $?`}`,
+                  `${`system ${GlobalConfiguration.get(`logCompileOutput`) ? `` : `-s`} "${command.replace(/[$]/g, `\\$&`)}" && exit $?`}`,
                 )
               ].join(` && `),
               directory: cwd,
--- a/src/api/CompileTools.ts
+++ b/src/api/CompileTools.ts
@@ -660,7 +660,7 @@ export namespace CompileTools {
               command: [
                 ...options.noLibList? [] : buildLiblistCommands(connection, ileSetup),
                 ...commands.map(command =>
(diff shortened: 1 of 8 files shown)


chrjorgensen (changes requested):
@sebjulliand When connecting to PUB400, I get the following error:

image

I checked, and there are no dataarea like that on PUB400...?

sebjulliand:
> @sebjulliand When connecting to PUB400, I get the following error:
> 
> I checked, and there are no dataarea like that on PUB400...?

@chrjorgensen The change I made doesn't impact this check. But it would mean that this call to `CHKOBJ` didn't return 0:
image

Do you have any more detail in debug, like the content of the `QCPTOIMPF` CommandResult?

chrjorgensen:
@sebjulliand It's the previously discussed error with `[[`:

```
"CPF9801:  Object QCPTOIMPF in library QSYS not found.\nqsh: 001-0019 Error found searching for command [[. No such path or directory."
```

chrjorgensen:
@sebjulliand This is the status after the `CRTLIB` command:

image

Should additional tests for message ID's be added? CPF0001?

worksofliam:
@chrjorgensen 

> I actually don't understand how the error code can be zero, when the command failed...!?

I don't think it can. This is likely my bad for sure!

worksofliam:
@chrjorgensen that is totally weird, but I'm guessing it's because of how runCommand returns the exit code. Defo got something to do with those brackets.

This line of code: https://github.com/codefori/vscode-ibmi/blob/7c7dfb42d6cbf7cf7cb43c36072a206ed32dbf33/src/api/CompileTools.ts#L663

worksofliam:
I have merged my PR into this one to make it easier. Now I am no longer getting the warning about the data area existing! 

@chrjorgensen back to you for some testing and I am going to do a quick code review too.

sebjulliand (commented):
Works fine here on PUB400 and other systems.
Looks like we could merge!

worksofliam:
@sebjulliand I agree on the merge, but I am thinking we should let @chrjorgensen have another go through as well.

src/api/IBMi.ts
```diff
@@ -348,47 +348,30 @@ export default class IBMi {
           command: `CRTLIB LIB(` + this.config.tempLibrary + `) TEXT('Code for i temporary objects. May be cleared.')`,
           noLibList: true
         });
+        console.log(createdTempLib?.stderr);
```
worksofliam, line 351: Should this be taken out?

sebjulliand: Agreed

src/api/CompileTools.ts
```diff
@@ -660,7 +660,7 @@ export namespace CompileTools {
               command: [
                 ...options.noLibList? [] : buildLiblistCommands(connection, ileSetup),
                 ...commands.map(command =>
-                  `${`system ${GlobalConfiguration.get(`logCompileOutput`) ? `` : `-s`} "${command.replace(/[$]/g, `\\$&`)}"; exit $?`}`,
+                  `${`system ${GlobalConfiguration.get(`logCompileOutput`) ? `` : `-s`} "${command.replace(/[$]/g, `\\$&`)}" && exit $?`}`,
```
chrjorgensen, line 663: @worksofliam How can this work? AFAIK, `&&` make the `exit` only being run if the previous command runs successfully?

worksofliam: You are completely right. I am taking out the `exit` in a new commit and the tests still pass.

chrjorgensen (approved):
@sebjulliand @worksofliam Working as expected - approved!

PR #1749, opened 2023-12-22
Updated release workflow

sebjulliand:
### Changes
This PR enhances the release workflow by automating the version number change. This allows to make a new release without the need to create a PR to bump the version number in `package.json`.

It does the following:
1. Updates the version to match the release tag's
2. Packages once using vsce (through npx)
3. Publishes the package on the marketplace and openvsx
4. Attaches the vsix as a release asset
5. Commits the version number change, moves the release tags to this commit and pushes it all
6. Updates the version with a new prerelease version, bearing the `-dev` suffix
7. Commits and pushes the version update

It also uses Node 18 for building.

### Checklist
* [x] have tested my change (on another repo)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,7 +12,7 @@ jobs:
       - name: Use Node.js 18
         uses: actions/setup-node@v3
         with:
-        node-version: 18
+          node-version: 18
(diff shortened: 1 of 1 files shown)


worksofliam (approved):
Looks ok to me. One way to find out if it works is by making a release, I guess? :)

PR #1750, opened 2023-12-23
Check if a library exists before listing its content

sebjulliand:
### Changes
This PR fixes the `getObjectList` method that showed a confusing error when trying to list objects from a non existing library:
fail1

Now the method will first check if the library exist before carrying on and show a relevant error message if it doesn't:
fail2

### Testing
- Create filter for a library that doesn't exist
- Expand the filter
- Witness the error messages

### Checklist
* [x] have tested my change


worksofliam (approved):
A very simple change! I like it!

At first I was confused why this doesn't just check the `DSPFD` result, but was reminded that this API does not only call `DSPFD`, but also `DSPOBJD` in another condition.

Thanks!

PR #1751, opened 2023-12-23
Replace mapping with label fixes

worksofliam:
### Changes

While looking at another PR, I noticed we were changing a lot of attributes of members when the user has the 'objects as lowercase' option enabled.

This replaces two maps with a function which corrects the label casing (the original intention).

### Checklist

* [x] have tested my change


sebjulliand (approved):
LGTM!
It's indeed way better to change the case of what's displayed. Love it! 😉

PR #1756, opened 2024-01-03
Show recently opened files :boom: 

chrjorgensen:
### Changes

This PR will show the recently opened files in the Go To file list, with the last opened file first in the list and a button to clear the list of recently opened files:

image

This makes it very easy to reopen a previously edited/browsed file. :boom: 

A configuration setting has also been added to control the number of recently opened files in the list:

image

### Checklist

* [x] have tested my change

### DIFF
--- a/package.json
+++ b/package.json
@@ -467,6 +467,23 @@
 					"default": false,
 					"description": "When lines with ILE diagnostics are edited, the diagnostic will be removed. Requires reload/restart if changed."
 				},
+        "code-for-ibmi.recentlyOpenedFilesLimit": {
+          "type": "integer",
+          "default": 10,
+          "description": "Number of files to keep in recently opened files list.",
+          "enum": [
+            0,
+            10,
+            20,
+            30
+          ],
+          "enumDescriptions": [
+            "Do not keep any recently opened files",
+            "Keep the 10 most recently opened files",
+            "Keep the 20 most recently opened files",
+            "Keep the 30 most recently opened files"
+          ]
+        },
 				"code-for-ibmi.safeDeleteMode": {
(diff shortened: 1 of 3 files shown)


sebjulliand (commented):
Great UX enhancement @chrjorgensenm, as always !
Code looks good and it works like a charm. Clearing is OK, and previous lists are saved by connections.

Before I approve, I have a question: shouldn't we limit the number of previously opened entries saved? So far it's unlimited and it's up to the user to clear this list now and then. Would it make sense to set a hard limit or add a new global setting to define this limit?

chrjorgensen:
> Before I approve, I have a question: shouldn't we limit the number of previously opened entries saved? So far it's unlimited and it's up to the user to clear this list now and then. Would it make sense to set a hard limit or add a new global setting to define this limit?

Good question - I haven't thought of that... What is an appropriate limit? 10? 20? 100? Maybe a global settings, as you suggest, so the user can decide - with a limit of 20... Agree?

I also thought of having the timestamp for the file open somewhere (description, tooltip, whatever) - could be helpful for the user...
But I decided that it could be in another PR...

sebjulliand:
I see that VSCode keeps track of the 10 last opened files, so let's say that a global setting that defaults to 10 may be good!
Do as you see fit for the timestamp, it sounds great anyway.

chrjorgensen:
There you go @sebjulliand:

- `Previously opened files` has been changed to `Recently opened files`
- A configuration setting has been added to control the number of recently opened files in the list

sebjulliand (approved):
Works like a charm!
Thank you @chrjorgensen , let's start this week-end with a merge!

PR #1757, opened 2024-01-04
Feature/edit vs browse

sebjulliand:
### Changes
Resolves https://github.com/codefori/vscode-ibmi/issues/1752

As discussed in https://github.com/orgs/codefori/discussions/1331, browsing/editing support could use a little love. This PR brings some enhancement to address what was discussed:

- A new `Open` sub-menu replaces the single `Browse` action in stream files/member's right click menu. The sub-menu offers to explicitly open a streamfile/member read-only or for editing.
image
- A new global setting lets the user define what should be the open action by default when clicking on a streamfile/member to open it
image
- A new connection settings lets the user define `protected paths`: a comma-separated list of libraries and/or IFS paths that will always be opened read-only.
image

Under the hood, all the actions that open a streamfile/member now end up using `code-for-ibmi.openEditable` that holds all the logic for checking if the target should be opened read-only or not. This command has also be enhanced to allow to open a resource and highlight a range in the opened document.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/webviews/settings/index.ts
+++ b/src/webviews/settings/index.ts
@@ -1,10 +1,11 @@
 import vscode from "vscode";
 import { ConnectionConfiguration, GlobalConfiguration } from "../../api/Configuration";
 import { ComplexTab, CustomUI, Section } from "../../api/CustomUI";
+import { Tools } from "../../api/Tools";
+import { isManaged } from "../../api/debug";
 import * as certificates from "../../api/debug/certificates";
 import { instance } from "../../instantiate";
 import { ConnectionData, Server } from '../../typings';
-import { isManaged } from "../../api/debug";
 
 const ENCODINGS = [`37`, `256`, `273`, `277`, `278`, `280`, `284`, `285`, `297`, `500`, `871`, `870`, `905`, `880`, `420`, `875`, `424`, `1026`, `290`, `win37`, `win256`, `win273`, `win277`, `win278`, `win280`, `win284`, `win285`, `win297`, `win500`, `win871`, `win870`, `win905`, `win880`, `win420`, `win875`, `win424`, `win1026`];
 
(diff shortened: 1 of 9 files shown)


src/webviews/settings/index.ts
```diff
@@ -230,7 +229,11 @@ export class SettingsUI {
                     if (data[key].trim() === ``) data[key] = null;
                     break;
                   case `hideCompileErrors`:
-                    data[key] = data[key].split(`,`).map((item: string) => item.trim().toUpperCase()).filter((item: string) => item !== ``);
+                  case `protectedPaths`:
+                    data[key] = data[key].split(`,`)
+                      .map((item: string) => item.trim().toUpperCase())
```
chrjorgensen, line 234: @sebjulliand You're uppercasing every item in the list, including IFS paths... this makes IFS paths never match and protect IFS directories (unless their name is in CAPITALS)...

Only library names should be transformed to uppercase...

chrjorgensen (changes requested):
@sebjulliand Great PR - only a few issues:

- directory uppercase (see code comment)
- filter does not show the library is protected - only when the filter is marked as protected. Could this be done?

src/webviews/settings/index.ts
sebjulliand: Silly me...

chrjorgensen: Not as silly as me, who made PR #1756 without checking if it would collide with other PR's - like this! 😄 

Seems like I have to wait for this one and then remap my changes into the new version...

chrjorgensen: Oh, actually - maybe my PR came first anyway....! 😉

sebjulliand: Yup, we'll merge yours first and then I'll take care of this one. That's only fair!

sebjulliand:
There you go @chrjorgensen :
- Only libraries will be uppercase when saving the protected paths setting
- Both IFS browser and Object browser will take into account the protected paths
- IFS browser will show a lock on folders that are protected (or parent of a protected folder)
- Changing the protected paths setting will automatically refresh the browsers to immediately protect what needs to be

chrjorgensen:
@sebjulliand Looking very good  and promising! 😍 

I will do some more testing tomorrow - before I'll approve this great enhancement.

chrjorgensen (approved):
@sebjulliand All is fine - as always with your PR's! :heart_eyes: 

Rebase and merge when appropriate...

sebjulliand:
> @sebjulliand All is fine - as always with your PR's! 😍
> 
> Rebase and merge when appropriate...

Will do! Thank you so much!

PR #1760, opened 2024-01-05
Fix error when referencing undefined parameters variable

krethan:
When parameters variable is undefined it is generating an error when trying to reference parameters.name. Added check to ensure parameters.name is defined before updating the connection information. There is also no need to update the connection information if parameters.name is undefined.

### DIFF
--- a/src/api/Configuration.ts
+++ b/src/api/Configuration.ts
@@ -144,9 +144,12 @@ export namespace ConnectionConfiguration {
   }
 
   export async function update(parameters: Parameters) {
-    let connections = getConnectionSettings();
-    connections.filter(conn => conn.name === parameters.name).forEach(conn => Object.assign(conn, parameters));
-    await updateAll(connections);
+    let connections;
+    if(parameters?.name) {
+      connections = getConnectionSettings();
+      connections.filter(conn => conn.name === parameters.name).forEach(conn => Object.assign(conn, parameters));
+      await updateAll(connections);
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


krethan:
2 commits as I updated the contributions.md seperately.

chrjorgensen:
Hi @krethan !

Thank you for your contribution to this project - really appreciated!

One question: How can I reproduce the error, that your change is fixing? I haven't seen this before...

worksofliam:
@chrjorgensen I believe this shows up in the debug console (during debug) or the developer consider (during runtime). I have seen this!

krethan:
When you try to open a connection parameters will be undefined and gives this error in the debug console: 

image

src/api/Configuration.ts
```diff
@@ -144,9 +144,12 @@ export namespace ConnectionConfiguration {
   }
 
   export async function update(parameters: Parameters) {
-    let connections = getConnectionSettings();
-    connections.filter(conn => conn.name === parameters.name).forEach(conn => Object.assign(conn, parameters));
-    await updateAll(connections);
+    let connections;
+    if(parameters?.name) {
+      connections = getConnectionSettings();
```
worksofliam, line 149: Any chance this can be `const connections =` instead of using `let` above?

worksofliam (changes requested):
@krethan I left one small request - otherwise, working as expected. Thanks!

krethan:
@worksofliam Changed the let to const and moved it into the if block. 

This issue is occuring when setting up the connection for the first time after the extension is installed. After a connection is successful this is not an issue anymore.

worksofliam:
Just retested and works as expected. This error had been lingering for a while, so it's nice to see it fixed. 

Thanks for your PR!

worksofliam (approved):

PR #1767, opened 2024-01-08
Delete password when private key is used

worksofliam:
### Changes

There is a bug with starting the debugger if the user once used a password, but is now using a private key, as the user would never be prompted for the correct password when debugging.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/webviews/settings/index.ts
+++ b/src/webviews/settings/index.ts
@@ -303,6 +303,7 @@ export class SettingsUI {
               if (!data.privateKeyPath?.trim()) {
                 if (connection.privateKeyPath?.trim()) {
                   data.privateKeyPath = connection.privateKeyPath;
+                  context.secrets.delete(`${name}_password`);
                 }
                 else {
                   delete data.privateKeyPath;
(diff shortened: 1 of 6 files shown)


edmundreinhardt (approved):

src/webviews/settings/index.ts
```diff
@@ -303,6 +303,7 @@ export class SettingsUI {
               if (!data.privateKeyPath?.trim()) {
                 if (connection.privateKeyPath?.trim()) {
                   data.privateKeyPath = connection.privateKeyPath;
+                  context.secrets.delete(`${name}_password`);
```
sebjulliand, line 306: It's an async call, we should await.
```suggestion
                  await context.secrets.delete(`${name}_password`);
```

sebjulliand (changes requested):
It works as expected! Good job!
But `context.secrets` function are asynchronous, I'd suggest we `await` on them everywhere in the extension (here, but there are other places in the extension where `await` is missing too).

worksofliam:
@sebjulliand Ready for another review - thanks!

sebjulliand (approved):
Looking good now!
I added a commit to add `await` in some other places where the secret store is used, I hope you won't mind 😁
Feel free to merge whenever you want.

PR #1768, opened 2024-01-08
Fix path splitting when opening a member on an IASP

sebjulliand:
### Changes
Fixes https://github.com/codefori/vscode-ibmi/issues/1765

The path parsing was incorrect when opening a member located on an iASP. Since this kind of path begins with the iASP, it would return the iASP name as being the library.

This PR adds a method in `Tools` that will correctly parse the path in any case.

### Checklist
* [x] have tested my change


worksofliam (approved):
@sebjulliand works for me. My test was adding and removing items from the protected paths list and ensuring the item could be opened and editable when applicable.

PR #1780, opened 2024-01-14
Enhance Object Browser filter's Object field

sebjulliand:
### Changes
Resolves https://github.com/codefori/vscode-ibmi/issues/1779

This PR enhances the Object browser filters by adding support for more advanced filtering options.

#### Multi level filtering
Every field of a filter now supports filtering:
- Libraries
- Objects
- Object types
- Members
- Member types

#### Mixed browsing
Multiple libraries can now be browsed with a single filter.
image

When source files are found in the list (if `Object Types` list includes `*SRCPF` or `*ALL`), they can be browsed as well - they are not treated as simple Objects anymore:
image

#### Filtering type
The filtering type is defined for every field of the filter.
image

- **Simple**: each filed will support multiple comma-separated values that can contain multiple `*` as wildcard. Example: `QSYS*, *TEST, *SAMPLE*, A*B*C, YAY`
- **Regex**: each field will support a single RegEx for filtering values

#### API: new `getQTempTable` content method
This method allows to prepare a file in QTEMP using multiple SQL statements and get the result. This allows for not relying on temporary tables written in the temporary library defined in the configuration.

`content.getTable` has been replaced by `content.getQTempTable` where relevant.

#### Tests
Additional unit tests have been provided to test the filter engine and the library/object/member methods

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -475,7 +475,7 @@ export default class IBMiContent {
   }
 
   async getLibraries(filters: { library: string; filterType?: FilterType }) {
-    return this.getObjectList({ library: "QSYS", object: filters.library, types: ["*LIB"] });
+    return this.getObjectList({ library: "QSYS", object: filters.library, types: ["*LIB"], filterType:filters.filterType });
   }
 
   /**
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -447,7 +447,24 @@ export default class IBMiContent {
       throw new Error(`Library ${library} does not exist.`);
     }
 
-    const object = (filters.object && filters.object !== `*` ? filters.object.toUpperCase() : `*ALL`);
+    const objects = filters.object?.split(',')
+      .map(o => o.trim().toLocaleUpperCase())
+      .filter(Tools.distinct)
+      .map(pattern => {
+        if (pattern.startsWith('*')) {
+          return (name: string) => name.endsWith(pattern.substring(1));
+        }
+        else if (pattern.endsWith('*')) {
+          return (name: string) => name.startsWith(pattern.substring(0, pattern.length - 1));
+        }
+        else {
+          return (name: string) => name === pattern;
+        }
+      })
+      || [];
+    const filterNames = objects.length > 1 ? (file: IBMiFile) => objects.some(test => test(file.name)) : undefined;
+
+    const object = filters.object && !filterNames && filters.object !== `*` ? filters.object.toUpperCase() : `*ALL`;
     const sourceFilesOnly = (filters.types && filters.types.includes(`*SRCPF`));
 
     const tempLib = this.config.tempLibrary;
@@ -473,6 +490,7 @@ export default class IBMiContent {
           text: String(object.PHTXT),
           count: Number(object.PHNOMB),
         } as IBMiFile))
+        .filter(object => !filterNames || filterNames(object))
         .sort((a, b) => a.library.localeCompare(b.library) || a.name.localeCompare(b.name));
     } else {
       const objectTypes = (filters.types && filters.types.length ? filters.types.map(type => type.toUpperCase()).join(` `) : `*ALL`);
@@ -494,6 +512,7 @@ export default class IBMiContent {
         attribute: String(object.ODOBAT),
         text: String(object.ODOBTX)
       } as IBMiFile))
+        .filter(object => !filterNames || filterNames(object))
         .sort((a, b) => {
           if (a.library.localeCompare(b.library) != 0) {
             return a.library.localeCompare(b.library)
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -447,9 +447,11 @@ export default class IBMiContent {
       throw new Error(`Library ${library} does not exist.`);
     }
 
-    const objects = filters.object?.split(',')      
+    let hasEndsWith = false;
+    const objects = filters.object?.split(',')
       .map(pattern => {
         if (pattern.startsWith('*')) {
+          hasEndsWith = true;
           return (name: string) => name.endsWith(pattern.substring(1));
         }
         else if (pattern.endsWith('*')) {
@@ -460,7 +462,8 @@ export default class IBMiContent {
         }
       })
       || [];
-    const filterNames = objects.length > 1 ? (file: IBMiFile) => objects.some(test => test(file.name)) : undefined;
+
+    const filterNames = objects.length && (objects.length > 1 || hasEndsWith) ? (file: IBMiFile) => objects.some(test => test(file.name)) : undefined;
 
     const object = filters.object && !filterNames && filters.object !== `*` ? filters.object.toUpperCase() : `*ALL`;
     const sourceFilesOnly = (filters.types && filters.types.includes(`*SRCPF`));
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -282,9 +282,9 @@ export default class IBMiContent {
 
   /**
    * Download the contents of a table.
-   * @param library 
-   * @param file 
-   * @param member Will default to file provided 
+   * @param library
+   * @param file
+   * @param member Will default to file provided
    * @param deleteTable Will delete the table after download
    */
   async getTable(library: string, file: string, member: string, deleteTable?: boolean): Promise<Tools.DB2Row[]> {
@@ -404,7 +404,7 @@ export default class IBMiContent {
       `;
       results = await this.runSQL(statement);
     } else {
-      results = await this.getQTempTable([`CALL QSYS2.QCMDEXC('DSPOBJD OBJ(QSYS/*ALL) OBJTYPE(*LIB) DETAIL(*TEXTATR) OUTPUT(*OUTFILE) OUTFILE(QTEMP/LIBLIST)`], "LIBLIST");
+      results = await this.getQTempTable([`CALL QSYS2.QCMDEXC('DSPOBJD OBJ(QSYS/*ALL) OBJTYPE(*LIB) DETAIL(*TEXTATR) OUTPUT(*OUTFILE) OUTFILE(QTEMP/LIBLIST)')`], "LIBLIST");
       if (results.length === 1 && !results[0].ODOBNM?.toString().trim()) {
         return [];
       }
@@ -479,9 +479,9 @@ export default class IBMiContent {
   }
 
   /**
-   * @param filters 
+   * @param filters
    * @param sortOrder
-   * @returns an array of IBMiFile 
+   * @returns an array of IBMiFile
    */
   async getObjectList(filters: { library: string; object?: string; types?: string[]; filterType?: FilterType }, sortOrder?: SortOrder): Promise<IBMiObject[]> {
     const library = filters.library.toUpperCase();
@@ -568,9 +568,9 @@ export default class IBMiContent {
   }
 
   /**
-   * 
+   *
    * @param filter: the criterias used to list the members
-   * @returns 
+   * @returns
    */
   async getMemberList(filter: { library: string, sourceFile: string, members?: string, extensions?: string, sort?: SortOptions, filterType?: FilterType }): Promise<IBMiMember[]> {
     const sort = filter.sort || { order: 'name' };
@@ -579,7 +579,7 @@ export default class IBMiContent {
 
     const memberFilter = parseFilter(filter.members, filter.filterType);
     const singleMember = memberFilter.noFilter && filter.members && !filter.members.includes(",") ?  filter.members.toLocaleUpperCase().replace(/[*]/g, `%`) : undefined;
-    
+
     const memberExtensionFilter = parseFilter(filter.extensions, filter.filterType);
     const singleMemberExtension = memberExtensionFilter.noFilter && filter.extensions && !filter.extensions.includes(",") ?  filter.extensions.toLocaleUpperCase().replace(/[*]/g, `%`) : undefined;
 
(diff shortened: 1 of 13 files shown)


chrjorgensen (changes requested):
@sebjulliand As always a great enhancement from your hands! :heart_eyes: 

I have some comments besides the suggestions in the code:
- Previously `*ALL` was allowed as object name and returned all objects (like a single `*`). This does not work anymore. I guess that is okay, since the description says a single `*` will do the trick.
- When I specify a generic object name like `*C*` in the filter, nothing is returned. I guess your generic-test does not handle this? Maybe a regex test could solve this?

You know what's coming next... **_Can we have the same for members?!_** :rofl: 
Seriously, I'm convinced the users will ask for this - now they're given the option to have multiple objects in a filter.

We're still a bit behind RDi ("shh, don't mention the war!" :laughing:) in our filter options - in RDi you can specify a generic source file name linked to a generic member name and type, and have multiple of these groups in a filter. When I was thinking of coding this, that was my intention - to turn the object, member and type into a list, so the three would be linked together and had all to be met to be included in the output. This allows for more granular selection...

WDYT?

But still, a great enhancement...!!! :heart_eyes:

sebjulliand:
> @sebjulliand As always a great enhancement from your hands! 😍
> 
> I have some comments besides the suggestions in the code:
> ...
> WDYT?
> 
> But still, a great enhancement...!!! 😍

Thank you so much @chrjorgensen ! As always, your review is on point.
I'll look into the suggested changes and add support for *xxx* type of filter.

And then I'll apply the filtering logic to members and types; it should be fairly easy.

src/api/IBMiContent.ts
```diff
@@ -447,7 +447,25 @@ export default class IBMiContent {
       throw new Error(`Library ${library} does not exist.`);
     }
 
-    const object = (filters.object && filters.object !== `*` ? filters.object.toUpperCase() : `*ALL`);
+    let hasEndsWith = false;
```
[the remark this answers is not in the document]
sebjulliand: It's used right after to check if at least one filter is a `ends with` kind of filter. In this case, even if there is one filter, the filtering logic (i.e. AFTER all the items are loaded) must be applied.

src/api/IBMiContent.ts
```diff
@@ -447,7 +447,25 @@ export default class IBMiContent {
...
+    const objects = filters.object?.split(',')
```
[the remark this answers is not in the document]
sebjulliand: Yep, I'm going to add this capability. Thanks for the suggestion!

worksofliam:
@sebjulliand Since you are enhancing `getObjectList`, I would really love to see some additional tests to support these changes

jkyeung:
First, thank you for this enhancement. It truly transforms the usability of the filter.

Second, I want to amplify the point about the members: If I could only choose to have "advanced" filtering on objects or members (I realize this is not a true choice, just trying to illustrate), I would choose members. In any shop that uses source physical files, the members are where all the most important naming conventions happen. Everyone mainly sticks to the object names provided by IBM (QRPGLESRC, QCLSRC, etc.), and there are not too many of them. But there are usually hundreds or even thousands of members in a given source file, and every shop has their own naming system for them.

Finally, to that end, I think a regex option would be extremely useful, but only for those users who are comfortable with regex. It would be confusing to many users. So if it's too much trouble to handle multiple styles of wildcard, I would stick to the basics: either just something equivalent to regex's `.*` (like the current `*` most IBM i users are used to, or `%` in SQL's LIKE predicate), or that plus something equivalent to regex's `.` (like SQL's `_`). Those would cover most cases for most users.

I could imagine a setting, where you can choose regex or "basic IBM" or "SQL-style", especially if you're going to implement any of them using regex under the covers anyway.

sebjulliand:
Thanks for the feed @jkyeung !
I'm actually going to add support for regex in there, at every level (library, file and member) - we should end up with the same filtering ability as in RDi, but with regex support.

By default, filters will only support the `*` wildcard to keep it simple for most use cases and not confuse people. A checkbox in the filter definition will allow to turn on RegEx support if needed.

jkyeung:
> I'm actually going to add support for regex in there, at every level (library, file and member)

I can't wait!

> By default, filters will only support the `*` wildcard to keep it simple for most use cases and not confuse people. A checkbox in the filter definition will allow to turn on RegEx support if needed.

Perfect!

sebjulliand:
@codefori/core , you're on!
I updated the PR description to match what's done here. Let me know how it goes! 😄

src/api/IBMiContent.ts
```diff
@@ -360,16 +404,10 @@ export default class IBMiContent {
       `;
       results = await this.runSQL(statement);
     } else {
-      await this.ibmi.runCommand({
-        command: `DSPOBJD OBJ(QSYS/*ALL) OBJTYPE(*LIB) DETAIL(*TEXTATR) OUTPUT(*OUTFILE) OUTFILE(${tempLib}/${TempName})`,
-        noLibList: true
-      });
-      results = await this.getTable(tempLib, TempName, TempName, true);
-
+      results = await this.getQTempTable([`CALL QSYS2.QCMDEXC('DSPOBJD OBJ(QSYS/*ALL) OBJTYPE(*LIB) DETAIL(*TEXTATR) OUTPUT(*OUTFILE) OUTFILE(QTEMP/LIBLIST)`], "LIBLIST");
```
chrjorgensen, line 407: The SQL statement is not correct:
```suggestion
      results = await this.getQTempTable([`CALL QSYS2.QCMDEXC('DSPOBJD OBJ(QSYS/*ALL) OBJTYPE(*LIB) DETAIL(*TEXTATR) OUTPUT(*OUTFILE) OUTFILE(QTEMP/LIBLIST)')`], "LIBLIST");
```

chrjorgensen, line 407: I think your command is missing a quote and parenthesis:

```suggestion
      results = await this.getQTempTable([`CALL QSYS2.QCMDEXC('DSPOBJD OBJ(QSYS/*ALL) OBJTYPE(*LIB) DETAIL(*TEXTATR) OUTPUT(*OUTFILE) OUTFILE(QTEMP/LIBLIST)')`], "LIBLIST");
```

src/api/IBMiContent.ts
```diff
@@ -436,192 +474,168 @@ export default class IBMiContent {
...
-        if (memberExt) {
-          patternExt = new RegExp(`^` + memberExt.replace(/[*]/g, `.*`).replace(/[$]/g, `\\$`) + `$`);
-        }
+      )
+      Select * From MEMBERS
+      Where LIBRARY = '${library}'
+        ${sourceFile !== `*ALL` ? `And SOURCE_FILE = '${sourceFile}'` : ``}
+        ${singleMember ? `And NAME Like '${singleMember}'` : ''}
+        ${singleMemberExtension ? `And TYPE Like '${singleMemberExtension}'` : ''}
+      Order By ${sort.order === 'name' ? 'NAME' : 'CHANGED'} ${!sort.ascending ? 'DESC' : 'ASC'}`;
```
chrjorgensen, line 610: Why do you have a common table expression here? There are no WHERE clause in the CTE and it will become the product of _ALL_ files and members on the system.

I would suggest simply removing it:

```suggestion
    const statement =
      `SELECT
          rtrim(cast(a.system_table_schema as char(10) for bit data)) as LIBRARY,
          b.avgrowsize as RECORD_LENGTH,
          a.iasp_number as ASP,
          rtrim(cast(a.system_table_name as char(10) for bit data)) AS SOURCE_FILE,
          rtrim(cast(b.system_table_member as char(10) for bit data)) as NAME,
          coalesce(rtrim(cast(b.source_type as varchar(10) for bit data)), '') as TYPE,
          coalesce(rtrim(b.partition_text), '') as TEXT,
          b.NUMBER_ROWS as LINES,
          extract(epoch from (b.CREATE_TIMESTAMP))*1000 as CREATED,
          extract(epoch from (b.LAST_SOURCE_UPDATE_TIMESTAMP))*1000 as CHANGED
        FROM qsys2.systables AS a
          JOIN qsys2.syspartitionstat AS b
            ON b.table_schema = a.table_schema AND
              b.table_name = a.table_name
      Where LIBRARY = '${library}'
        ${sourceFile !== `*ALL` ? `And SOURCE_FILE = '${sourceFile}'` : ``}
        ${singleMember ? `And NAME Like '${singleMember}'` : ''}
        ${singleMemberExtension ? `And TYPE Like '${singleMemberExtension}'` : ''}
      Order By ${sort.order === 'name' ? 'NAME' : 'CHANGED'} ${!sort.ascending ? 'DESC' : 'ASC'}`;
```

src/api/IBMiContent.ts
```diff
@@ -436,192 +474,168 @@ export default class IBMiContent {
...
-    let sorter: (r1: IBMiMember, r2: IBMiMember) => number;
-    if (sort.order === 'name') {
-      sorter = (r1, r2) => r1.name.localeCompare(r2.name);
+    let results: Tools.DB2Row[];
+    if (this.config.enableSQL) {
+      results = await this.runSQL(statement);
     }
     else {
-      sorter = (r1, r2) => r1.changed!.valueOf() - r2.changed!.valueOf();
+      results = await this.getQTempTable([`Create Table QTEMP.MEMBERSLST As (${statement}) With DATA`], "MEMBERSLST");
```
chrjorgensen, line 616: You have removed the `DSPFD` command completely and use SQL always, whether SQL is enabled or not.
Is this correct behavior? I've never really understood the purpose of having a toggle for using SQL or not... @worksofliam?

worksofliam: I guess the original reasons for having the SQL toggle was in case users didn't have `db2util` or `db2` in QSH, but since we now use a system API, which I guess we can assume is always there, then maybe we can remove it one day?

chrjorgensen: One day, hopefully - would make things easier...

src/api/IBMiContent.ts
chrjorgensen: I get it now... finally... 😆

src/api/IBMiContent.ts
chrjorgensen: The command is executed when SQL is not enabled and library descriptions are enabled, and results in this error and an empty library list view:

image

chrjorgensen: Fixed in commit [8887227](https://github.com/codefori/vscode-ibmi/pull/1780/commits/8887227ca3bb471be72cacbd33a31a148cc08ebf)

chrjorgensen:
@sebjulliand 🚀 🚀 🚀 🚀 🚀 - 5 out of 6 rockets because

- so far in my tests the simple filters work perfectly
- only the regex filter won't work for me - do you have an example?

Beautiful work, (IBM) Champ!! 😍

sebjulliand:
@chrjorgensen it runs better now! I fixed what needed to be and RegExp filtering is fine now 😄

chrjorgensen (changes requested):
@sebjulliand Regexp works much better now! 😍 

One issue was found during testing: the regular express meta characters like `\d` (for digits) doesn't work, because the filter is changed to uppercase and the `\d` becomes `\D` - the reverse!

Can this be fixed? Maybe by not converting regex filter values to uppercase and filter with ignore-case flag? Only if it's easy - I could use `[0-9]` to achieve the same result. We could also check for any meta characters and send an error message, saying they are not allowed?

Finally only a comment about this implementation and the one in THAT PRODUCT:

- In THAT PRODUCT the filter can have multiple values, where each value selects generic library AND generic object AND generic member. The multiple values are OR'ed. This is great for a project, where the filter can select e.g. library ABC*, all source files, member DEF* combined with library QWE*, all sourcefiles, member RTY*.

- In our implementation we have one value, selecting generic library AND generic object AND generic member. This does not allow filtering only some members in library A and other members in library B.

To solve this, we would have to allow _a list_ of filter values (libraries, objects, objects types, member names and member types) and have the output be the combined selection of the filters in the list. How easy would that be? Not easy, I would guess - and this is NOT a challenge! 😉

sebjulliand:
@chrjorgensen I see your point. THAT PRODUCT let's you define filter strings, for members, and OR them.
In our case, we define one _filter string_ per filter.

I guess we could achieve the same result by associating multiple filters, as you said. We would need to list all the members from each filter and display them all at once (I just checked and THAT PRODUCT display them all at the same level).

We would need a different kind of filter; an association of filters, as you said. This should be done in a separate PR if we feel like it's needed. This one is already big enough 😄

chrjorgensen:
@sebjulliand 

> This should be done in a separate PR if we feel like it's needed. This one is already big enough 😄

I agree completely - I was just anticipating the reaction from our users who also have experience with THAT PRODUCT... 😆

chrjorgensen:
@sebjulliand I'm happy with this PR - you raise the bar once again! 😍 

🚀 🚀 🚀 🚀 🚀 🚀  - 6 out of 6! 😉 😆 

@worksofliam @Wright4i Please give this great PR a spin, so we can get it merged.

src/testing/index.ts
```diff
@@ -1,21 +1,17 @@
...
+  //ActionSuite,
+  //ConnectionSuite,
   ContentSuite,
-  DeployToolsSuite,
-  ToolsSuite,
-  ILEErrorSuite
+  //DeployToolsSuite,
+  //ToolsSuite,
+  //ILEErrorSuite,
+  //FilterSuite
```
worksofliam, line 14: Accidently commit by the looks! I do this too when testing!!! Haha.

worksofliam (approved):
Code looks good, my user testing has been good, and the test cases are passing.

Thank you @sebjulliand!

PR #1788, opened 2024-01-18
Feature/pseudo source

### LINKED ISSUE #1387: Move CL runner logic to new extension
And also add a way to run an enter CL file without compiling - e.g. one command at a time as requested by @edmundreinhardt

### LINKED ISSUE #1785: Support for pseudo source types
A feature idea came in on Ryver to allow some extensions to be seen as pseudo source.

For example, perhaps we have a member/streamfile/local with extension: `dtaara` and inside of it is a set of commands to create and change a data area:

```cl
CRTDTAARA DTAARA(...) ...
CHGDTAARA ....
```

When the 'Run Action' command is run on a pseudo source type, a new option would show up, perhaps something like 'Execute Pseudo Source' which would run each command top to bottom. 

We kind of already do this with the `code-for-ibmi.cl.runSelected` command.

worksofliam:
### Changes

Adds a new default Action called 'Run as pseudo source', which runs the individual CL commands from the source code and prints info the existing diagnostic instance to show the results.

image

Only applies to sources where the extension matches the list as defined in the new user settings.

image

* Closes #1785
* Closes #1387 as we now use [and need] the built in CL logic in the base extension.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/CompileTools.ts
+++ b/src/api/CompileTools.ts
@@ -115,10 +115,10 @@ export namespace CompileTools {
           };
         });
 
-        const pseudoExtensions = config.psuedoSourceExtensions.map(ext => ext.toUpperCase());
+        const pseudoExtensions = config.pseudoSourceExtensions.map(ext => ext.toUpperCase());
         allActions.push({
-          name: `⚒️ Run as psuedo source`,
-          type: `psuedo`,
+          name: `⚒️ Run as pseudo source`,
+          type: `pseudo`,
           command: '',
           environment: 'ile',
           extensions: pseudoExtensions,
@@ -126,7 +126,7 @@ export namespace CompileTools {
         })
 
         // Then we get all the available Actions for the current context
-        availableActions = allActions.filter(action => (action.type === uri.scheme || action.type === `psuedo`) && (!action.extensions || action.extensions.includes(extension) || action.extensions.includes(fragment) || action.extensions.includes(`GLOBAL`)) && (!isProtected || action.runOnProtected))
+        availableActions = allActions.filter(action => (action.type === uri.scheme || action.type === `pseudo`) && (!action.extensions || action.extensions.includes(extension) || action.extensions.includes(fragment) || action.extensions.includes(`GLOBAL`)) && (!isProtected || action.runOnProtected))
           .sort((a, b) => (actionUsed.get(b.name) || 0) - (actionUsed.get(a.name) || 0))
           .map(action => ({
             label: action.name,
@@ -172,8 +172,8 @@ export namespace CompileTools {
             fromWorkspace = vscode.workspace.workspaceFolders[workspaceId || 0];
           }
 
-          if (chosenAction.type === `psuedo`) {
-            const pseudoResult = await runAsPsuedoSource(instance, uri);
+          if (chosenAction.type === `pseudo`) {
+            const pseudoResult = await runAsPseudoSource(instance, uri);
             return pseudoResult || false;
           }
 
@@ -781,7 +781,7 @@ export namespace CompileTools {
     return command;
   }
 
-  export async function runAsPsuedoSource(instance: Instance, uri: vscode.Uri) {
+  export async function runAsPseudoSource(instance: Instance, uri: vscode.Uri) {
     const connection = instance.getConnection();
     const config = instance.getConfig();
     if (config && connection) {
@@ -794,12 +794,12 @@ export namespace CompileTools {
       const commands = getCommandStrings(document);
 
       const result = await vscode.window.withProgress({
-        title: `Executing psuedo source`, 
+        title: `Executing pseudo source`, 
         location: vscode.ProgressLocation.Notification,
         cancellable: true
       }, async (progress, token) => {
         token.onCancellationRequested(() => {
-          vscode.window.showInformationMessage(`Psuedo source execution cancelled.`);
+          vscode.window.showInformationMessage(`Pseudo source execution cancelled.`);
         });
 
         let overallSuccess = true;
@@ -821,7 +821,7 @@ export namespace CompileTools {
 
           progress.report({ message: commandString });
           const commandResult = await runCommand(instance, {
-            title: `Psuedo source`,
+            title: `Pseudo source`,
             environment: `ile`,
             command: commandString,
           });
--- a/src/api/CompileTools.ts
+++ b/src/api/CompileTools.ts
@@ -124,17 +124,19 @@ export namespace CompileTools {
           }));
       }
 
-      const pseudoExtensions = [`clle`, `dtaara`];
-      if (pseudoExtensions.includes(extension.toLowerCase())) {
-        availableActions.push({
-          label: `Run as psuedo source`,
-          action: {
-            name: `Run as psuedo source`,
-            type: `psuedo`,
-            command: '',
-            environment: 'ile'
-          }
-        })
+      if (!isProtected) {
+        const pseudoExtensions = [`clle`, `dtaara`];
+        if (pseudoExtensions.includes(extension.toLowerCase())) {
+          availableActions.push({
+            label: `Run as psuedo source`,
+            action: {
+              name: `Run as psuedo source`,
+              type: `psuedo`,
+              command: '',
+              environment: 'ile'
+            }
+          })
+        }
       }
 
       if (customAction || availableActions.length) {
@@ -822,7 +824,7 @@ export namespace CompileTools {
           const variables = getDefaultVariables(instance, ileSetup);
           const commandString = replaceValues(command.content, variables);
 
-          progress.report({ message: `Executing ${commandString}` });
+          progress.report({ message: commandString });
           const commandResult = await runCommand(instance, {
             title: `Psuedo source`,
             environment: `ile`,
@@ -836,7 +838,7 @@ export namespace CompileTools {
           if (command.range) {
             diags.push({
               success: commandResult.code === 0,
-              message: commandResult.stderr,
+              message: commandResult.stderr || `No output.`,
               range: new Range(command.range.start, 0, command.range.end, document.lineAt(command.range.end).range.end.character)
             })
           }
--- a/src/api/CompileTools.ts
+++ b/src/api/CompileTools.ts
@@ -115,28 +115,23 @@ export namespace CompileTools {
           };
         });
 
+        const pseudoExtensions = [`CLLE`, `DTAARA`];
+        allActions.push({
+          name: `⚒️ Run as psuedo source`,
+          type: `psuedo`,
+          command: '',
+          environment: 'ile',
+          extensions: pseudoExtensions,
+          runOnProtected: false
+        })
+
         // Then we get all the available Actions for the current context
-        availableActions = allActions.filter(action => action.type === uri.scheme && (!action.extensions || action.extensions.includes(extension) || action.extensions.includes(fragment) || action.extensions.includes(`GLOBAL`)) && (!isProtected || action.runOnProtected))
+        availableActions = allActions.filter(action => (action.type === uri.scheme || action.type === `psuedo`) && (!action.extensions || action.extensions.includes(extension) || action.extensions.includes(fragment) || action.extensions.includes(`GLOBAL`)) && (!isProtected || action.runOnProtected))
           .sort((a, b) => (actionUsed.get(b.name) || 0) - (actionUsed.get(a.name) || 0))
           .map(action => ({
             label: action.name,
             action
           }));
-      }
-
-      if (!isProtected) {
-        const pseudoExtensions = [`clle`, `dtaara`];
-        if (pseudoExtensions.includes(extension.toLowerCase())) {
-          availableActions.push({
-            label: `Run as psuedo source`,
-            action: {
-              name: `Run as psuedo source`,
-              type: `psuedo`,
-              command: '',
-              environment: 'ile'
-            }
-          })
-        }
       }
 
       if (customAction || availableActions.length) {
--- a/src/api/CompileTools.ts
+++ b/src/api/CompileTools.ts
@@ -1,6 +1,6 @@
 
 import path from 'path';
-import vscode, { CustomExecution, EventEmitter, Pseudoterminal, TaskGroup, TaskRevealKind, WorkspaceFolder, commands, tasks } from 'vscode';
+import vscode, { CustomExecution, EventEmitter, Pseudoterminal, Range, TaskGroup, TaskRevealKind, WorkspaceFolder, commands, tasks } from 'vscode';
 import { parseFSOptions } from '../filesystems/qsys/QSysFs';
 import { Action, BrowserItem, CommandResult, DeploymentMethod, RemoteCommand, StandardIO } from '../typings';
 import { GlobalConfiguration } from './Configuration';
@@ -8,10 +8,11 @@ import { CustomUI } from './CustomUI';
 import IBMi from './IBMi';
 import Instance from './Instance';
 import { Tools } from './Tools';
-import { EvfEventInfo, refreshDiagnosticsFromLocal, refreshDiagnosticsFromServer, registerDiagnostics } from './errors/diagnostics';
+import { EvfEventInfo, PseudoDiagnostic, refreshDiagnosticsFromLocal, refreshDiagnosticsFromServer, registerDiagnostics, setDiagnosticsFromPseudo } from './errors/diagnostics';
 import { getLocalActions } from './local/actions';
 import { DeployTools } from './local/deployTools';
 import { getBranchLibraryName, getEnvConfig } from './local/env';
+import { getCommandStrings } from '../languages/clle/clRunner';
 
 const NEWLINE = `\r\n`;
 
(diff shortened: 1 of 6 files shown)


sebjulliand:
The feature is cool but it raises a few concerns of mine.
- It sounds like something related to the "build" side of things. But build tools won't interpret this kind of source like Code4i does (i.e. library list, variable replacement, etc).
- If it's not related to the build, and the goal is to be able to run multiple CL commands at once, isn't it what actions with multiple commands do in this case?

I'm just trying to figure out the actual use case behind it 😉

chrjorgensen:
At first I did not get this concept - what was "pseudo source"? Then I found [this comment](https://github.com/codefori/vscode-ibmi/issues/1387#issuecomment-1608059803) on the idea issue #1387, which explains the concept of pseudo source and the use case for it.

I definitely think the concept needs a very thorough description in the documentation, since the term and purpose and use is not obvious. If I'm having trouble understanding this, other users may as well.

This concept is not for me... I don't have a use case for it and use other build methods. Mainly the [BUILD command](https://www.scottklement.com/build/) from Scott Klement, which I modified to allow me to use a sub-directory from the full path as the object library. This tool has suited me well through the years. 😃 

I don't know if the majority of our users has the need for this feature - I guess most already have their build procedures in place for object types other than programs, like me, and won't switch to use pseudo sources.

I was skeptical about the idea of pseudo sources on issue #1387, and I still am. Sorry about that, but I want to be honest. 

But for those having a need for pseudo sources, this is a great enhancement, for sure.  👍 
I just wish it was an extension by itself, so it doesn't "pollute" the current tool with additional options very rarely used. As it is now, the feature should at least be configurable, so users not using the feature can disable it and not be presented with a default action on top of the action list.

Again, sorry for being so grumpy - and even close to the weekend! 😆 
Go ahead with this feature, if you feel it brings value to the users... ❤️

worksofliam:
I have asked the original feature requester to leave his feedback to your comments here.

JDubbTX:
I was the original requestor for this enhancement, and it helps to understand my background is a large organization (LUG member) that has used Aldon / Lmi / RDO i for many years, as that is what our shop uses.  I'm also interested in Git, modernization, and Devops, and Ci/Cd, and bringing these concepts to the i.

I have been using VS Code and Liam's extension since it was a thing, and RDi for years before that.  These days I still go to RDi for some things, but prefer VS Code for most things.

Why use pseudo source?  LMi /Aldon has managed non source objects for many years, both without source, and with source.  Over the years our shop decided to start using pseudo source for all non source objects that we manage through Aldon.  You have greater observability, and a better understanding of the objects on your system if you do this.  A specific instance I can recall as an on call developer was spending hours in the middle of the night when the system went down due to a dtaq that had reached its limit.  Not having the pseudosource for the dtaq object meant i had to go to a third party tool (taatool) to see how the object was defined (what is the max size attribute, did it reclaim records, etc).  When you manage non-source objects with pseudo source, you not only have a quick reference to the current object definition, but you also have a historical record of the object, assuming you use some kind of source control.

JDubbTX:
I was the original requestor for this enhancement, and it helps to understand my background is a large organization (LUG member) that has used Aldon / Lmi / RDO i for many years, as that is what our shop uses.  I'm also interested in Git, modernization, and Devops, and Ci/Cd, and bringing these concepts to the i.

I have been using VS Code and Liam's extension since it was a thing, and RDi for years before that.  These days I still go to RDi for some things, but prefer VS Code for most things.

Why use pseudo source?  LMi /Aldon has managed non source objects for many years, both without source, and with pseudo source.  Over the years our shop decided to start using pseudo source for all non source objects that we manage through Aldon.  You have greater observability, and a better understanding of the objects on your system if you do this.  A specific instance I can recall as an on call developer was spending hours in the middle of the night when the system went down due to a dtaq that had reached its limit.  Not having the pseudosource for the dtaq object meant i had to go to a third party tool (taatool) to see how the object was defined (what is the max size attribute, did it reclaim records, etc).  When you manage non-source objects with pseudo source, you not only have a quick reference to the current object definition, but you also have a historical record of the object, assuming you use some kind of source control.

JDubbTX:
Why have pseudo source as a built in action?  Why not just use the extension's cl command function that is already built in?

There are several answers to this.

1) I want to build non-source objects the same way I build source objects.

2) It makes the feature more visible, accessible, and convenient.

3) It encouraged the use of pseudo source, which is good for the reasons outlined above

4) It adds to the already great feature set of the extension, which encourages adoption of the tool.

JDubbTX:
> - It sounds like something related to the "build" side of things. But build tools won't interpret this kind of source like Code4i does (i.e. library list, variable replacement, etc).

 I can think of at least 3 build tools that already handle pseudo source.  Rocket Aldon, Arcad, and Bob.  I'm sure the other big ones do too, though I don't have any first hand knowledge.

chrjorgensen:
@JDubbTX 

Thank you for the detailed description of the background for your request for pseudo source in Code for IBM i. I had already guessed that the request came from a LUG member, so I'm not surprised.

I understand why you like pseudo source for objects that is not compiled from source. It helps in documentation and version control, as you describe. I also understand your desire for having an action to make it more convenient to "build" objects like programs etc.

However, this extension is used by all kind of users - users from LUG members and users in small shops with only one or few developers. The extension is a great and free alternative to the commercial, bigger products like Merlin and RDi, which often are out of reach of the developers in the small shops. So the extension should work optimally for the majority of the users, and special functions geared toward a minority of users should be avoided if possible - especially if the function introduces new elements in the UI, that all users are exposed to, using the function or not.

Since VS Code is made up of extensions, it is my opinion that the pseudo source function should be implemented in its own extension - using Code for IBM i as the API for the IBM i. Then the (few?) users needing this functionality could install the pseudo code extension on top of Code for IBM i, and the (many?) users not interested in this feature could continue with Code for IBM i as it is.

I will go through your arguments:

> I want to build non-source objects the same way I build source objects.

Understandable. You build source objects using an action - and you could create your own action for building non-source objects (maybe using Bob?) in Code for IBM i today. There is no need for a default action for this.

> It makes the feature more visible, accessible, and convenient.

Actions are very convenient and easy to access. However, default actions are shown to every user - not just the ones requiring the action. Remember, what's convenient for you may be inconvenient for other users.

> It encourages the use of pseudo source, which is good for the reasons outlined above

Agree - but the UI for users not using pseudo source are affected by the pseudo source feature. An action never used by them will take up space and may require additional keystrokes or mouse movement to move over.

> It adds to the already great feature set of the extension, which encourages adoption of the tool.

Agreed - but it still should be optional and not forced to all users of this extension. I'm for opt-in and not opt-out.

Thank you for taking the time to describe the reasons for your request and arguments for why this should be included in Code for IBM i. I'm not opposing the functionality, just that it should be forced into Code for IBM i. I hope you understand my reasoning...

worksofliam:
@chrjorgensen 

> it is my opinion that the pseudo source function should be implemented in its own extension - using Code for IBM i as the API for the IBM i.

Funnily enough, this was my original idea. For sure it more likely belongs in another extension (as does the ability to execute single CL lines that exists today, e.g. #1387), but does it belong in the CL extension? Or another unique extension?

chrjorgensen:
@worksofliam A new extension - all users have the CL extension and inserting the feature here would be the same as in this extension...

worksofliam:
Okay, so my plan is to move both the CL runner and pseudo runner into a new extension. I'll do that next week!

sebjulliand:
Thanks for sharing your thoughts @JDubbTX ! It sparkled a good discussion and it looks we're headed in the right direction 😄

JDubbTX:
I'm fine with the opt-in, and not making it a default - but disagree it should be in its own extension.  

I'm not crazy about extensions that are not stand alone, and require other extensions to work.

How would this work without code for IBM i?  What is the reasoning for making it its own extension instead of just adding a setting to turn it on?

Even if it could be stand alone, what is the use case for having it while not having code4i?

JDubbTX:
> @chrjorgensen
> 
> > it is my opinion that the pseudo source function should be implemented in its own extension - using Code for IBM i as the API for the IBM i.
> 
> Funnily enough, this was my original idea. For sure it more likely belongs in another extension (as does the ability to execute single CL lines that exists today, e.g. #1387), but does it belong in the CL extension? Or another unique extension?

I also think what was lost in my request in ryver was the ability to run DDL as pseudo source as well.  It is commone to have a list of SQL commands that define DDL objects, like tables, indexes, triggers, etc.  Admitedly I don't know if code 4 i already has this capability.  Even if it did, I would want to be able to create those objects the same way - as a built in action that I can turn on, instead of having to define my own.

chrjorgensen:
@JDubbTX 

Links to other extensions are already in play today for Code for IBM i: it requires other extensions and some extensions require Code for IBM i:

1. the IBM i language syntax highlighting is done by the [vscode-ibmi-languages](https://marketplace.visualstudio.com/items?itemName=barrettotte.ibmi-languages) extension. When you installed C4i, this extension was installed automatically in your VS Code by VS Code, together with [Code for IBM i Walkthroughs](https://marketplace.visualstudio.com/items?itemName=HalcyonTechLtd.vscode-ibmi-walkthroughs). You don't have to know which required extensions to install, VS Code handles that for you.
1. the [RPGLE](https://marketplace.visualstudio.com/items?itemName=HalcyonTechLtd.vscode-rpgle), [DB2 for IBM i](https://marketplace.visualstudio.com/items?itemName=HalcyonTechLtd.vscode-db2i), [IBM i Renderer](https://marketplace.visualstudio.com/items?itemName=HalcyonTechLtd.vscode-displayfile) and other extensions require C4i. When you install any of them and don't have C4i installed already, VS Code will install C4i when installing the requested extension.

RDi has the same pluggable architecture - think of the great iSphere plugin, which adds functionality to RDi like TODO's, labels, compare across systems etc. on top of the functionality already in RDi.

Actually, all IDE I can think of has this pluggable architecture... and relations between extensions / plugins are not an issue in my experience. VS Code is great at handling related extensions. I've only had plugin problems in RDi, where I had to install iSphere manually because the upgrade function did not work.

Another reason for having code in a separate extension is to keep the Code for IBM i code base to a minimum - making it easier to enhance, test and maintain the code. If all code and functions related to IBM i were in Code for IBM i, development would have slowed tremendously and the extension would not have been at the level it is today.

I don't think you will experience any problems if pseudo source becomes a separate extension. As said, VS Code will handle it for you.

> I also think what was lost in my request in ryver was the ability to run DDL as pseudo source as well. It is common to have a list of SQL commands that define DDL objects, like tables, indexes, triggers, etc. Admitedly I don't know if code 4 i already has this capability. Even if it did, I would want to be able to create those objects the same way - as a built in action that I can turn on, instead of having to define my own.

Having SQL DDL commands to define SQL objects like tables, indexes, view, procedures, functions, sequences etc.  stored in a source member / streamfile and having build procedures for these objects is common practice in most IBM i shops that I know of. The build procedures can be initiated by an action in Code for IBM i - like the build for programs and other non-SQL objects. The action would issue the RUNSQLSTM CL command, which can execute SQL scripts in members and streamfiles. If you want to run a single SQL statements in a group of SQL DDL statements, you can install the [DB2 for IBM i](https://marketplace.visualstudio.com/items?itemName=HalcyonTechLtd.vscode-db2i) extension, which allows you to do that.


IMHO it's best to keep the build procedure outside C4i and not turn C4i into a "build machine", since there are many different build methods (CL program, SQL script, Bob, make file etc) and people / companies use the one(s) suited for them. Default actions for each and every one of the build methods would clutter the UI and annoy the users - but actions can be added and modified by the users to fit their IBM i installation, thus precisely configured to each users needs.


I hope and expect you will have a great experience, even with pseudo source as a separate extension and not a native feature of Code for IBM i. 😃

JDubbTX:
> @JDubbTX
> 
> Links to other extensions are already in play today for Code for IBM i: it requires other extensions and some extensions require Code for IBM i:
> 
> 1. the IBM i language syntax highlighting is done by the [vscode-ibmi-languages](https://marketplace.visualstudio.com/items?itemName=barrettotte.ibmi-languages) extension. When you installed C4i, this extension was installed automatically in your VS Code by VS Code, together with [Code for IBM i Walkthroughs](https://marketplace.visualstudio.com/items?itemName=HalcyonTechLtd.vscode-ibmi-walkthroughs). You don't have to know which required extensions to install, VS Code handles that for you.

> 2. the [RPGLE](https://marketplace.visualstudio.com/items?itemName=HalcyonTechLtd.vscode-rpgle), [DB2 for IBM i](https://marketplace.visualstudio.com/items?itemName=HalcyonTechLtd.vscode-db2i), [IBM i Renderer](https://marketplace.visualstudio.com/items?itemName=HalcyonTechLtd.vscode-displayfile) and other extensions require C4i. When you install any of them and don't have C4i installed already, VS Code will install C4i when installing the requested extension.

I understand where you are coming from.  I really do.  I feel like this line of thinking can be taken too far though.  I think we both agree that basic functionality should be included with the base extension.  Imagine if every little function of code 4 i were in its own extension?  That would gum up the marketplace and cause great confusion.  I would also say that just because it _has_ been done, isn't necessarily an argument for what _should_ be done - provide the specific reasoning that those extensions are not part of the base code4i, and apply the same reasoning to what we are talking about here.  Putting code into separate extensions isn't the same as writing good, modularized code.

> RDi has the same pluggable architecture - think of the great iSphere plugin, which adds functionality to RDi like TODO's, labels, compare across systems etc. on top of the functionality already in RDi.

I'm familiar with iSphere and use it a bunch.  Some of its functionality, especially the member text decorations, I wonder why is wasn't included in base RDi.  (It is included by default in code4i).  The good folks that maintain RDi really should include that, its pretty basic.
 
> Actually, all IDE I can think of has this pluggable architecture... and relations between extensions / plugins are not an issue in my experience. VS Code is great at handling related extensions. I've only had plugin problems in RDi, where I had to install iSphere manually because the upgrade function did not work.

I'm not arguing against having a plugin architecture.  I love the extensibility of VSCode and I use it in many capacities, not just IBM i development.  My argument is that it doesn't make sense to put this specific function in a plugin / extension and that it deserves to be part of the main code4i extension, with an additional notion that it doesn't provide enough functionality to warrant its own extension.

> Another reason for having code in a separate extension is to keep the Code for IBM i code base to a minimum - making it easier to enhance, test and maintain the code. If all code and functions related to IBM i were in Code for IBM i, development would have slowed tremendously and the extension would not have been at the level it is today.

Respectfully, I feel like you are overestimating the amount of code bloat and UI obtrusiveness that the functionality will bring, especially if you make it opt in, which I've already conceded.  That being said, I'm not a javascript or typescript guru - having just some exposure in the last couple of years, so if this adds a thousand lines of code, by all means keep it seperate.  My hunch is that Liam can get this done in a few paragraphs, but again, perhaps I'm wrong.

> I don't think you will experience any problems if pseudo source becomes a separate extension. As said, VS Code will handle it for you.

I agree I won't have problems.  I mainly work in development containers these days, and when I build my development container for IBM i one of the first things I do is install the IBM i development pack.  This would be part of the development pack, correct?  I would think it would certainly disserve to be, over some less disserving extensions that are already part of it.  Do you use the development pack, or do you prefer to install extensions individually?  If you do use the development pack, you would have to turn around and uninstall the extension each time you installed the development pack, so by putting it in its own extension, you are actually adding steps to the majority of folks who won't use it.

> Having SQL DDL commands to define SQL objects like tables, indexes, view, procedures, functions, sequences etc. stored in a source member / streamfile and having build procedures for these objects is common practice in most IBM i shops that I know of. The build procedures can be initiated by an action in Code for IBM i - like the build for programs and other non-SQL objects. The action would issue the RUNSQLSTM CL command, which can execute SQL scripts in members and streamfiles. If you want to run a single SQL statements in a group of SQL DDL statements, you can install the [DB2 for IBM i](https://marketplace.visualstudio.com/items?itemName=HalcyonTechLtd.vscode-db2i) extension, which allows you to do that.

Doesn't this make my point, not yours?  If its common practice in IBM i shops to have build procedures, why not support a build action for these by default?  

I'm really not trying to be difficult, I just want to take a moment and clarify what we are discussing.  It sounds like you are agreeing with me that having pseudo source is both _a good thing_, and "common practice". 

The main points of contention I see are 1) should this pseudo source functionality be enabled by default (I've already conceded that it doesn't need to be)? and 2) Should it be in its own extension?

> IMHO it's best to keep the build procedure outside C4i and not turn C4i into a "build machine", since there are many different build methods (CL program, SQL script, Bob, make file etc) and people / companies use the one(s) suited for them. Default actions for each and every one of the build methods would clutter the UI and annoy the users - but actions can be added and modified by the users to fit their IBM i installation, thus precisely configured to each users needs.

I disagree with this wholeheartedly.  If not for building, then code 4 i would just be for browsing and looking at code.  You can do that without code4i, and even browse _syntax highlighted_ RPG and CL code without code4i.  I'm not advocating for a full fledged build solution like Bob, Arcad, and the like.  To me, building pseudo source is just simple, and no different than building source objects, and encourages good development practice.  If you don't like having the extra functionality, just don't turn it on.  Since we have agreed to make it  opt in, most folks probably won't even realize its there.

> I hope and expect you will have a great experience, even with pseudo source as a separate extension and not a native feature of Code for IBM i. 😃

I have enjoyed this discourse, and I DO have a "great experience".  You maintainers of Code4i are absolute superstars in my book. I appreciate the addition of my suggestion in whatever form it ends up taking.

Issue #1804, opened 2024-01-25
Server Side Settings


worksofliam:
Settings that are defined on the Server which can then be applied to every VS Code user that connects to it.

This file will exist at `/etc/vscode/settings.json`. Every time the client connects, it will grab these values and assign them to the local client settings. The local client cannot change these settings in the existing Connection Settings and the fields managed by the server will be readonly.

Rough plan

* [x] Search for settings file on connect
* [x] Update client settings
* [x] Mark specific properties/fields (in the settings) as readonly when managed by server
* [ ] ~Ability to edit server settings in the same Connection Settings UI~
* [ ] Ability to open server settings file from Connection Settings UI
* [x] Schema for settings file

worksofliam:
@codefori/core before I continue work on this PR, can you review what I have now and provide any initial feedback for this idea? Thanks!

PR #1805, opened 2024-01-25
Feature/server_settings

worksofliam:
### Changes

Implements features as defined in #1804

### Checklist

* [ ] have tested my change

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -54,6 +54,8 @@ export default class IBMi {
   remoteFeatures: { [name: string]: string | undefined };
   variantChars: { american: string, local: string };
   lastErrors: object[];
+
+  loadedServerSettings: string[];
   config?: ConnectionConfiguration.Parameters;
 
   commandsExecuted: number = 0;
@@ -108,6 +110,7 @@ export default class IBMi {
      * */
     this.lastErrors = [];
 
+    this.loadedServerSettings = [];
   }
 
   /**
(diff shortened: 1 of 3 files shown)


worksofliam:
@codefori/core before I continue work on this PR, can you review what I have now and provide any initial feedback for this idea? Thanks!

src/api/IBMi.ts
```diff
@@ -16,6 +16,8 @@ export interface MemberParts extends IBMiMember {
   basename: string
 }
 
+const SERVER_SETTINGS_PATH = `/.vscode/connections.json`;
```
chrjorgensen, line 19: This is my only concern for this PR - the location of the server setting JSON file!

I don't like creating a hidden directory in the root - especially since the recommendation from IBM is to remove write capability for users for root. I think it's better to store the directory in `/etc` where server configuration files are normally placed.

For the name of the directory and JSON file, I suggest the following:

```suggestion
const SERVER_SETTINGS_PATH = `/etc/Code4IBMi/common_server_settings.json`;
```

This is more intuitive IMHO - the file contains server settings common for all users of the Code for IBM i extension on this server.

WDYT? Too long filename?

chrjorgensen (changes requested):
@worksofliam I just gave my thoughts on the settings filename - my test will wait for the server settings UI to be implemented.

Looking good so far... :heart_eyes:

src/api/IBMi.ts
worksofliam: I will shorted to simply `settings.json`!

worksofliam:
Needs to be restarted.

Issue #1712, opened 2023-12-12
File cleanups:


worksofliam:
* The `connect` API needs to be cleaned up and modularised: https://github.com/codefori/vscode-ibmi/blob/master/src/api/IBMi.ts
* The commands defined in `loadAllOfExtension` need to be moved into unique files to clean it up because they're so big: https://github.com/codefori/vscode-ibmi/blob/master/src/instantiate.ts

krethan:
@worksofliam I am working on cleaning up the connect API and modularising. Can the rest of this one be assigned to me?

worksofliam:
This is done

PR #1807, opened 2024-01-27
This is for issue #1712

krethan:
This is for issue #1712

Simplify the logic in the IBMi class by creating new IBMiSettings class.


worksofliam:
@krethan This is a nice change. I am going to work with the team to get a formal review ready - but in the meantime, I think the built-in formatter should be run over the `IBMiSettings.ts` file. You can open the command palette while inside that file, and search for 'Format Document'.

Thanks!

krethan:
@worksofliam Formatted IBMi.ts and IBMiSettings.ts

krethan:
@worksofliam Fixed merge issue in case it was preventing review.

worksofliam:
@krethan just wanted you to know that we are planning on taking a stab at reviewing this next week. We have been quite busy! Sorry about the wait.

src/api/IBMiSettings.ts
```diff
@@ -0,0 +1,723 @@
...
+    names: [`QZDFMDB2.PGM`]
+  },
+  {
+    path: `/QIBM/ProdData/IBMiDebugService/bin/`,
+    specific: `startDebugService.sh`,
+    names: [`startDebugService.sh`]
+  }
+];
+
+export default class IBMiSettings {
```
worksofliam, line 34: @sebjulliand and I were chatting about this class name and I think something more like `ConnectionSettings` or `InstanceSettings` might be more appropriate.

src/api/IBMiSettings.ts
```diff
@@ -0,0 +1,723 @@
...
+
+  private connection: IBMi;
+  private progress: vscode.Progress<{ message?: string; increment?: number }>;
+  private reconnecting?: boolean;
+  private connectionObject: ConnectionData;
+  private tempLibrarySet: boolean;
+  private tempDirSet: boolean;
+  private delayedOperations: Function[];
+
+  constructor(connection: IBMi, progress: vscode.Progress<{ message?: string; increment?: number }>, connectionObject: ConnectionData, delayedOperations: Function[], reconnecting?: boolean) {
```
worksofliam, line 44: This constructor is likely too long considering most of the parameters end up as private member fields.

Please consider using `private connection: IBMi, private progress: ...`, etc, which then will mean you don't need to assign in the constructor or define them above the constructor.

worksofliam: More detail here: https://stackoverflow.com/a/48571668

worksofliam (changes requested):
Generally on the right track. I've left some initial comments, but I believe there is likely a more to do outside of my review.

I don't know if the rest of @codefori/core agree with me, but ideally, I'd like to see this settings class not make use of the `vscode` namespace at all, and instead return the suffice information from each class method so the UI can be handled independently.

src/api/IBMiSettings.ts
krethan: @worksofliam I agree we should seperate the UI from this class. I will refactor the UI code back into the connection method of the IBMi class.

krethan: @worksofliam @sebjulliand I will go with ConnectionSettings for the class name

krethan:
@worksofliam I am about 75% through refactoring the ConnectionSettings.ts class now.

krethan:
@worksofliam Have couple introduced errors now. Working my way through them. Bug in home directory and another when asked to set bash to default, disconnect, connect again and still asking to set bash. Will ask for code review after these are fixed and committed.

krethan:
@worksofliam Okay ready for new review. I hope this one is good.

worksofliam:
@krethan As we get ready for the 2.8.0 release, is it possible you can fix the conflicts here? Thanks!

worksofliam:
@krethan It looks like the entire `out` directory has been committed. This likely shouldn't be the case. Any chance you can take a look at that?

krethan:
@worksofliam I reverted the commits back to remove everything in the out directory. Makes sense.

worksofliam:
@krethan We are so close! Please could you also fix the last remaining conflict?

Sorry this is taking a while. This is a very high impact PR so we are treading lightly right now.

src/api/ConnectionSettings.ts
```diff
@@ -0,0 +1,563 @@
...
+          if (badLib) badLibs.push(badLib);
+        });
+      }
+    }
+    
+    return badLibs;
+
+  }
+
+  private appendOutput(content: string) {
```
worksofliam, line 557: Looks like this method could be removed as it is unreferenced.

src/api/ConnectionSettings.ts
```diff
@@ -0,0 +1,563 @@
...
+      command: `CHKOBJ OBJ(QSYS/QCPTOIMPF) OBJTYPE(*DTAARA)`,
+      noLibList: true
+    });
+    if (QCPTOIMPF?.code === 0) {
+      dataArea = true;
+    }
+    return dataArea;
+  }
+
+  async deleteQCPTOIMPF(): Promise<CommandResult>{
```
worksofliam, line 295: This function is returning a function, which returns a Promise, so there is no need for the `async` keyword on the method here.

src/api/ConnectionSettings.ts
```diff
@@ -0,0 +1,563 @@
...
+    });
+
+    if (QCPFRMIMPF?.code === 0) {
+      dataArea = true;
+    }
+
+    return dataArea;
+  }
+
+  async deleteQCPFRMIMPF(): Promise<CommandResult> {
```
worksofliam, line 318: Same here with the `async` keyword.

src/api/ConnectionSettings.ts
```diff
@@ -0,0 +1,563 @@
...
+            if (!messages.findId(`CPF2125`)) {
+              // @ts-ignore We know the config exists.
+              throw(`Temporary data not cleared from ${this.config.tempLibrary}.`);
+            }
+          }
+        })
+    }
+  }
+
+  async clearTempDataIFS() {
```
worksofliam, line 265: `async` keyword is not required as `await` is not being used in this method.

src/api/ConnectionSettings.ts
```diff
@@ -0,0 +1,563 @@
...
+        // Directory not created
+      }
+    }
+
+    if (!this.tempDirSet && this.connection.config) {
+      this.connection.config.tempDir = `/tmp`;
+    }
+  }
+
+  async clearTempDataSys(){
```
worksofliam, line 246: `async` keyword not required here either.

src/api/IBMi.ts
```diff
@@ -210,274 +183,112 @@ export default class IBMi {
...
-          } else {
-            // Directory not created
-          }
-        }
-
-        if (!tempDirSet) {
-          this.config.tempDir = `/tmp`;
-        }
+        //Clear temporary data
+        if (await connSettings.getTempLibrarySet() && this.config?.autoClearTempData) {
```
worksofliam, line 245: `await` not required here because `getTempLibrarySet` does not return a Promise.

src/api/IBMi.ts
```diff
@@ -486,35 +297,29 @@ export default class IBMi {
             });
           }
 
-          const QCPFRMIMPF = await this.runCommand({
-            command: `CHKOBJ OBJ(QSYS/QCPFRMIMPF) OBJTYPE(*DTAARA)`,
-            noLibList: true
-          });
-
-          if (QCPFRMIMPF?.code === 0) {
+          if (await connSettings.checkQCPFRMIMPF) {
```
worksofliam, line 300: Brackets are missing from this method call.

src/api/IBMi.ts
```diff
@@ -526,39 +331,14 @@ export default class IBMi {
...
-                } else {
-                  for (const name of feature.names)
-                    if (files.includes(name))
-                      this.remoteFeatures[name] = feature.path + name;
-                }
-              }
-            } catch (e) {
-              console.log(e);
-            }
+          const remoteApps = await connSettings.getRemoteApps();
```
worksofliam, line 336: `await` not required since `getRemoteApps` does not return a Promise.

worksofliam (changes requested):
I have left you a review to get you started. There are lot of ends that need some cleaning up, espcially in regards to the use of `async` and `await`. I have high hopes for this PR, but it needs to be tight and tested before we can merge.

Other suggestions:

* Use the VS Code built-in formatter on `ConnectionSettings.ts`: F1 -> Format Document
* A few methods in `ConnectionSettings` use try catch, even though the calls inside of them do not crash - for example, `sendCommand` does not crash, but returns the exit code. It'd be nice to have those removed.
* A lot of the methods return `void`. It would be nice to make them more verbose, perhaps returning `boolean` or other in the future so we can write test cases for them.

src/api/ConnectionSettings.ts
```diff
@@ -0,0 +1,563 @@
...
+          "export PATH=/QOpenSys/pkgs/bin:$PATH"
+        );
+      }
+      await content.writeStreamfile(bashrcFile, bashrcContent.join("\n"));
+    }
+  }
+
+  async validateLibraryList(): Promise<string[]> {
+    
+    let validLibs: string[] = [];
```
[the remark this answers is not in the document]
krethan: Removed.

src/api/ConnectionSettings.ts
krethan: Method removed.

src/api/ConnectionSettings.ts
krethan: removed Async.

krethan:
@worksofliam Should be good for review again.

PR #1811, opened 2024-01-30
`checkObject` to support checking member

worksofliam:
### Changes

* Adds `member` property to `checkObject` API, allowing us to check if members exist
* Created a new `toCl` static method to generate the formatted CL commands. I'd like to move towards using this instead of hardcoding CL strings everywhere.
* Tests for `checkObject` API uses

### Checklist

* [x] have tested my change


sebjulliand (approved):
@worksofliam good one; `toCl` is a nice addition!
I pushed a few commits to fix what I think was a bug and add a test.
I'll let you cross-review it if you will. Then you can merge if you feel like it.

PR #1818, opened 2024-02-01
Correctly read files from the IFS

worksofliam:
### Changes

I noticed our IFS filesystem wasn't forcing everything as UTF8, which meant we couldn't open up non UTF8 files like PDFs, images, and spreadsheets that reside in the IFS. 

I have added created new raw APIs that we should use instead, and made the previous download/write APIs call the raw API and convert to UTF8 to keep the previous functionality.

Test cases are continuing to pass.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -67,7 +67,12 @@ export default class IBMiContent {
     }
   }
 
-  async downloadStreamfile(remotePath: string, localPath?: string) {
+  /**
+   * 
+   * @param remotePath Remote IFS path
+   * @param localPath Local path to download file to
+   */
+  async downloadStreamfileRaw(remotePath: string, localPath?: string) {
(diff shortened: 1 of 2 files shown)


chrjorgensen (changes requested):
@worksofliam What is this supposed to change? I get the same output with and without this change, whatever the streamfile CCSID - but that is maybe the goal?

I had hoped this PR was for opening spreadsheets and images stored in the IFS with VS Code... 😭 😄

worksofliam:
@chrjorgensen That is funny because I was not able to open any non UTF8 files (with extensions installed to render PDFs and spreadsheets) prior to this PR.

Instead of our API returning strings to the file system provider, we now provide the raw bytes - which is appropriate for the APIs anyway.

chrjorgensen (approved):
@worksofliam Works as expected - approved and merged. 👍

PR #1820, opened 2024-02-01
Display more explicit error message when connection fails

sebjulliand:
### Changes
This PR makes the connection failure messages more explicit, depending on the error.

There are three error cases for which we provide an explicit message:
- Host is unreachable
- Port is unreachable
- Credentials are wrong

If the error case cannot be determined from the error returned by SSH, then the error is displayed as it used to be.

Example of the three possible messages:
image

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -874,7 +874,7 @@ export default class IBMi {
         };
       });
 
-    } catch (e) {
+    } catch (e: any) {
 
       if (this.client.isConnected()) {
         this.client.dispose();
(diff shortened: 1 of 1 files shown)


src/api/IBMi.ts
```diff
@@ -887,9 +887,20 @@ export default class IBMi {
         return this.connect(connectionObject, true);
       }
 
+      let error = e;
```
chrjorgensen, line 890: Why this line? Is `error` not a string instead of an error object?

chrjorgensen (changes requested):
@sebjulliand Nice PR - works as expected. 👍 

Only a small comment...

src/api/IBMi.ts
sebjulliand: This used to be the default behavior; we returned `e`, which is an object of any kind. I kept it like that because, in case an unexpected error happens, we can still report it as it is.

chrjorgensen (approved):
@sebjulliand Okay - approved, merge when you feel for it... 🎉

PR #1826, opened 2024-02-02
Fix/no_member_validation

worksofliam:
### Changes

Removes additional check to find out if the chosen member exists. There are two reasons why I think this is allowed:

1. Because if the user enters a path manually, we can assume it does exist. Even if it doesn't exist, the tab will open with a valid error saying the member cannot be found. Fine by me.
2. If the user is using the ability to search with the asterisk filter, then we already know that the member exists.
3. Turns out, on systems with a huge member count this statement was taking a long time to run. That is a problem.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/instantiate.ts
+++ b/src/instantiate.ts
@@ -422,27 +422,13 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
                 const file = `${connection!.sysNameInAmerican(selectionSplit[1])}`;
                 const member = path.parse(`${connection!.sysNameInAmerican(selectionSplit[2])}`);
                 member.ext = member.ext.substring(1);
-                const fullMember = await content!.runSQL(`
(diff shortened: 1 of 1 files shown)


worksofliam:
@chrjorgensen Yes! Be my guest. Perhaps it is also an option to use our `checkObject` API which has a member attribute now?

chrjorgensen:
@worksofliam Unfortunately the `checkObject`API does not support member extension (uses `CHKOBJ` behind the scenes), and the member extension is the problem here... but I have other ideas, stay tuned... :smiley:

chrjorgensen:
@worksofliam New solution is ready - but not using your commits. Should I put it in another PR?

chrjorgensen:
PR closed, replaced by #1867.

PR #1828, opened 2024-02-05
API change: use default CCSID to run SQL queries when needed

sebjulliand:
### Changes
Running SQL queries in a job whose CCSID is 65535 is bad. This PR tries to avoid conversion errors by running a `CHGJOB CCSID(xx)` command before every SQL statement run through `QZDFMDB2`.

In above command, `xx` is the job's default CCSID which is retrieved at connection time (once then cached).

Noticeable changes:
- Default CCSID is retrieved using the `ACTIVE_JOB` SQL service. If it fails, we fall back to using `DSPJOB OPTION(*DFNA)` and parsing the output.
- the `runSQL` method has been simplified a bit. The logic to sanitize the statement about to be run has been put in a separate method in `Tools` and there is a unit test to check it.
- `QZDFMDB2` allows to run CL command simply by prefixing them with `@`. It's convenient but it turns out such statements are always run in a separate job. So the method that sanitizes `QZDFMDB2` statements will also rewrite every `@command` into `Call QSYS2.QCMDEXC('command')`.
- A varchar conversion has been added in the `getMemberList` function to avoid a 1200 -> 65535 conversion issue, apparently unavoidable even with `CHGJOB`. The member's TEXT field is using CCSID 1200.
- The Default CCSID has been added to the `Report Issue` template. It is also reported in the Code for i output if QCCSID is bad.
image

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -7,7 +7,6 @@ import path from 'path';
 import { instance } from "../instantiate";
 import { CommandData, CommandResult, ConnectionData, IBMiMember, RemoteCommand } from "../typings";
 import { CompileTools } from "./CompileTools";
-import IBMiContent from "./IBMiContent";
 import { CachedServerSettings, GlobalStorage } from './Storage';
 import { Tools } from './Tools';
 import * as configVars from './configVars';
@@ -15,8 +14,9 @@ import * as configVars from './configVars';
 export interface MemberParts extends IBMiMember {
   basename: string
 }
+const CCSID_SYSVAL = -2;
 
-let remoteApps = [ // All names MUST also be defined as key in 'remoteFeatures' below!!
+const remoteApps = [ // All names MUST also be defined as key in 'remoteFeatures' below!!
   {
     path: `/usr/bin/`,
     names: [`setccsid`, `iconv`, `attr`, `tar`, `ls`]
@@ -50,7 +50,8 @@ export default class IBMi {
   defaultUserLibraries: string[];
   outputChannel?: vscode.OutputChannel;
   aspInfo: { [id: number]: string };
-  qccsid: number | null;
+  qccsid: number;
+  defaultCCSID: number;
   remoteFeatures: { [name: string]: string | undefined };
   variantChars: { american: string, local: string };
   lastErrors: object[];
@@ -74,8 +75,8 @@ export default class IBMi {
      * the root of the IFS, thus why we store it.
      */
     this.aspInfo = {};
-
-    this.qccsid = null;
+    this.qccsid = CCSID_SYSVAL;
+    this.defaultCCSID = 0;
 
     this.remoteFeatures = {
       git: undefined,
@@ -560,8 +561,20 @@ export default class IBMi {
         }
 
         if (this.remoteFeatures[`QZDFMDB2.PGM`]) {
-          let statement;
-          let output;
+          //Temporary function to run SQL
+          const runSQL = async (statement: string) => {
+            const output = await this.sendCommand({
+              command: `LC_ALL=EN_US.UTF-8 system "call QSYS/QZDFMDB2 PARM('-d' '-i')"`,
+              stdin: statement
+            });
+
+            if (output.code === 0) {
+              return Tools.db2Parse(output.stdout);
+            }
+            else {
+              throw new Error(output.stdout);
+            }
+          };
 
           // Check for ASP information?
           if (quickConnect === true && cachedServerSettings?.aspInfo) {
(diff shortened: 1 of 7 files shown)


src/api/Tools.ts
```diff
@@ -242,20 +243,35 @@ export namespace Tools {
...
     }
   }
+
+  export function fixQZDFMDB2Statement(statement: string) {
+    return statement.split("\n").map(line => {
+      if (line.startsWith('@')) {
+        //- Escape all '
+        //- Remove any trailing ;
+        //- Put the command in a Call QSYS2.QCMDEXC statement
+        line = `Call QSYS2.QCMDEXC('${line.substring(1, line.endsWith(";") ? line.length - 1 : undefined).replaceAll("'", "''")}');`;
```
worksofliam, line 269: What is the purpose of this? Why can't you just pass in the `@<command>` instead of replacing with `QCMDEXC`?

sebjulliand: Checkout the PR description; `@commands` don't run in the same job as the other statements. Using QCMDEXC commands do.

src/api/IBMiContent.ts
```diff
@@ -217,25 +217,24 @@ export default class IBMiContent {
...
     if (QZDFMDB2) {
-      // Well, the fun part about db2 is that it always writes to standard out.
-      // It does not write to standard error at all.
-
-      // if comments present in sql statement, sql string needs to be checked
-      if (statement.search(`--`) > -1) {
-        statement = this.fixCommentsInSQLString(statement);
-      }
+      if (this.chgJobCCSID === undefined) {
+        this.chgJobCCSID = (this.ibmi.qccsid < 1 || this.ibmi.qccsid === 65535) && this.ibmi.defaultCCSID > 0 ? `@CHGJOB CCSID(${this.ibmi.defaultCCSID});\n` : '';
```
worksofliam, line 232: I don't like that we are assigning to this class field every time we call `runSQL`. In theory, this expression should not change after the connection has been created and therefore don't think we should run this every time.

src/api/Tools.ts
```diff
@@ -242,20 +243,35 @@ export namespace Tools {
...
-    else{
+    else {
       return {
         library: parts[0],
         name: parts[1]
       }
     }
   }
+
+  export function fixQZDFMDB2Statement(statement: string) {
```
worksofliam, line 263: I think we should have a shorter name with a jsdoc comment to indicate that it is to be used in combination with `QZDFMDB2`.

Perhaps something like `fixLegacyStatement` or other?

worksofliam (changes requested):
Requested some changes, simple stuff mostly.

I would also appreciate an additional test for `runSQL` that includes running a CL command and then a statement.

Perhaps something like `CHGLIBL` and then `select * from QSYS2. LIBRARY_LIST_INFO`.

src/api/IBMiContent.ts
sebjulliand: I don't like it either which is why it's lazily assigned 😁
The `if` tests specifically for `undefined` and after that, the field is either a CHGJOB command or a blank string, but not `undefined` anymore.

sebjulliand:
Thanks @worksofliam !
I pushed a new test and the name modification. See my comment for your other comment.

src/api/IBMiContent.ts
worksofliam: Super silly of me to not see that! Of course it can't be assigned to again because it will never be `undefined` again!! Thanks for making me realize that.

src/testing/content.ts
```diff
@@ -591,5 +591,15 @@ export const ContentSuite: TestSuite = {
         assert.ok(exists);
       }
     },
+    {
+      name: `Test @clCommand + select statement`, test: async () => {
+        const content = instance.getContent()!;
+        
+        const [result] = await content.runSQL(`@CRTSAVF FILE(QTEMP/UNITTEST) TEXT('Code for i test');\nSelect * From Table(QSYS2.OBJECT_STATISTICS('QTEMP', '*FILE')) Where OBJATTRIBUTE = 'SAVF';`);
```
worksofliam, line 598: This is a great test case!

worksofliam:
@sebjulliand Before I leave my final review, I do think as part of this, before we release it especially, is that we should document this at a high level on our CCSID page. Something like... yep, we're setting the CCSID, but you really should set it on your user profile.

sebjulliand:
@worksofliam agreed, and maybe add a button to open the URL on the notification that says naught-no-ccsid.

worksofliam:
@sebjulliand Is that something you want to include in this PR or another time?

src/api/IBMi.ts
```diff
@@ -50,7 +50,8 @@ export default class IBMi {
   defaultUserLibraries: string[];
   outputChannel?: vscode.OutputChannel;
   aspInfo: { [id: number]: string };
-  qccsid: number | null;
+  qccsid: number;
```
worksofliam, line 53: I am starting to wonder if this `qccsid` property should be renamed. Years ago we used to only check for `QCCSID`, but now we also store the job CCSID in this first.

What do you think?

worksofliam:
Baring my two comments about:

1. When do we want to update the docs?
2. `qccsid` property name

The tests are passing and I am happy with the code.

image

sebjulliand:
Thanks @worksofliam ! I'll merge now, and open another PR for renaming `qccsid` to something better and display a better error message when it is `65535`.

PR #1837, opened 2024-02-09
Stop letting the same document be opened twice

worksofliam:
### Changes

* If you try and open a doucment, but already have it open in the oppisite edit mode, then you will get a warning and that open document will take focus.
* Clicking on an error should no longer open up a new document will try to use existing document URIs to inherit the readonly mode.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/Tools.ts
+++ b/src/api/Tools.ts
@@ -277,7 +277,6 @@ export namespace Tools {
   function uriStringWithoutFragment(uri: vscode.Uri) {
     // To lowercase because the URI path is case-insensitive
     const baseUri = uri.scheme + `:` + uri.path;
-    const isCaseSensitive = (uri.scheme === `streamfile` && uri.path.startsWith(`/QOpenSys/`));
     const isCaseSensitive = (uri.scheme === `streamfile` && /^\/QOpenSys\//i.test(uri.path));
   }
 
--- a/src/api/Tools.ts
+++ b/src/api/Tools.ts
@@ -266,8 +266,8 @@ export namespace Tools {
    * Without this, it's possible for the same document to be opened twice simply due to the readonly flag.
    */
   export function findExistingDocumentUri(uri: vscode.Uri) {
-    const bathUriString = uriStringWithoutFragment(uri);
-    const possibleDoc = vscode.workspace.textDocuments.find(document => uriStringWithoutFragment(document.uri) === bathUriString);
+    const baseUriString = uriStringWithoutFragment(uri);
+    const possibleDoc = vscode.workspace.textDocuments.find(document => uriStringWithoutFragment(document.uri) === baseUriString);
     if (possibleDoc) {
       return possibleDoc.uri;
     }
--- a/src/api/Tools.ts
+++ b/src/api/Tools.ts
@@ -259,50 +259,51 @@ export namespace Tools {
       }
     }
   }
-}
 
-/**
- * We do this to find previously opened files with the same path, but different case OR readonly flags.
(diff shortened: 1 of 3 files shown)


chrjorgensen (changes requested):
@worksofliam Before testing, one question:

Some filesystems in IFS are case-sensitive, e.g. /QOpenSys... Isn't it a problem when you change the URI to lowercase?

Could we test for this before making the URI lowercase - or would it not work? I'm not that familiar with the URI in VS Code...

worksofliam:
@chrjorgensen 

I assume you're talking [about this line of code](https://github.com/codefori/vscode-ibmi/pull/1837/files#diff-dd1627a4d8b56dfba1fca9ceb939163f01edd1f2e5581d7ea33f6ef99d9e10ccR268) where we compare two URIs.

Yes, you're right. For IFS (`streamfile` internally) we should not be lowercasing the value. I am going to change this in a new commit.

chrjorgensen (changes requested):
> For IFS (`streamfile` internally) we should not be lowercasing the value.

@worksofliam This was not quite what I meant - lowercasing IFS is fine, since most IFS filesystems are case-insensitive and `readme.md` is the same as `README.MD`. 

But `/QOpenSys` (and UDFS, which I think we can ignore) is case-sensitive, and `readme.md` is NOT the same as `README.MD`! So only IFS paths starting with '/QOpenSys' should be kept as-is and not be lowercased.

I'm after this since IBM YUM packages and other OSS tools are most often stored in `/QOpenSys`...

worksofliam:
@chrjorgensen I have applied the logic as you recommended. Take another look when you get a moment - thanks!!

chrjorgensen (changes requested):
I ran the extension tests - and a lot of them failed (didn't on the master branch):

image

???

worksofliam:
@chrjorgensen Fixed now.. give it another go

image

src/api/Tools.ts
```diff
@@ -260,6 +260,31 @@ export namespace Tools {
     }
   }
 
+
+  /**
+   * We do this to find previously opened files with the same path, but different case OR readonly flags.
+   * Without this, it's possible for the same document to be opened twice simply due to the readonly flag.
+   */
+  export function findExistingDocumentUri(uri: vscode.Uri) {
+    const bathUriString = uriStringWithoutFragment(uri);
```
sebjulliand, line 269: `bath`?! 😅
```suggestion
    const baseUriString = uriStringWithoutFragment(uri);
```

src/instantiate.ts
```diff
@@ -120,6 +120,18 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
       }
 
       const uri = getUriFromPath(path, options);
+
+      const existingUri = Tools.findExistingDocumentUri(uri);
+
+      if (existingUri) {
+        const existingOptions = parseFSOptions(existingUri);
+        if (existingOptions.readonly !== options.readonly) {
+          vscode.window.showWarningMessage(`The file is already open in another mode.`);
```
sebjulliand, line 129: Typo
```suggestion
          vscode.window.showWarningMessage(`The file is already opened in another mode.`);
```

sebjulliand (changes requested):
Looks good to me, save for that case-sensitivity issue and some very minor suggestions.
Otherwise, it works great and as expected. I'll approve after your final changes!

worksofliam:
@sebjulliand Also added one more fix. Back to you for review!

worksofliam:
@sebjulliand Fixed a bug.. now it is ready!

PR #1839, opened 2024-02-11
Add option to show hidden files and directories in IFS browser

chrjorgensen:
### Changes

Up until now the IFS browser shows all files and directories, even the hidden ones with a `.` in front of the name.
This PR will add an option to show or hide these files and directories in the IFS browser:

image

image

The option can be changed in the connection settings:

image

- it is on by default to show hidden files as before this change
- an icon is added to the IFS browser navigation menu (replacing 'Remove IFS Shortcut'):

image

- special files and directories are always shown (see constant `ALWAYS_SHOW_FILES` in `ifsBrowser.ts`) to always show `.vscode` and `.gitignore`

### Checklist

* [x] have tested my change


worksofliam:
@chrjorgensen Any chance you can fix the conflict? 😬

chrjorgensen:
@worksofliam Conflict fixed...

src/views/ifsBrowser.ts
```diff
@@ -23,6 +23,10 @@ function isProtected(path: string) {
   return PROTECTED_DIRS.test(path) || instance.getContent()?.isProtectedPath(path);
 }
 
+function alwaysShow(name: string) {
+  return ALWAYS_SHOW_FILES.test(name);
```
worksofliam, line 27: One day, it'd be cool if this was grabbing from config instead of a constant - similar to `grepIgnoreDirs` - not needed for now.

worksofliam:
Initial reaction and usage of the code is awesome. I need to play with it more before I merge.

src/views/ifsBrowser.ts
chrjorgensen: I agree to some extent, thought about making a config value at first - but then, why have a config for that? The user can just show hidden files and directories, if they need to see them...

Can you think of any other files/directories, that should always be shown? The two in the constant were the ones I could think of...

src/views/ifsBrowser.ts
```diff
@@ -153,11 +157,16 @@ class IFSDirectoryItem extends IFSItem {
...
       try {
         const objects = await content.getFileList(this.path, this.sort, handleFileListErrors);
-        const directories = objects.filter(o => o.type === `directory`);
-        const streamFiles = objects.filter(o => o.type === `streamfile`);
+        const directories = objects
+          .filter(o => o.type === `directory`)
+          .filter(o => showHidden || !o.name.startsWith(`.`) || alwaysShow(o.name));
+        const streamFiles = objects
+          .filter(o => o.type === `streamfile`)
+          .filter(o => showHidden || !o.name.startsWith(`.`) || alwaysShow(o.name));
```
sebjulliand, line 169: How about factorizing the filtering here?
```suggestion
    if (content) {
      try {
        const showHidden = instance.getConfig()?.showHiddenFiles;
        const filterIFSFile = (file:IFSFile, type: "directory" | "streamfile") => file.type === type && (showHidden || !file.name.startsWith(`.`) || alwaysShow(file.name));
        const objects = await content.getFileList(this.path, this.sort, handleFileListErrors);
        const directories = objects.filter(f => filterIFSFile(f, "directory"));
        const streamFiles = objects.filter(f => filterIFSFile(f, "streamfile"));
```

src/webviews/settings/index.ts
```diff
@@ -56,6 +56,7 @@ export class SettingsUI {
...
           .addCheckbox(`enableSQL`, `Enable SQL`, `Must be enabled to make the use of SQL and is enabled by default. If you find SQL isn't working for some reason, disable this. If your QCCSID system value is set to 65535, it is recommended that SQL is disabled. When disabled, will use import files where possible.`, config.enableSQL)
           .addCheckbox(`showDescInLibList`, `Show description of libraries in User Library List view`, `When enabled, library text and attribute will be shown in User Library List. It is recommended to also enable SQL for this.`, config.showDescInLibList)
+          .addCheckbox(`showHiddenFiles`, `Show hidden files and directories in IFS browser.`, `When diabled, hidden files and directories (i.e. names starting with '.') will not be shown in the IFS browser, except for special config files.`, config.showHiddenFiles)
```
sebjulliand, line 59: Small typo
```suggestion
          .addCheckbox(`showHiddenFiles`, `Show hidden files and directories in IFS browser.`, `When disabled, hidden files and directories (i.e. names starting with '.') will not be shown in the IFS browser, except for special config files.`, config.showHiddenFiles)
```

sebjulliand (changes requested):
Very good enhancement!
I have a few suggestions to make and a question:
- Shouldn't a button be added to the browser to toggle between hide/show hidden files? I'm quite sure users will either miss the feature of hit `ctrl+h` by mistake and ask why some files disappeared 😅

package.json
```diff
@@ -1973,7 +1990,11 @@
 				{
 					"command": "code-for-ibmi.removeIFSShortcut",
 					"when": "never"
-				}
+				},
+        {
+          "command": "code-for-ibmi.ifs.toggleShowHiddenFiles",
+          "when": "code-for-ibmi:connected && view == ifsBrowser"
```
worksofliam, line 1996: Is this `when` clause in the right place? Even if my focus on was on the IFS Browser, I couldn't get it to show up.

package.json
```diff
@@ -1632,6 +1643,12 @@
 				"key": "ctrl+r",
 				"mac": "cmd+r",
 				"when": "editorLangId == cl"
+			},
+			{
+				"command": "code-for-ibmi.ifs.toggleShowHiddenFiles",
+				"key": "ctrl+h",
+				"mac": "cmd+h",
+        "when": "!editorFocus || editorReadonly"
```
worksofliam, line 1651: This works, but Command+H on Mac is 'hide current window' :joy:

src/views/ifsBrowser.ts
worksofliam: That makes complete sense!

worksofliam (changes requested):
I left some comments which I don't expect any changes for, because I love this feature! I have so many hidden folders I don't care about. But, do see Seb's comments for minor fixes and then assign a review back to me. Thanks!

package.json
chrjorgensen: Any suggestion for another hot-key? I took the one from GNOME Nautilus, it was natural for me... 😉

src/views/ifsBrowser.ts
chrjorgensen: @sebjulliand Did your parents nick-name you "The Typescript Zip Master"?! 😆 

Suggestion committed... 👍

package.json
chrjorgensen: I will remove the command and add an icon in the IFS browser instead.

package.json
chrjorgensen: I removed the hotkey - the user can add it to the settings if needed.

chrjorgensen:
@worksofliam The keybinding has been removed and replaced by an icon in the IFS browser navigation menu, as suggested by @sebjulliand 

Feel free to test and merged, if approved. 🙏

worksofliam (approved):
@chrjorgensen Thanks for taking my shortcut consideration into account! I prefer the icon over the shortcut for sure due to it already being used on Mac.

Works good to me. Thanks!

PR #1842, opened 2024-02-14
Fix removing (temp) IFS directories

chrjorgensen:
### Changes

When removing temp directories in IFS, the `rm` would fail because it won't delete directories without the `-r` flag (recursive):

image

The `-r` flag has been added to the `rm` commands.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -429,7 +429,7 @@ export default class IBMi {
             })
 
           this.sendCommand({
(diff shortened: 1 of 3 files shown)


chrjorgensen:
Oh, my master branch was not updated with the latest changes, so I had to force-push after rebase on master and resolving merge-conflicts... Should be okay now...

worksofliam:
Looks good to me. I am happy with the merge and there is no major change here.

PR #1843, opened 2024-02-14
Move auto sort IFS shortcuts setting to Features tab

chrjorgensen:
### Changes

This PR will move the setting `Sort IFS shortcuts automatically` to the correct tab `Features` - previously it was on the wrong tab `Temporary Data`.

### Checklist

* [x] have tested my change


sebjulliand (approved):
It's indeed a better location for this preference!

PR #1844, opened 2024-02-14
`getTable` API: Make member optional

worksofliam:
### Changes

Makes the member parameter on the `getTable` API optional, and assumes it to be the same as the table if not provided.

It used to work like this before, so not sure when this change was made.


sebjulliand (approved):
Looks good!

PR #1845, opened 2024-02-14
Connection browser enhancements

chrjorgensen:
### Changes

This PR adds some enhancements to the server browser:
- add a sort connection option 

image

- add Rename Connection menu item

image

- remove most connection commands from command palette
- only show Connection Settings in command palette when connected
- remove cached server settings when option is changed to disable

### Checklist

* [x] have tested my change


src/locale/ids/fr.ts
```diff
@@ -45,6 +45,9 @@ export const fr: Locale = {
   'connectionBrowser.connectTo.lastUsed': `Dernière utilisation: {0}`,
   'connectionBrowser.connectTo.title': `Dernières connexions IBM i`,
   'connectionBrowser.connectTo.error': `Utilisez l'Explorateur de Serveur pour sélectionner le serveur auquel se connecter.`,
+  'connectionBrowser.renameConnection.prompt': `Rename connection {0}`,
```
chrjorgensen, line 48: @sebjulliand Could I ask you to translate these three new texts into French? :pray:

worksofliam:
Not part of my review but an issue nonetheless:

If I rename a connection, enter a new name, and get this error:

image

But then if I refresh the view, the name is updated correctly.

chrjorgensen:
@worksofliam Could your error be caused by some "molested" configuration settings in VS Code, where the `name` attribute is not there as expected?

I've seen many examples of this in my own settings - like the `"button" = "save"`example, which is also seen in other users settings in some of the issues...

A connection has values in up to four different places in the settings / storage, and currently there is a danger in the code, that only part of them gets the new name for the connection. So be careful when testing - do not test on important connections!

I will add a try-catch block to catch any errors and not update anything if errors occurred - stay tuned...

chrjorgensen:
@worksofliam Error handling has been added and you should get more error information - and not get any connection changed if error occurred. (Tested this with `settings.json` opened - cool to see the file being updated by the rename command!)

@sebjulliand Now even more messages to translate to French... 😉

src/views/ConnectionBrowser.ts
```diff
@@ -75,6 +75,70 @@ export class ObjectBrowserProvider {
...
+
+              // Then get the password key
+              const secret = await context.secrets.get(`${server.name}_password`);
+
+              // No errors - update the settings.
+              await GlobalConfiguration.set(`connectionSettings`, connectionSettings);
+              await GlobalConfiguration.set(`connections`, connections);
+              if(cachedConnectionSettings) {
+              GlobalStorage.get().setServerSettingsCache(newName, cachedConnectionSettings);
+              GlobalStorage.get().deleteServerSettingsCache(server.name);
```
sebjulliand, line 119: Indentation
```suggestion
                GlobalStorage.get().setServerSettingsCache(newName, cachedConnectionSettings);
                GlobalStorage.get().deleteServerSettingsCache(server.name);
```

src/views/ConnectionBrowser.ts
```diff
@@ -75,6 +75,70 @@ export class ObjectBrowserProvider {
...
+            } catch (e: any) {
+              vscode.window.showErrorMessage(
+                t(`connectionBrowser.renameConnection.error`, server.name, e.message || String(e)));
+            }
+          }
+        }
+      }),
+
+      vscode.commands.registerCommand(`code-for-ibmi.sortConnections`, async () => {
+        let connections = GlobalConfiguration.get<ConnectionData[]>(`connections`) || [];
```
sebjulliand, line 136: Use `const` whenever you can
```suggestion
        const connections = GlobalConfiguration.get<ConnectionData[]>(`connections`) || [];
```

sebjulliand (changes requested):
Looks good and works fine so far! Thank you!
Just some very minor style change were requested, but that's just me being me.

src/views/ConnectionBrowser.ts
chrjorgensen: Agreed - just me getting confused that you can change a `const`... :roll_eyes:

sebjulliand: The object state can change and that's fine.It's reference cannot, giving you the assurance that you'll always manipulate the same object and it won't be replaced somehow.

sebjulliand (approved):
Brilliant! Let's merge!

PR #1848, opened 2024-02-16
Trim whitespaces in object filter's library and object field

sebjulliand:
### Changes
Fixes https://github.com/codefori/vscode-ibmi/issues/1847

Whitespaces in object filter's `library` and `object` fields were not trimmed, resulting in the filtering not working correctly.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/webviews/filters/index.ts
+++ b/src/webviews/filters/index.ts
@@ -66,22 +66,21 @@ export async function editFilter(filter?: ConnectionConfiguration.ObjectFilters,
         //In case we need to play with the data
         switch (key) {
           case `name`:
-          case `filterType`:
-          case `library`:
+          case `filterType`:          
             data[key] = String(data[key]).trim();
             break;
           case `types`:
             data[key] = String(data[key]).split(`,`).map(item => item.trim().toUpperCase()).filter(item => item !== ``);
             break;
+          case `library`:
           case `object`:
             data[key] = (String(data[key].trim()) || `*`)
               .split(',')
-              .map(o => useRegexFilters ? o : o.toLocaleUpperCase())
+              .map(o => useRegexFilters ? o : o.toLocaleUpperCase().trim())
               .filter(Tools.distinct)
               .join(",");
             break;
-          case `member`:
-          case `member`:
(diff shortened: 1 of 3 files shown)


src/webviews/filters/index.ts
```diff
@@ -66,22 +66,21 @@ export async function editFilter(filter?: ConnectionConfiguration.ObjectFilters,
...
             break;
           case `types`:
             data[key] = String(data[key]).split(`,`).map(item => item.trim().toUpperCase()).filter(item => item !== ``);
             break;
+          case `library`:
           case `object`:
             data[key] = (String(data[key].trim()) || `*`)
               .split(',')
-              .map(o => useRegexFilters ? o : o.toLocaleUpperCase())
+              .map(o => useRegexFilters ? o : o.toLocaleUpperCase().trim())
```
chrjorgensen, line 79: @sebjulliand Why trimming filters when saving the filter? The filter works great with or without blanks in-between, and not trimming the filters allows the user to space the filter out for better readability...

chrjorgensen (changes requested):
@sebjulliand Code looks good to me - only one question... 😉

src/webviews/filters/index.ts
chrjorgensen: Maybe just trimming before splitting would be sufficient - then the user can have spaces in-between, and the filter would still be left-justified without trailing spaces.

worksofliam:
Can we get a test case for this? :)

src/webviews/filters/index.ts
sebjulliand: That may indeed be better! Thanks Christian!

sebjulliand:
@chrjorgensen filters are not trimmed on saving anymore
@worksofliam I added a test for filters with whitespaces

chrjorgensen (approved):
@sebjulliand You rock :rocket: :+1: - and work in the weekend! :-1: :wink: 

But so do I, so it's okay... :laughing: 

Approved and merged. We should think of making a prerelease for users to avoid this error...

PR #1850, opened 2024-02-18
Add tooltip for source files showing member count, source length and CCSID

chrjorgensen:
### Changes

This PR will add a tooltip on source files in the Object Browser. The tooltip contains
- object description
- member count
- source length
- CCSID

image

And all tooltips now have a HTML table containing the information, allowing for advanced formatting.

### Checklist

* [x] have tested my change


chrjorgensen:
All tooltip texts has been changed, so the text with tabs are combined and the tooltip value is a value in the text.
This allows for individual formatting between languages and gives a cleaner look..

chrjorgensen:
@sebjulliand There are some new texts for the source file tooltips - could you... 😉

src/locale/ids/fr.ts
```diff
@@ -215,6 +211,14 @@ export const fr: Locale = {
   'objectBrowser.doSearchInSourceFile.notFound': `Aucun résultat trouvé pour '{0}' dans {1}.`,
   'objectBrowser.doSearchInSourceFile.noMembers': `Aucun membre à rechercher.`,
   'objectBrowser.doSearchInSourceFile.errorMessage': `Erreur lors de la recherche de membres source: {0}`,
+  'objectBrowser.sourceFile.tooltip.text':'Text:\t\t{0}',
+  'objectBrowser.sourceFile.tooltip.members': 'Members:\t{0}',
+  'objectBrowser.sourceFile.tooltip.length': 'Length:\t\t{0}',
+  'objectBrowser.sourceFile.tooltip.CCSID': 'CCSID:\t\t{0}',
```
sebjulliand, line 217: 📖
```suggestion
  'objectBrowser.sourceFile.tooltip.text':'Texte:\t\t{0}',
  'objectBrowser.sourceFile.tooltip.members': 'Membres:\t{0}',
  'objectBrowser.sourceFile.tooltip.length': 'Longueur:\t{0}',
```

src/locale/ids/fr.ts
```diff
@@ -116,6 +109,9 @@ export const fr: Locale = {
   'ifsBrowser.doSearchInStreamfiles.progressMessage': `'{0}' dans {1}.`,
   'ifsBrowser.doSearchInStreamfiles.noResults': `Aucun résultat trouvé pour '{0}' dans {1}.`,
   'ifsBrowser.doSearchInStreamfiles.errorMessage': `Erreur lors de la recherche de fichiers.`,
+  'ifsBrowser.file.tooltip.size': `Taille:\t\t\t{0}`,
```
sebjulliand, line 112: Remove a tab to align
```suggestion
  'ifsBrowser.file.tooltip.size': `Taille:\t\t{0}`,
```

src/locale/ids/fr.ts
```diff
@@ -215,6 +211,14 @@ export const fr: Locale = {
   'objectBrowser.doSearchInSourceFile.notFound': `Aucun résultat trouvé pour '{0}' dans {1}.`,
   'objectBrowser.doSearchInSourceFile.noMembers': `Aucun membre à rechercher.`,
   'objectBrowser.doSearchInSourceFile.errorMessage': `Erreur lors de la recherche de membres source: {0}`,
+  'objectBrowser.sourceFile.tooltip.text':'Text:\t\t{0}',
+  'objectBrowser.sourceFile.tooltip.members': 'Members:\t{0}',
+  'objectBrowser.sourceFile.tooltip.length': 'Length:\t\t{0}',
+  'objectBrowser.sourceFile.tooltip.CCSID': 'CCSID:\t\t{0}',
+  'objectBrowser.member.tooltip.text':'Texte:\t\t\t{0}',
+  'objectBrowser.member.tooltip.lines': 'Lignes:\t\t\t{0}',
```
sebjulliand, line 219: Fixed alignment
```suggestion
  'objectBrowser.member.tooltip.lines': 'Lignes:\t\t{0}',
```

src/locale/ids/fr.ts
```diff
@@ -215,6 +211,14 @@ export const fr: Locale = {
   'objectBrowser.doSearchInSourceFile.notFound': `Aucun résultat trouvé pour '{0}' dans {1}.`,
   'objectBrowser.doSearchInSourceFile.noMembers': `Aucun membre à rechercher.`,
   'objectBrowser.doSearchInSourceFile.errorMessage': `Erreur lors de la recherche de membres source: {0}`,
+  'objectBrowser.sourceFile.tooltip.text':'Text:\t\t{0}',
+  'objectBrowser.sourceFile.tooltip.members': 'Members:\t{0}',
+  'objectBrowser.sourceFile.tooltip.length': 'Length:\t\t{0}',
```
sebjulliand, line 216: 📖
```suggestion
  'objectBrowser.sourceFile.tooltip.text':'Texte:\t\t{0}',
  'objectBrowser.sourceFile.tooltip.members': 'Membres:\t{0}',
  'objectBrowser.sourceFile.tooltip.length': 'Longueur:\t{0}',
```

sebjulliand (changes requested):
Very nice! You can never get enough information 😄
Find the missing FR translations and fixes in the suggestions. Then it's a merge!

chrjorgensen:
@sebjulliand Thanks - and the CCSID was for you! 😘😂

src/locale/ids/fr.ts
chrjorgensen: The current number of tabs makes the columns align - because the text length is different:

image

Is that not what you get in your instance of VS Code?

src/locale/ids/fr.ts
chrjorgensen: Same as above - number of tabs should be different between the tooltip lines:

image

chrjorgensen:
@sebjulliand PR is ready for final review - and merge, if approved... :pray:

src/locale/ids/fr.ts
sebjulliand: image
Sad, right? 😅

src/locale/ids/fr.ts
sebjulliand: image

sebjulliand (approved):
Good!
As for the alignment being different between you and me, that may be related to our display configuration. There may not be much we can do about it 🙄 So feel free to merge whenever you want.

chrjorgensen:
@sebjulliand The alignment in the tooltips for sourcefiles and members has been fixed - by changing the tooltip into a HTML table!

Actually didn't know it was possible until I found this: 
https://github.com/microsoft/vscode/blob/6d2920473c6f13759c978dd89104c4270a83422d/src/vs/base/browser/markdownRenderer.ts#L296

💣

chrjorgensen:
@sebjulliand @worksofliam Sorry, I went berserk in tooltips! 😆

sebjulliand (changes requested):
It looks good, no question asked!
But having the formatting in the translations looks a bit wrong 😅
It would be better, IMHO, to have a tool function that generates a simple table for you and reuse it where you need the tooltips.

Something along these lines:
```typescript
export function generateTooltipTable(header:string, rows: Record<string, any>){
  return /* html */ `<table>
      <thead>${header}</thead>
      ${Object.entries(rows).map(([key, value]) => `<tr><td>${t(key)}</td><td>${value}</td></tr>`)}      
    <table>`;
  }
```

Used like that:
```typescript
this.tooltip = new vscode.MarkdownString(Tools.generateTooltipTable(this.path, {
  text:sourceFile.text,
  members:sourceFile.memberCount,
  length:sourceFile.sourceLength,
  CCSID: sourceFile.CCSID
}));
```

Yes, the translation keys (text, member, length...) are not tied to a specific domain in this case, but it may be something we should think about too, having generic translations.

What do you think?

chrjorgensen:
@sebjulliand I agree - will make the necessary changes in a new commit...

The translation keys were actually generic before my changes - so they will go back to become that.

This has been a bit trial-and-error and going back-and-forth... I should probably squash the commits and force-push the final commits. Right?

sebjulliand:
@chrjorgensen Trying to hide your mess under the carpet I see, hmm? 😁
Just kidding, squash and force push as you please; make it clean!

chrjorgensen:
@sebjulliand Cleaned up my mess - don't look under the carpet! 😄 

Final review - then we can hopefully merge... 😍

PR #1851, opened 2024-02-18
Ignore error copying empty member

chrjorgensen:
### Changes

This PR will fix the error occurring, when copying an empty member.

Currently the copy ends with the following errors:

image

and the member list is not updated, even though the new member is created (as empty).

The PR changes the code to only report errors if CPF2869 (Empty member &3 in file &1 in library &2 is not copied) is not among the errors. So now copying an empty member will act exactly like copy members with lines.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/Tools.ts
+++ b/src/api/Tools.ts
@@ -239,7 +239,8 @@ export namespace Tools {
     }) as IBMiMessage);
     return {
       messages,
-      findId: id => messages.find(m => m.id === id)
+      findId: id => messages.find(m => m.id === id),
+      length: messages.length
     }
   }
 
--- a/src/api/Tools.ts
+++ b/src/api/Tools.ts
@@ -239,8 +239,7 @@ export namespace Tools {
     }) as IBMiMessage);
     return {
       messages,
-      findId: id => messages.find(m => m.id === id),
-      length: messages.length
+      findId: id => messages.find(m => m.id === id)
     }
   }
 
--- a/src/views/objectBrowser.ts
+++ b/src/views/objectBrowser.ts
@@ -540,8 +540,8 @@ export function initializeObjectBrowser(context: vscode.ExtensionContext) {
               noLibList: true
             })
 
-            const messages = Tools.parseMessages(copyResult.stderr);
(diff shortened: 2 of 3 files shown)


chrjorgensen:
@sebjulliand I added a property `length` to `IBMiMessages` - will it cause any conflicts in other places parsing messages from stderr? I haven't seen any in my tests, but just want to be share (always anxious when changing types... :laughing:)

src/api/Tools.ts
```diff
@@ -239,7 +239,8 @@ export namespace Tools {
     }) as IBMiMessage);
     return {
       messages,
-      findId: id => messages.find(m => m.id === id)
+      findId: id => messages.find(m => m.id === id),
+      length: messages.length
```
sebjulliand, line 243: Why bother with an extra `length` field when we already expose `messages`? We can simply call `messages.length`.

src/views/objectBrowser.ts
```diff
@@ -541,7 +541,7 @@ export function initializeObjectBrowser(context: vscode.ExtensionContext) {
             })
 
             const messages = Tools.parseMessages(copyResult.stderr);
-            if (messages.findId(`CPF2869`)) {
+            if (messages.length && !messages.findId(`CPF2869`)) {
```
sebjulliand, line 544: Following up with the above comment, we can simply access the length `property` from `messages`.
```suggestion
            if (messages.messages.length && !messages.findId(`CPF2869`)) {
```

sebjulliand (changes requested):
The change works fine, thanks for that!
See my comment about `length`. It's harmless to add it, but is it useful?

src/api/Tools.ts
chrjorgensen: See reply in other comment...

src/views/objectBrowser.ts
chrjorgensen: My lack of knowledge... couldn't make `.length` work, because I didn't duplicate `messages.`...

Of course you're right - I'll fix it.

sebjulliand (approved):
Perfect 🥰

PR #1852, opened 2024-02-18
Add object commands to sourcefiles in Object browser

chrjorgensen:
### Changes

This is a PR to add the object commands, like `Change Object Description`, to sourcefiles as well:

image

Please comment on this - if it's a bad idea or whatever you think. :smiling_face_with_tear: 

### Checklist

* [x] have tested my change


chrjorgensen:
I made another ligther solution for the object browser, after some advice from @sebjulliand ... :pray:

sebjulliand (changes requested):
It works fine!
Now, since the interface `SourcePhysicalFileItem` is strictly identical to `ObjectItem`, it may be better to remove it and let `ObjectBrowserSourcePhysicalFileItem` implements `ObjectItem` instead. It would make more sense than having a duplicated interface (and it's just logical since a source physical file IS an Object).
What do you think?

chrjorgensen:
@sebjulliand

- I removed `SourcePhysicalFileItem` as you suggested.
- I also fixed Run Action for sourcefiles - had to set the `resourceUri` on the browser item.
- I also cleaned up my mess - again...! 😄 

Will you do a final review?

worksofliam:
@chrjorgensen I checked out this branch but I am getting these reference errors which is not letting me build!

image

chrjorgensen:
@worksofliam How about now? I force-pushed my local branch after rebasing on newest master branch and fixing conflicts... 🙏

worksofliam (approved):
Works as expected. I am able to use typically object context items on source files now!

worksofliam:
Just retested and it's looking good!!

image

PR #1866, opened 2024-02-22
Store that CL component was installed after installing

worksofliam:
### Changes

Solves a bug that was causing the CL installer message to always a appear after connecting for the first time, and choosing to ignore the installer.

Now, if you install on the second connect, you will no longer get the message.

### Checklist

* [x] have tested my change


chrjorgensen (approved):

chrjorgensen:
@worksofliam The issue is (also?) that even if you answer `Yes` to the CL install prompt and the install succeeds, it will reappear on the following connects. Maybe the PR description should be changed to reflect this?

The fix works as expected. Approved.

PR #1867, opened 2024-02-22
Speed up logic to validate member and extension exists

chrjorgensen:
### Changes

This PR is an alternative to PR #1826 and will keep the check for member existence.

Instead of using QSYS2.SYSPARTITIONSTAT to retrieve the member information, which could be slow on small server with many, many members, the code introduces a new SQL table function `<tempLib>.GETMBRINFO` to get the information for the specific member. API `QUSRMBRD` is used by `GETMBRINFO` behind the scenes.

### Checklist

* [x] have tested my change


src/languages/sql/getmbrinfo.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    content!.writeStreamfile(tempSourcePath, getSource(config.tempLibrary)).then(() => {
+      connection.runCommand({
+        command: `RUNSQLSTM SRCSTMF('${tempSourcePath}') COMMIT(*NONE) NAMING(*SQL)`,
+        cwd: `/`,
+        noLibList: true
+      })
+      .then((result) => {
+        if (result.code === 0) {
+          connection.remoteFeatures[`GETMBRINFO.SQL`] = `${config.tempLibrary}.GETMBRINFO`;
```
worksofliam, line 24: Similar to what I have done in #1866. you will need to store into the cache that this has been installed otherwise you will keep getting the popup:

https://github.com/codefori/vscode-ibmi/pull/1866/files#diff-3c2633023201c776a18178afef7bde70914ff9c816de4afe480dc2c920af3211R52

worksofliam (changes requested):
Looks okay. Initial glance is that it will continue to ask about installation since you don't update the settings cache for the connection. I do this in #1866 with a new API. Maybe that needs to be reviewed/merged first so you can use that API here?

src/api/IBMiContent.ts
```diff
@@ -631,6 +631,45 @@ export default class IBMiContent {
     }
   }
 
+  /**
+   *
+   * @param filter: the criterias used to list the members
+   * @returns
+   */
+  async getMemberInfo(library: string, sourceFile: string, member: string): Promise<IBMiMember | undefined> {
```
worksofliam, line 639: I think throwing `undefined` if the member does not exist makes sense. But, I think this API should throw an error if the feature (`GETMBRINFO.SQL`) is not installed.

src/instantiate.ts
```diff
@@ -430,32 +431,21 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
...
-                  return resultSet.length !== 1 ? {} :
-                    {
-                      base: `${resultSet[0].MEMBER}.${resultSet[0].TYPE}`,
-                      name: `${resultSet[0].MEMBER}`,
-                      ext: `${resultSet[0].TYPE}`,
-                    }
-                });
-                if (!fullMember) {
-                  vscode.window.showWarningMessage(`Member ${lib}/${file}/${member.base} does not exist.`);
+                const memberInfo = await content!.getMemberInfo(lib, file, member.name);
```
worksofliam, line 438: Handle if `getMemberInfo` ends with a crash.

worksofliam (changes requested):
It works very well. You are right, the performance is so amazing. Couple of things:

1. The SQL function is being created every time I connect to a system. Similar to CL, we need to store that we are installing it after it is installed.
2. We need to handle it better if the installation failed at runtime. Returning the member, undefined or crashed to represent the different states would be good in my opinion.

worksofliam:
@sebjulliand I have requested your review here will Christian is away. No rush whatsoever.

chrjorgensen:
@sebjulliand Seems like you and @Wright4i are the only ones left, who hasn't made commits and can review...? 😉

chrjorgensen:
@Wright4i @sebjulliand Merge conflicts have been resolved - ready for review...

worksofliam:
@chrjorgensen I am going to give this another review today and I will approve and merge if acceptable. Thanks!

worksofliam (approved):
@chrjorgensen Great works. Made one more commit to cleanup unused variables and added some small comments. Let's go!

PR #1869, opened 2024-02-23
Fix object default tooltips still showing

chrjorgensen:
### Changes

This PR will remove the default toolip being shown for non-sourcefile objects in the Object browser:

image

Seems like if the tree has items with markdown tooltips, all items must use markdown tooltips...

### Checklist

* [x] have tested my change


sebjulliand (approved):
LGTM!

PR #1871, opened 2024-02-23
Allow open members in object filters for *FILE

chrjorgensen:
### Changes

This PR is an short, but effective addition to PR #1780 and will make *FILE object filters behave like filters for *SRCPF or *ALL, allow browsing of sourcefiles:

image

### Checklist

* [x] have tested my change


sebjulliand (approved):
Of course! Good catch!

PR #1872, opened 2024-02-23
Add tooltip for objects in object browser

chrjorgensen:
### Changes

This PR is a continuation of the tooltips changes - it adds a tooltip to objects in the object browser, when the object is not a source file:

image

### Checklist

* [x] have tested my change

### DIFF
--- a/src/locale/ids/fr.ts
+++ b/src/locale/ids/fr.ts
@@ -31,8 +31,8 @@ export const fr: Locale = {
   'modified': 'Modifié',
   'owner': 'Propriétaire',
   'type': 'Type',
-  'attribute': 'Attribute',
-  'created_by': 'Created by',
+  'attribute': 'Attribut',
+  'created_by': 'Créé par',
   // Sandbox:
   'sandbox.input.user.title': `Nom d'utilisateur`,
   'sandbox.input.user.prompt': `Entrez le nom d'utilisateur pour {0}`,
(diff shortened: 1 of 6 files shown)


src/locale/ids/fr.ts
```diff
@@ -30,6 +30,9 @@ export const fr: Locale = {
   'size': 'Taille',
   'modified': 'Modifié',
   'owner': 'Propriétaire',
+  'type': 'Type',
```
chrjorgensen, line 33: @sebjulliand We need your translation skills here! 😍

src/locale/ids/fr.ts
```diff
@@ -30,6 +30,9 @@ export const fr: Locale = {
...
+  'attribute': 'Attribute',
+  'created_by': 'Created by',
```
chrjorgensen, line 35: @sebjulliand We need your translation skills here! 😉

sebjulliand, line 35: Voilà!
```suggestion
  'type': 'Type',
  'attribute': 'Attribut',
  'created_by': 'Créé par',
```

sebjulliand (approved):
Tested over a few different objects and it works like a charm.
A fine addition! I guess we have big enough tooltips now 😅

I'll approve and let you merge whenever you want, if you'd like others to review too. But I'm fine with it!

chrjorgensen:
@sebjulliand

> I guess we have big enough tooltips now 😅

I remember someone recently telling me "we can't have enough information"... 😉 

Just kidding - I agree and had to think about when deciding which information to include. We really need that FS extension...! 😍

PR #1874, opened 2024-02-24
Fix error copying empty member

chrjorgensen:
### Changes

This PR will fix the change from PR #1851 - the test in the PR was not correct.
Now we check for return code first, and only if error code 1 and messages and not messages
`CPF2869 (Empty member &3 in file &1 in library &2 is not copied.)` and
`CPF2817 (Copy command ended because of error.)`
do we ignore the errors and accept the outcome.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -68,7 +68,7 @@ export default class IBMiContent {
   }
 
   /**
-   * 
+   *
    * @param remotePath Remote IFS path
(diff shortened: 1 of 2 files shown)


sebjulliand (approved):
Works fine. Before the PR, copying a non-empty member would show an error message.
Now copying both an empty and non-empty member works. Good job @chrjorgensen!

Issue #1490, opened 2023-08-18
Show SEU Colors feature is overlaying characters 


m-tyler:
Some of our source have hex code coloring that begins and ends within the comment and non space characters follow.
Example, /*x'21'/M01x'20'12345 PGxx mm/dd/YY Change text                  */
becomes  /*#?/M01?2345 PGxx mm/dd/yy Change text                  */ 
The following characters in example (x'21', x'20', #?, ?) are actually only one character in the source, see bellow for images.

I only write this as I noticed the issue today on a member that another programmer was modifying and I was reviewing the code.   I normally run a personal command, UNCOLOR, to remove these hex values from the source.   So fixing this is a `low priority` for me but I understand this might be more meaningful to others to have fixed.  

Attached is also a test member, seen below. 

<hr />

|Context|Version|
|-|-|
|Code for IBM i version|2.0.2|
|Visual Studio Code version|1.78.2|
|Operating System|win32_x64|

<details>
<summary>Active extensions</summary>

```
Binary Viewer (binary-viewer): 1.1.1
CL (vscode-clle): 1.1.1
Code Spell Checker (code-spell-checker): 2.20.5
Code for IBM i Walkthroughs (vscode-ibmi-walkthroughs): 0.3.1
Db2 for IBM i (vscode-db2i): 0.4.0
ESLint (vscode-eslint): 2.4.2
Emmet (emmet): 1.0.0
Error Lens (errorlens): 3.13.0
Excel Viewer (gc-excelviewer): 4.2.58
Git (git): 1.0.0
Git Base (git-base): 1.0.0
Git Graph (git-graph): 1.30.0
GitHub (github): 0.0.1
GitHub Authentication (github-authentication): 0.0.2
IBM i Notebooks (vscode-ibmi-notebooks): 0.0.6
JSON Language Features (json-language-features): 1.0.0
Merge Conflict (merge-conflict): 1.0.0
Microsoft Account (microsoft-authentication): 0.0.1
Node Debug Auto-attach (debug-auto-launch): 1.0.0
Overtype (overtype): 0.5.0
Partial Diff (partial-diff): 1.4.3
Project Dashboard (vscode-dashboard): 2.6.0
RPGLE (vscode-rpgle): 0.20.2
Rainbow CSV (rainbow-csv): 3.7.0
SQL Formatter VSCode (sql-formatter-vsc): 2.1.0
Text Marker (Highlighter) (text-marker): 1.11.0
[...]

PR #1878, opened 2024-02-27
Update SEU Colors Feature

steph-beneschan-256:
### Changes

Update SEU Colors Feature
* Updated SEUColorProvider.ts to address issue with overlapping characters (#1490) (the code now calculates ranges based on character count rather than byte count)
* Updated the SEUColorProvider.ts code that looks for color code sequences in the source member to use a Map-based function rather than individually check against every possible color code for each byte (also updated SEUColors.ts to provide the aforementioned function)
* When a color code is found, the styling now only applies until another color code is found (or until the line ends)

* Added a function to SEUColors.ts that uses a Map to check whether a given byte sequence matches any of the color sequences
* When reading bytes from a source member, SEUColorProvider.ts now uses the aforementioned Map-based function on each byte rather than individually checking it against every possible color sequence, to save time

### Checklist

* [x] have tested my change

### DIFF
--- a/src/languages/general/SEUColorProvider.ts
+++ b/src/languages/general/SEUColorProvider.ts
@@ -3,7 +3,6 @@ import vscode from 'vscode';
 import { SEUColors } from './SEUColors';
 
 const hidden = vscode.window.createTextEditorDecorationType({
-  //letterSpacing: `-1em`,
   opacity: `0`,
   rangeBehavior: vscode.DecorationRangeBehavior.ClosedClosed,
 });
--- a/src/languages/general/SEUColorProvider.ts
+++ b/src/languages/general/SEUColorProvider.ts
@@ -43,37 +43,72 @@ export namespace SEUColorProvider {
           colorDecorations[name] = [];
         });
 
-        // Find the lines and the bytes and all that...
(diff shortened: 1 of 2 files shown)


worksofliam:
@steph-beneschan-256 This PR is awesome. Any chance you can share the raw hex of a member that you tested this with? I'd like to be able to test this but I have no test data.

steph-beneschan-256:
Hello @worksofliam ! Here is the hex dump from the member that I used: [COLORTEST.txt](https://github.com/codefori/vscode-ibmi/files/14428733/COLORTEST.txt)

worksofliam:
@steph-beneschan-256 What is the record length for that source file you used? I am going to write a test case to create that member with an example.

worksofliam:
@steph-beneschan-256 Is it possible you could get me a hex dump to that member using SQL instead?

```sql
select hex(srcdta) from lib.sourcefile
```

You may have to create an alias to the member to read it with SQL.

worksofliam:
We are removing SEU colour support in 3.0.0, so I will close this PR. Thanks for you work, nonetheless!

https://github.com/orgs/codefori/discussions/2057

PR #1880, opened 2024-02-27
Speed up library list loading when SQL is disabled

sebjulliand:
### Changes
When SQL is disabled, loading the library list items take a significant amount of time because it runs `DSPOBJD OBJ(QSYS/*ALL)` to load the librarys' description.
This PR makes it run one `DSPOBJD` for each library which speeds up the loading of the Library List view.

The other possible fix for this is to disregard `config.enableSQL` entirely and always use `OBJECT_STATISTICS`.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -398,7 +398,7 @@ export default class IBMiContent {
       `;
       results = await this.runSQL(statement);
     } else {
-      results = await this.getQTempTable([`CALL QSYS2.QCMDEXC('DSPOBJD OBJ(QSYS/*ALL) OBJTYPE(*LIB) DETAIL(*TEXTATR) OUTPUT(*OUTFILE) OUTFILE(QTEMP/LIBLIST)')`], "LIBLIST");
+      results = await this.getQTempTable(libraries.map(library => `@DSPOBJD OBJ(QSYS/${library}) OBJTYPE(*LIB) DETAIL(*TEXTATR) OUTPUT(*OUTFILE) OUTFILE(QTEMP/LIBLIST) OUTMBR(*FIRST *ADD)`), "LIBLIST");
       if (results.length === 1 && !results[0].ODOBNM?.toString().trim()) {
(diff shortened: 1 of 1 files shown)


chrjorgensen (approved):
@sebjulliand LGTM - same result as before the change.

But I don't see any speed difference with my 6 libraries in the library list. Have we had any reports of slow library list view?

sebjulliand:
> @sebjulliand LGTM - same result as before the change.
> 
> But I don't see any speed difference with my 6 libraries in the library list. Have we had any reports of slow library list view?

I did not report it, but it's been bothering me lately as I kept seeing the view loading for too long on one of the LPAR I work on.
But this is a development LPAR, and it has tons of libraries, so that must be why!

sebjulliand:
For the record: that development LPAR has 11010 libraries 😅 And since my profile has `*ALLOBJ`, I can list and access them all.
I tried on another LPAR with 845 libraries and in this case, the change is not noticeable.

But thanks for testing and validating, much appreciated!

worksofliam:
That's a lot of `DSPOBJD`! Makes sense, but yes, I wondering if using `OBJECT_STATISTICS` would be better - esp. since we don't need to worry about the CCSID stuff now :)

worksofliam (approved):

chrjorgensen:
> For the record: that development LPAR has 11010 libraries

Wauw - you should do some cleanup! 😆 

I tested in one of our LPARs with 1227 libraries and didn't see any difference after the change. I remember once thinking about why all libraries were included, but back then we could not send multiple commands at once, so I decided that it was better keep getting all libraries with one command / round-trip than getting one at a time and having multiple commands and round-trips. Your excellent change allowing us to run multiple commands in only one round-trip now made this change possible. Thanks!

sebjulliand:
Thank you guys!
I guess we'll be able to keep only the `OBJECT_STATISTICS` part for `3.0.0`.

PR #1882, opened 2024-02-27
Feature/save last location

sebjulliand:
### Changes
This PR makes the download actions save the last selected download location which gets reused when a new download location must be picked.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/Configuration.ts
+++ b/src/api/Configuration.ts
@@ -1,4 +1,4 @@
-
+import os from "os";
 import * as vscode from 'vscode';
 import { DeploymentMethod } from '../typings';
 import { FilterType } from './Filter';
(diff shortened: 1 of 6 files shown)


chrjorgensen (approved):
@sebjulliand LGTM - works as expected. 👍 

One nice addition could be having the location stored in the connection - so I could close VS Code and return tomorrow and continue downloading into the same location... If no previous location was set, then use the local home directory...

sebjulliand:
Good idea @chrjorgensen ! Give it another try!

chrjorgensen (approved):
Works as expected - really nice to have the download location saved in the connection. 👍

chrjorgensen:
@sebjulliand My issue yesterday was that I had marked multiple members and selected `Download...` - but only one member was downloaded... I may have been caught by the "multi-object-selection" fever dragged into believing that multi-selected members would all be downloaded in one operation, like drag'n'drop into IFS browser, but that if of course not (yet) coded... 😍 

Would it be easy to implement - or should we save that for another PR? 😃

sebjulliand:
@chrjorgensen, that's right!
Let's take that to another PR - because I see two things that needs to be adapted for the multi-selection:
- Download multiple files
- Download directories

I'll merge this one and put the enhancements in a new PR 😄

PR #1884, opened 2024-02-28
Allow multiple IFS files/folders download

sebjulliand:
### Changes
This PR enables to download multiple files and/or folders at once from the IFS explorer.
A progress will now be shown during the download to track download progress.

If one file is selected for download, then a file selection dialog shows up to select the download location.
If a folder or multiple files/folders are selected, then the folder selection dialog will show up instead. Then, any file or folder that will be overwritten will show a modal warning to validate its remplacement.

The PR also fixes the `reduceIFSPath` function that didn't filter out directories whose parent was in the selection as well.

### Checklist
* [x] have tested my change


chrjorgensen (approved):
@sebjulliand Brilliant - works perfectly! :rocket:

sebjulliand:
Thanks for the translations @chrjorgensen !

chrjorgensen:
Merged into master => will be included in v2.8.0 => milestone changed to 2.8.0.

PR #1888, opened 2024-02-29
Allow multi members download

sebjulliand:
### Changes
This PR allows to download multiple source members at once.
It is possible to select both physical source files and members to download. Selecting a physical source file will download every member it contains.
If a member exist in the local download directory, a dialog will ask what to do: overwrite, overwrite all, skip, skip all or cancel.

@chrjorgensen, you're up for the Danish translations 🙏🏻😄

### Checklist
* [x] have tested my change


worksofliam:
@sebjulliand Is there any chance you can put together a list of different things to test. From your PR, it sounds like only a few:

1. Test downloading a single member
2. Test downloading multiple members in a single source file
3. Test downloading multiple members in two different sources file
4. Test downloading an entire source file
5. Test downloading two source files

Anything I am missing?

sebjulliand:
> @sebjulliand Is there any chance you can put together a list of different things to test. From your PR, it sounds like only a few:
> 
>     1. Test downloading a single member
> 
>     2. Test downloading multiple members in a single source file
> 
>     3. Test downloading multiple members in two different sources file
> 
>     4. Test downloading an entire source file
> 
>     5. Test downloading two source files
> 
> 
> Anything I am missing?

I would add:
6. Test downloading a mix of members and source files

chrjorgensen (changes requested):
@sebjulliand Awesome work, Mate! Brilliant code... Love it! :heart_eyes: 

Only some few issues:

- I downloaded a file and got all members as expected. Then I download one of the members into the same directory, and after accepting overwriting, it starts download but ends with this error:

image

- And when downloading the member, the file dialog has a type of `Source member` and does not show the existing file:

image

- When I download the source file to same directory, the buttons in the "File exists" dialog are a bit out of order: `Skip all` before `Skip`, `Overwrite All` before `Overwrite` and `Cancel` before `Overwrite All`. To me the logical order would be `Skip`, `Skip All`, `Overwrite`, `Overwrite All`, `Cancel`... Can this be fixed?

image

- the line endings of the downloaded files are always CRLF (the default from CPYTOSTMF). I try to avoid M$' way of doing things and always use LF instead of CRLF, as is also the system line ending in my Debian Linux. Should this be selectable or based on the client default? Is it just me? Probably an old issue, this...


But the functionality is there and works - and I love your work! :clap:

chrjorgensen:
A new issue was found:

I opened a filter with multiple libraries, opened the first library and accidentally selected this together with two source files and chose `Download...`. After selecting the download directory, I got this error:

image

There is no download context menu item for the library, so I would expect it to be excluded (filtered out) from the download selection and only have the source files with members. This probably also applies to other object types being selected with the source file(s).

After de-selecting the library, I could download the two source files with all their members without problem.

sebjulliand:
Thanks for translations and the review @chrjorgensen !
I fixed all of the reported issues, save for the action order in the dialog that seems to be managed by the OS window manager.

sebjulliand:
@chrjorgensen there you go 😊

chrjorgensen (approved):
@sebjulliand You're the best! Great feature, well executed! 😍 

"Signed, Sealed, Delivered" by @sebjulliand 🎵 

Approved - merge when appropriate. 🎉

sebjulliand:
> @sebjulliand You're the best! Great feature, well executed! 😍
> 
> "Signed, Sealed, Delivered" by @sebjulliand 🎵
> 
> Approved - merge when appropriate. 🎉

Much obliged!

chrjorgensen:
Merged into master => will be included in v2.8.0 => milestone changed to 2.8.0.

PR #1895, opened 2024-02-29
Changes/debugger_certs

worksofliam:
### Changes

High level API:

* Adds new `usingBash` method to the `IBMi` class

Debug specific:

* [x] Changes default location for debugger certificates out of ProdData
* [x] Logic written to support uses who are already using the legacy debug certificate directory (inside ProdData)
* [x] Require bash setup before generating certificates
* [x] No longer use a CA when generating certificates
* [x] Documentation PR created and attached to this PR: https://github.com/codefori/docs/pull/27

The walkthroughs need to be updated too, but that will be done in another PR after this as we plan on migrating all walkthroughs to this extension.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/debug/certificates.ts
+++ b/src/api/debug/certificates.ts
@@ -28,7 +28,7 @@ function resolveHostnameToIP(hostName: string): Promise<string | undefined> {
   });
 }
 
-async function getExtFileConent(host: string, connection: IBMi) {
+async function getExtFileContent(host: string, connection: IBMi) {
   const ipRegexExp = /^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$/gi;
   let hostname = undefined;
   let ipAddr = undefined;
@@ -90,7 +90,7 @@ export async function remoteServerCertExists(connection: IBMi, legacy = false) {
  */
 export async function setup(connection: IBMi) {
   const host = connection.currentHost;
-  const extFileContent = await getExtFileConent(host, connection);
+  const extFileContent = await getExtFileContent(host, connection);
 
   if (!connection.usingBash()) {
     if (connection.remoteFeatures[`bash`]) {
--- a/src/api/debug/certificates.ts
+++ b/src/api/debug/certificates.ts
@@ -11,7 +11,7 @@ const clientCertName = `debug_service.crt`;
 export const LEGACY_CERT_DIRECTORY = `/QIBM/ProdData/IBMiDebugService/bin/certs`;
 export const DEFAULT_CERT_DIRECTORY = `/QIBM/UserData/IBMiDebugService/certs`;
 
-function getRemoteCertDirectory(connection: IBMi) {
+export function getRemoteCertDirectory(connection: IBMi) {
   return connection.config?.debugCertDirectory!;
 }
 
--- a/src/api/debug/certificates.ts
+++ b/src/api/debug/certificates.ts
@@ -113,11 +113,9 @@ export async function setup(connection: IBMi) {
   }
 
   const commands = [
-    `openssl genrsa -out debug_service_ca.key 2048`,
-    `openssl req -x509 -new -nodes -key debug_service_ca.key -sha256 -days 1825 -out debug_service_ca.pem -subj '/CN=${host}'`,
     `openssl genrsa -out debug_service.key 2048`,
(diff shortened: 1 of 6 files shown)


src/api/debug/certificates.ts
```diff
@@ -96,12 +104,18 @@ export async function setup(connection: IBMi) {
...
+  }
+
   const commands = [
-    `openssl genrsa -out debug_service_ca.key 2048`,
-    `openssl req -x509 -new -nodes -key debug_service_ca.key -sha256 -days 1825 -out debug_service_ca.pem -subj '/CN=${host}'`,
     `openssl genrsa -out debug_service.key 2048`,
     `openssl req -new -key debug_service.key -out debug_service.csr -subj '/CN=${host}'`,
-    `openssl x509 -req -in debug_service.csr -CA debug_service_ca.pem -CAkey debug_service_ca.key -CAcreateserial -out debug_service.crt -days 1095 -sha256 -sha256 -req -extfile <(printf "${extFileContent}")`,
+    `openssl x509 -req -in debug_service.csr -signkey debug_service.key -out debug_service.crt -days 1095 -sha256 -sha256 -req -extfile <(printf "${extFileContent}")`,
     `openssl pkcs12 -export -out debug_service.pfx -inkey debug_service.key -in debug_service.crt -password pass:${host}`
```
sebjulliand, line 119: A couple of suggestions here:
- Removed a seemingly redundant -sha256 flag
- Delete unused file when the pfx is generated; looks like only the pfx is required to run the service. I removed them manually and it was ok.
- Give read-only access to everyone (better than trusting the default access, which may depend on an unexpected `umask`)
```suggestion
    `openssl genrsa -out debug_service.key 2048`,
    `openssl req -new -key debug_service.key -out debug_service.csr -subj '/CN=${host}'`,
    `openssl x509 -req -in debug_service.csr -signkey debug_service.key -out debug_service.crt -days 1095 -sha256 -req -extfile <(printf "${extFileContent}")`,
    `openssl pkcs12 -export -out debug_service.pfx -inkey debug_service.key -in debug_service.crt -password pass:${host}`,
    `rm debug_service.key debug_service.csr debug_service.crt`,
    `chmod 444 debug_service.pfx`
```

src/api/debug/certificates.ts
```diff
@@ -96,12 +104,18 @@ export async function setup(connection: IBMi) {
...
+    }
+  }
+
   const commands = [
-    `openssl genrsa -out debug_service_ca.key 2048`,
-    `openssl req -x509 -new -nodes -key debug_service_ca.key -sha256 -days 1825 -out debug_service_ca.pem -subj '/CN=${host}'`,
     `openssl genrsa -out debug_service.key 2048`,
     `openssl req -new -key debug_service.key -out debug_service.csr -subj '/CN=${host}'`,
-    `openssl x509 -req -in debug_service.csr -CA debug_service_ca.pem -CAkey debug_service_ca.key -CAcreateserial -out debug_service.crt -days 1095 -sha256 -sha256 -req -extfile <(printf "${extFileContent}")`,
+    `openssl x509 -req -in debug_service.csr -signkey debug_service.key -out debug_service.crt -days 1095 -sha256 -sha256 -req -extfile <(printf "${extFileContent}")`,
```
sebjulliand, line 118: Why an arbitrary 1095 days here?

sebjulliand (changes requested):
It works as expected. I changed the certificate location, and I could still debug.
Just a suggestion to think about and we'll be good to go!

src/api/debug/certificates.ts
sebjulliand: Better idea: remove the `crt` file and get the certificate from the service itself, so you're guaranteed to get the right client certificate.

Get the output from this when the service runs (here on port 8005):
```bash
openssl s_client -connect localhost:8005 -showcerts < /dev/null 2> /dev/null | openssl x509 -outform PEM
```

worksofliam: Can you make an issue for that idea itself?

sebjulliand: I can even PR the hell out of it if you want.

sebjulliand: Hold my tea...

sebjulliand (approved):
@worksofliam looks good to me.
I pushed a commit to clean up the code a bit and also make it clear that there is a service certificate and a client certificate. I also made sure only a PFX file can be selected when importing a service certificate (even though that feature would require more work).

Anyway, have a look and merge when you're comfortable.

PR #1899, opened 2024-03-01
Abstract API for running multiple statements

worksofliam:
### Changes

I have added an API called `runStatements`, which just allows us to pass an array of statements, be it SQL or CL, to be executed in the same job.

I also changed `getQtempTable` to use this API as an example of how it works.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -246,6 +246,14 @@ export default class IBMiContent {
   }
 
   /**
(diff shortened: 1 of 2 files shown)


worksofliam:
Test cases are passing. I also updated a test case to try the new API.

image

sebjulliand (approved):
LGTM @worksofliam ; I added a commit to propose the use of a rest parameter with `runStatements` to make it even easier to use.
Feel free to revert if you don't like it.
Then merge!

PR #1905, opened 2024-03-03
'Add to library list' and 'Set as current library' actions on browsers items

sebjulliand:
### Changes
This PR adds a new `Set as current library` action, available from:
- Libraries in the `User Library List` view
- Single library Filters and .lib objects from the `Object Browser`

It adds the `Add to library list` action on:
- Single library Filters and .lib objects from the `Object Browser`

image

It also allows to run `Create source file` action on:
- Single library  Filters and .lib objects from the `Object Browser`

Finally, it removes `Add to library list...` from command palette when library list view is disabled

@chrjorgensen, I'll need your help for one translation 😄 

### How to test
- Try `Add to library list` action from:
  - A library from the `User Library List` view
  - A filter from the `Object Browser`
  - A .lib object from the `Object Browser`
  - The `User Library List` view's title
- Try `Set as current library` action from:
  - A library from the `User Library List` view
  - A filter from the `Object Browser`
  - A .lib object from the `Object Browser`

### Checklist
* [x] have tested my change

### DIFF
--- a/package.json
+++ b/package.json
@@ -1138,9 +1138,16 @@
 				"category": "IBM i"
 			},
 			{
-				"command": "code-for-ibmi.addToLibraryList",
+				"command": "code-for-ibmi.addToLibraryList.prompt",
 				"enablement": "code-for-ibmi:connected",
 				"title": "Add to Library List...",
+				"category": "IBM i",
+				"icon": "$(add)"
+			},
+			{
+				"command": "code-for-ibmi.addToLibraryList",
+				"enablement": "code-for-ibmi:connected",
+				"title": "Add to Library List",
 				"category": "IBM i",
 				"icon": "$(add)"
 			},
@@ -2049,7 +2056,7 @@
 					"when": "view == libraryListView"
 				},
 				{
-					"command": "code-for-ibmi.addToLibraryList",
+					"command": "code-for-ibmi.addToLibraryList.prompt",
 					"group": "navigation",
 					"when": "view == libraryListView"
 				},
--- a/package.json
+++ b/package.json
@@ -1108,7 +1108,7 @@
 				"command": "code-for-ibmi.setCurrentLibrary",
 				"enablement": "code-for-ibmi:connected",
 				"title": "Set as Current Library",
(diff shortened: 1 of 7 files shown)


chrjorgensen (changes requested):
Nice enhancement, @sebjulliand ! :+1: 

I translated the new error message to Danish and changed a command title to title case.

It's working as expected - only one issue was found:
- filters having multiple libraries
  - Add to Library List returns error `Library is too long.`
  - Set to Current Library does nothing

In addition, the context menu item `Add to Library List...` does not prompt the user for any information and should not have `...` in the end... however, when it's run from the command Palette, it prompts the user and should have `...`.
AFAICT the only solution to this situation is two commands with a correct title - where one should only be for the title and should call the other for the action. Too much trouble?

sebjulliand:
You're right @chrjorgensen, as always!
I introduced a new type (`WithLibrary`) and modified the object browser's items context so the library related actions (including `Create source file`) while only appear on library nodes. This gives a better UX and a cleaner code overall.

I updated the description of the PR accordingly.

package.json
```diff
@@ -1104,6 +1104,12 @@
 				"category": "IBM i",
 				"icon": "$(root-folder)"
 			},
+			{
+				"command": "code-for-ibmi.setCurrentLibrary",
+				"enablement": "code-for-ibmi:connected",
+				"title": "Set as Current Library",
+				"category": "IBM i"
```
chrjorgensen, line 1111: A few tabs appending the line...

src/typings.ts
```diff
@@ -157,7 +157,9 @@ export interface WithPath {
   path: string
 }
 
-export interface Library extends WithPath { }
+export interface WithLibrary {
+  library:string
```
chrjorgensen, line 161:
```suggestion
  library: string
```

sebjulliand:
@chrjorgensen formatting done!

chrjorgensen (approved):
@sebjulliand Works like a charm - and looks good! 🎉

PR #1907, opened 2024-03-04
Update editor tab on member rename

steph-beneschan-256:
### Changes

* When the user renames a member from the object explorer, if the member was open in an editor tab, the tab will now be replaced by a tab showing the member with the new name.
* If the user is currently editing a member, and there are unsaved changes, the user can no longer rename the member from the object explorer.

This is meant to provide a workaround to issue #1554 . It seems that the VS Code API does not let us directly change an open tab's label or uri, so instead I opted to have the system close the tab, then open a new one with the updated uri. Notably, the new tab will not necessarily be in the same position as the old one.

Hopefully this will suffice as a temporary solution, but I will continue to look into this.

### Checklist

* [x] have tested my change


worksofliam:
I will have a formal review in later for this, but should we consider implementing the same for stream files?

Perhaps the function you added to get the tab should be part of Tools?

worksofliam:
@steph-beneschan-256 

Two things:

1. Thanks for moving that API into tools. Are you also planning to add the same safeguarding logic for IFS objects too? E.g. when move/rename is used but the stream file is open, cancel the operation? This would be in `ifsBrowser.ts`. No is ok, just checking.
2. Please fix the conflicts and I will test further!

steph-beneschan-256:
I will add the check to the IFS browser, and also resolve the conflicts.

steph-beneschan-256:
I made a few more commits; the new logic should now apply to the IFS browser as well as the object browser.

Also, I noticed that if I use the IFS browser to rename/move an IFS file to a new path, if an IFS file already exists at the new path, it will be overwritten. Is that the intended behavior? (I ask because if I try to do the same thing using the IFS utility in ACS, the operation fails and no files are modified.) If it's a problem, I wouldn't mind looking into it.

worksofliam:
@chrjorgensen What do you you think about @steph-beneschan-256's finding of the file being overwritten?

chrjorgensen:
> @chrjorgensen What do you you think about @steph-beneschan-256's finding of the file being overwritten?

@worksofliam It has always been like that - since the command executed on the server is a `mv` command which overwrites an existing file, if the existing file is writable. See `mv` documentation here: https://www.gnu.org/software/coreutils/manual/html_node/mv-invocation.html#mv-invocation

I think our users has gotten used to this behavior, we don't have any issues about this...

src/locale/ids/fr.json
```diff
@@ -127,6 +127,8 @@
   "enddbgsvr.succeeded": "Serveur de débogage arrêté.",
   "error": "erreur",
   "errors": "erreurs",
+  "file.path.not.parsed": "The file path could not be parsed.",
```
sebjulliand, line 130:
```suggestion
  "file.path.not.parsed": "Le chemin du fichier n'a pas pu être analysé.",
```

PR #1916, opened 2024-03-06
Fixed file extension when downloading a single IFS file

sebjulliand:
### Changes
When downloading a single IFS file, the extension in the File selection dialog is not right (it starts with `.`) and the dialog doesn't show the files with this extension:
image

This PR removes the `.` in the extension name and fixes this issue:
image

### How to test this PR
1. With version 2.8.0: try downloading a single file from the IFS browser: extension is wrong, no existing files with the extension are shown.
2. With the PR checked out: try downloading a single file from the IFS browser: extension is OK, files with the same extension in the folder are shown.

### Checklist
* [x] have tested my change


worksofliam (approved):
Tested on Mac and it continued to work good before and after.

PR #1927, opened 2024-03-08
Added "Duplicate connection" action

sebjulliand:
### Changes
Resolves https://github.com/codefori/vscode-ibmi/issues/1925

This PR adds the `Duplicate...` action on the connections browser items:
image

When run, it will prompt the user for the new connection name. If a new, non existing name is provided, the login settings will be duplicated as well as the connection parameters, except the following fields:
- `homeDirectory`
- `currentLibrary`
- `libraryList`
- `objectFilters`
- `ifsShortcuts`
- `customVariables`
- `commandProfiles`
- `connectionProfiles`

But maybe this should be optional to keep them when duplicating? 🤔
@chrjorgensen your Danish translations will be much appreciated 😉

### How to test this PR
1. Run the action on a connection
2. Check it doesn't allow to use an existing name for the new connection
3. Check if the settings have been duplicated (except filters, profiles, etc)
4. Check if the new connection works

### Checklist
* [x] have tested my change

### DIFF
--- a/package.json
+++ b/package.json
@@ -1035,6 +1035,11 @@
 				"category": "IBM i"
 			},
 			{
+				"command": "code-for-ibmi.duplicateConnection",
+				"title": "Duplicate...",
+				"category": "IBM i"
+			},
+			{
 				"command": "code-for-ibmi.disconnect",
 				"enablement": "code-for-ibmi:connected",
 				"title": "Disconnect",
@@ -1829,6 +1834,10 @@
 					"when": "never"
 				},
 				{
+					"command": "code-for-ibmi.duplicateConnection",
+					"when": "never"
+				},
+				{
 					"command": "code-for-ibmi.disconnect",
 					"when": "code-for-ibmi:connected"
 				},
@@ -2273,9 +2282,14 @@
 					"group": "2_delete@1"
 				},
 				{
+					"command": "code-for-ibmi.duplicateConnection",
+					"when": "view == connectionBrowser && !listMultiSelection && viewItem == server",
+					"group": "2_delete@2"
+				},
+				{
 					"command": "code-for-ibmi.deleteConnection",
 					"when": "view == connectionBrowser && viewItem == server",
-					"group": "2_delete@2"
+					"group": "2_delete@3"
 				},
 				{
 					"command": "code-for-ibmi.showAdditionalSettings",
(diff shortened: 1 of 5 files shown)


chrjorgensen (changes requested):
@sebjulliand A small but very effective enhancement. :heart_eyes: 

I have only a few questions:
- Why have you chosen the term `Duplicate`? To me the term `Copy` is more straightforward and consistent with the rest of the treeviews in the extension. I also think `Copy` (or `Duplicate`) should be before the `Rename` in the command list (as in our other views).
- IMHO it would be great to have the remaining connections settings as a multiple choice selection in a step two, where all are checked and the user can uncheck what's not to be included in the new connection. What's your opinion?
- You don't copy the stored password, is that deliberate? I copied my PUB400 connection, which have a password stored, but the new connection asked for the password. I think a copy should be a fully working clone of the original connection, by default. 

I will translate when we have a consensus about the term - never want to repeat work! :laughing:

sebjulliand:
I agree with your points @chrjorgensen. I think I initially chose `Duplicate...` because Copy makes me think of copy/paste; but that would not be consistent with the rest, so....`Copy...` it is!

Point 2 answers my initial question, so let's do this.
And the password will be copied as well!

sebjulliand:
@chrjorgensen done as asked 😄  Let me know what you think!

chrjorgensen (changes requested):
Looks good! 😍 

Just nit-picking: Could the two dialogs be connected in a multi-step dialog, so the user can return to the name for the new connection? And maybe show the new name in step 2?

Please, please, please! 😉

sebjulliand:
@chrjorgensen : but of course 😉 
The second dialog shows "from `xxx` to `yyy`" now and the prompts are now connected to create a workflow.

chrjorgensen (approved):
@sebjulliand Works like a charm! ❤️ 

Sorry for my mistake: I remembered the multistep sample with step count and back arrow and thought it was available in standard API - but it is not. This is fine, Esc will go back in this multistep dialog.

You don't copy the cached server settings - I guess that's a deliberate choice of yours...

I translated the texts into Danish - and added quotes around the connection name parameter, since the name can contain spaces and should be quoted.

Looking good to me - approved! 🎉 

Merge when you feel for it. 👍

sebjulliand:
> @sebjulliand Works like a charm! ❤️
> 
> Sorry for my mistake: I remembered the multistep sample with step count and back arrow and thought it was available in standard API - but it is not. This is fine, Esc will go back in this multistep dialog.
> 
> You don't copy the cached server settings - I guess that's a deliberate choice of yours...
> 
> I translated the texts into Danish - and added quotes around the connection name parameter, since the name can contain spaces and should be quoted.
> 
> Looking good to me - approved! 🎉
> 
> Merge when you feel for it. 👍

Thank you @chrjorgensen ! Indeed not copying the cache is deliberate. It doesn't hurt starting from fresh just once for a new connection, just to be on the safe side. And it didn't feel right to copy a cache anyway. Thanks for the translation and the reviews! Let's merge 🥳

PR #1928, opened 2024-03-09
Fixed drag crashing for streamfiles

sebjulliand:
### Changes
This fixes an issue preventing streamfiles from being dragged and dropped from the IFS browser.
Dragging a streamfile would trigger the following error, shown in the developer console:
image

That happened during the item JSON serialization finding a circular reference (i.e. `IFSFileItem`'s `command` would reference `this`, which indeed creates a circular reference).

To fix this, the command won't reference `this` anymore but only the required field to call the command (i.e. `this.path`).

The same fix has been applied on `ObjectBrowserMemberItem` to prevent the same problem in case it gets serialized.

### How to test this PR
1. Confirm drag'n'drop fails for a single streamfile without the PR (checkout the developer console to confirm the error)
2. Checkout the PR
3. Confirm drag'n'drop succeeds for a single streamfile
4. Opening a streamfile for browse/edit must work
5. Opening a source member for browse/edit must work

### Checklist
* [x] have tested my change

### DIFF
--- a/src/views/ifsBrowser.ts
+++ b/src/views/ifsBrowser.ts
@@ -138,10 +138,12 @@ class IFSFileItem extends IFSItem {
     this.contextValue = "streamfile";
     this.iconPath = vscode.ThemeIcon.File;
 
-    this.resourceUri = vscode.Uri.parse(this.path).with({ scheme: `streamfile` }); this.command = {
+    this.resourceUri = vscode.Uri.parse(this.path).with({ scheme: `streamfile` });
+    
+    this.command = {
       command: "code-for-ibmi.openWithDefaultMode",
       title: `Open Streamfile`,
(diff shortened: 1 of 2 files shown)


chrjorgensen (commented):
@sebjulliand Drag and drop of streamfiles works now - also into the editor space.

I'm not sure what you mean by
>Opening a streamfile for browse/edit must work
>Opening a source member for browse/edit must work

Streamfiles can be opened by dragging or clicking.
Members can only be opened by clicking. Isn't that enough?

sebjulliand:
@chrjorgensen it's just to be safe, because I made a change on the command call bound to the ifs file item and source member item.

chrjorgensen (approved):
Understood. Approved and merged. :+1:

PR #1940, opened 2024-03-18
Correctly handle source members with variant characters in their name

sebjulliand:
### Changes
Fixes https://github.com/codefori/vscode-ibmi/issues/1935

Member names used to be uppercased regarless of the variant characters they may contain. Turns out that `à` is one of CCSID 297 (French) variant chars and it gets uppercased to `À`, which raises a `member not found` error.

This PR adds a new `upperCaseName` method to the `IBMi` class to correctly uppercase names, preservingthe variant chars case.

### How to test this PR
1. Set PASE_LANG environment variable to `FR_FR`
2. Open a connection with a profile whose CCSID is `297`
3. Try to open a member with one or more `à` in its name

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -61,6 +61,8 @@ export default class IBMi {
 
   commandsExecuted: number = 0;
 
+  dangerousVariants = false;
+
   constructor() {
     this.client = new node_ssh.NodeSSH;
     this.currentHost = ``;
@@ -877,6 +879,9 @@ export default class IBMi {
           defaultCCSID: this.defaultCCSID
         });
 
+        //Keep track of variant characters that can be uppercased
+        this.dangerousVariants = this.variantChars.local !== this.variantChars.local.toLocaleUpperCase();
+
         return {
           success: true
         };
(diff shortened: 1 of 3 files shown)


chrjorgensen (changes requested):
@sebjulliand I gave the PR a spin, but had no success... 😞 

With or without SQL enabled, when I tried to open a member having `à` in the name, I got the following error:

billede

The member name is correct in the list:

billede

???

Does it work at your end? Could be my system setup...

sebjulliand:
@chrjorgensen check out my comment here: https://github.com/codefori/vscode-ibmi/issues/1935#issuecomment-2003641178

I'll add a commit to this PR with the copy to a temporary member (I managed to open the member, I just need to apply the same logic when saving). We'll see how it goes (I'll un-draft the PR then).

sebjulliand:
@chrjorgensen you can give it another try; I managed to open and save my member with a `à` using this patch.

sebjulliand:
@chrjorgensen you're right, let's reduce the scope only to those who use dangerous variants 😅

sebjulliand:
@chrjorgensen there you go

chrjorgensen (approved):
@sebjulliand I can confirm being able to open and edit and save members having `à` in the member name when I simulate being French (CCSID 297 etc) when running your PR. 👍 

Approved - merge at will... 😃

sebjulliand:
Thank you @chrjorgensen !

PR #1941, opened 2024-03-19
Replace all toUpperCase whith IBMi.upperCaseName where relevant

sebjulliand:
### Changes
Fixes https://github.com/codefori/vscode-ibmi/issues/1935

The previous PR did not cover all the use cases for object names with a `à` on a connection with CCSID 297.
This PR replaces all the calls to `toUpperCase`/`toLocaleUpperCase` applied to object/member names with `IBMi.upperCaseName`. It fixes the issues where `à` gets wrongly upper cased.

### How to test this PR
1. Connect with CCSID 297
2. Create a library with a name containing `à`
3. In this library, create a source file with a name containing `à`
4. In this source file, create a member with a name containing `à`
5. Run actions from the library list view and object browsers on these items

### Checklist
* [x] have tested my change

### DIFF
--- a/src/webviews/settings/index.ts
+++ b/src/webviews/settings/index.ts
@@ -259,7 +259,7 @@ export class SettingsUI {
                     case `protectedPaths`:
                       data[key] = String(data[key]).split(`,`)
                         .map(item => item.trim())
-                        .map(item => item.startsWith('/') ? item : item.toUpperCase())
+                        .map(item => item.startsWith('/') ? item : connection?.upperCaseName(item) || item.toUpperCase())
                         .filter(item => item !== ``)
                         .filter(Tools.distinct);
                       break;
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -541,7 +541,7 @@ export default class IBMi {
           // We need to check if our remote programs are installed.
           remoteApps.push(
             {
-              path: `/QSYS.lib/${this.config.tempLibrary.toUpperCase()}.lib/`,
+              path: `/QSYS.lib/${this.upperCaseName(this.config.tempLibrary)}.lib/`,
               names: [`GETNEWLIBL.PGM`],
               specific: `GE*.PGM`
             }
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -515,14 +515,14 @@ export default class IBMiContent {
    * @returns an array of IBMiObject
    */
   async getObjectList(filters: { library: string; object?: string; types?: string[]; filterType?: FilterType }, sortOrder?: SortOrder): Promise<IBMiObject[]> {
-    const library = filters.library.toUpperCase();
+    const library = this.ibmi.upperCaseName(filters.library);
     if (!await this.checkObject({ library: "QSYS", name: library, type: "*LIB" })) {
       throw new Error(`Library ${library} does not exist.`);
     }
 
     const singleEntry = filters.filterType !== 'regex' ? singleGenericName(filters.object) : undefined;
     const nameFilter = parseFilter(filters.object, filters.filterType);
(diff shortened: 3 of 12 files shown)


worksofliam:
@sebjulliand Can you confirm the test cases are passing?

sebjulliand:
> @sebjulliand Can you confirm the test cases are passing?

They do, save for the `Create programs...` ones, but it doesn't seem related.
I also fixed the `runCommand ILE` that was based on a localized output.
And I added a test case for `upperCaseName`.

src/webviews/settings/index.ts
```diff
@@ -259,7 +259,7 @@ export class SettingsUI {
                     case `protectedPaths`:
                       data[key] = String(data[key]).split(`,`)
                         .map(item => item.trim())
-                        .map(item => item.startsWith('/') ? item : item.toUpperCase())
+                        .map(item => item.startsWith('/') ? item : connection?.upperCaseName(item) || item.toUpperCase())
```
chrjorgensen, line 262: @sebjulliand Is this for satisfying TypeScript? Isn't the connection always available?

chrjorgensen (changes requested):
@sebjulliand I have tested the following with success:

- creating a library called `TESTà`
- creating a source file called `TESTàGAIN` in `TESTà`
- creating a member called `TESTà.RPGLE` in source file `TESTà/TESTàGAIN`
- changing the description on the library, source file and member

However, resolving a member in an INCLUDE statement fails with the following output:

```
/home/<user>: for f in /QSYS.LIB/QGPL.LIB/TESTàGAIN.FILE/TESTÀCOPY.MBR /QSYS.LIB/QGPL.LIB/TESTàGAIN.FILE/TESTÀCOPY.MBR /QSYS.LIB/QTEMP.LIB/TESTàGAIN.FILE/TESTÀCOPY.MBR /QSYS.LIB/TESTà.LIB/TESTàGAIN.FILE/TESTÀCOPY.MBR; do if [ -f $f ]; then echo $f; break; fi; done
{
    "code": 0,
    "signal": null,
    "stdout": "",
    "stderr": ""
}
```

Seems like the member name is converted to uppercase using normal uppercase routines... 

@worksofliam Is it the RPGLE extension that does this? I think so...

@sebjulliand Maybe your new uppercase function needs to be exposed to the other extensions? I guess the CL extension and maybe others are hit as well.

worksofliam:
@chrjorgensen It looks like Seb is correctly using his function in the `memberResolve` method:

https://github.com/codefori/vscode-ibmi/pull/1941/files#diff-e1962db673dbf505e2c74deb034072677a2b4fca832d78a983eebeecac55df85R782

But, I have found where we are uppercasing in the vscode-rpgle extension for the member. I have removed that uppercase call to let the `memberResolve` API handle it, since Seb made the change above.

https://github.com/codefori/vscode-rpgle/commit/493643329c7120c6edb72543d34818c508ab6833#diff-0a08da2f6168c85dd231fc09362ce0c8ead2626b37926c63738004aee6b90a39R88

I am doing a release of vscode-rpgle, in which you should be able to test again with vscode-rpgle 0.26.3.

src/webviews/settings/index.ts
sebjulliand: It is weird, but it's not to satisfy TypeScript - connection can be `undefined` here if you change your settings while not connected.
image

chrjorgensen (approved):
@sebjulliand In addition to my tests from yesterday, I've run the following tests:

- rename member `TESTà.RPGLE`
- compile member `TESTà.RPGLE`
- search sourcefile `TESTà/TESTàGAIN`

All tests ran fine.

The failing tests yesterday are not caused by this change, but stem from other extensions / issues.

So I will approve this one - merge at will... 👍

sebjulliand:
Thank you guys! Let's gooooo!

PR #1949, opened 2024-03-20
Fix member and object resolve with variants

chrjorgensen:
### Changes

Fixes resolving members and objects having variant characters in their name.
Tests have been added to verify member and object resolve with variant characters.

Before this would not work:

billede

After this fix the member resolution now works:

billede

### How to test this PR

1. create a source member with variant characters in the name
2. open another source member and add /COPY or /INCLUDE referencing the source member with variant characters
3. Press `Alt-F12` with the cursor in the referenced member name.
4. Verify the referenced member content is shown.
5. Run the test cases

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -778,22 +778,24 @@ export default class IBMiContent {
   }
 
   async memberResolve(member: string, files: QsysPath[]): Promise<IBMiMember | undefined> {
+    const inAmerican = (s: string) => { return this.ibmi.sysNameInAmerican(s) };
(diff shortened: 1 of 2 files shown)


worksofliam:
@chrjorgensen What's the state of this PR? It is assigned to you, but is it ready for review?

chrjorgensen:
@worksofliam Yes, ready for review since I requested a review from you and Seb... 😉😘

worksofliam:
Code looks good and your new + previous tests are still passing. I also used your step to test the copy book and it's looking good to me. Let's merge for 2.9.0.

PR #1954, opened 2024-03-24
Added debugger status in status bar + status editor

sebjulliand:
### Changes
Resolves https://github.com/codefori/vscode-ibmi/issues/1921

- Adds a `Debugger (on/off)` status item in server's status bar.
- Adds a `Debugger status` dashboard to check `debug server` and `debug service` job statuses, job details, certificate status and actions to start/restart/stop the server/service, generate server certificate if needed and download client certificate.

Work in progress...🛠

### How to test this PR
Wait...

### Checklist
* [ ] have tested my change


worksofliam:
@sebjulliand I am just waiting for the SEP PR to be merged in before I review this. Initial code review looks good, but this is a major UX PR, which is where it will really need some love.

sebjulliand:
Updated for SEP
- Added debugger version in status + dashboard title
- Added support to display multiple ports in dashboard

sebjulliand:
- Added a `No debug PTF` status item
- Debug commands and settings now hidden depending on the debugger version

sebjulliand:
Added the IBM i Debugger treeview in the Debug view.
image

worksofliam:
@sebjulliand Wow, the Debugger tree view is so nice. But, I have some UX requests - sorry if it's a lot:

* Hover over offline Debug Service shows 'undefined' in tooltip
* For both debug service and service, I think the tooltips should show the same information that is on their tabs in the debugger dashboard.
* I think Debug Service, if no certificate exists (if the `.pfx` doesn't exist), should have a child node. This node wouldn't have a start/stop button, but instead would have the create button. The node would not be shown if the server cert does exist.
* We need a way in the tree to show the state of the client certificate, perhaps one of the following?
    * If no client cert is on the client system, then the Debug Service could show an exclamation mark and the description to something relevant ('service started by missing client certificate') - perhaps along with start/stop could be 'fetch' [client certificate]
    * Perhaps there could be an additional node below Debug Service specifically for client certificate

src/api/debug/server.ts
```diff
@@ -1,134 +1,183 @@
...
-  return;
+  let tries = 0;
+  while (!didNotStart && tries < 20) {
+    if (await getDebugServiceJob()) {
+      window.showInformationMessage(t("start.debug.service.succeeded"));
+      refreshDebugSensitiveItems();
+      return true;
+    }
+    else {
+      await Tools.sleep(500);
```
worksofliam, line 133: I personally think we should consider using setTimeout instead of `sleep`.

sebjulliand: Fair enough!

sebjulliand: But this `Tools.sleep` is a setTimeout in disguise though.

worksofliam: Sorry - I meant setInterval!

sebjulliand: It now uses `setInterval`. Is there an actual benefit for using it or is it just cleaner?

sebjulliand:
> @sebjulliand Wow, the Debugger tree view is so nice. But, I have some UX requests - sorry if it's a lot:
> 
>     * Hover over offline Debug Service shows 'undefined' in tooltip

It won't show anything now when the server/service is `offline`.

>     * For both debug service and service, I think the tooltips should show the same information that is on their tabs in the debugger dashboard.

Done!
image
image
 
>     * I think Debug Service, if no certificate exists (if the `.pfx` doesn't exist), should have a child node. This node wouldn't have a start/stop button, but instead would have the create button. The node would not be shown if the server cert does exist.
> 
>     * We need a way in the tree to show the state of the client certificate, perhaps one of the following?
>       
>       * If no client cert is on the client system, then the Debug Service could show an exclamation mark and the description to something relevant ('service started by missing client certificate') - perhaps along with start/stop could be 'fetch' [client certificate]
>       * Perhaps there could be an additional node below Debug Service specifically for client certificate

I merged your ideas under one dedicated sub-node of the service node. When there is a certificate problem, an exclamation mark is shown on the debug service node, and it is expanded by default to show what the problem is.

The problem can be:
- There is no remote certificate: the service can't be started and an action on the problem node allows to generate the certificate
image

- Secure Debug is on and there is no local certificate or it exists but doesn't match the remote certificate: the service can be started and there is an action to download the certificate.
image

@worksofliam I guess it's not too bad now 😅 Have another swing at it!

package.json
```diff
@@ -1737,6 +1769,13 @@
 					"name": "Results",
 					"when": "code-for-ibmi:searchViewVisible"
 				}
+			],
+			"debug": [
+				{
+					"id": "ibmiDebugBrowser",
```
worksofliam, line 1775: I think this view should be collapsed by default.

package.json
```diff
@@ -1737,6 +1769,13 @@
...
+					"name": "IBM i debugger",
+					"when": "code-for-ibmi:connected && code-for-ibmi:debug"
```
worksofliam, line 1777: Need an additional check here: `code-for-ibmi:debugManaged == false`

worksofliam (changes requested):
I only have minor feedbacks now. Overall it's bloody great.

* I think the new tree view should be collapsed by default.
* As noted, the tree view should never be visible when in managed state.

* The text in this window doesn't make sense anymore as it doesn't generate and download - it solely generates - there is then another step to download the certificate.

image

Now for the big ones:

1. I think we can remove the webview totally in place of the tree view.
2. The button in the hover menu to open the web view should instead focus on the tree with, which in return will also expand it.

image

package.json
sebjulliand: I'd rather say we keep the view, even when managed, but we disable the actions.

package.json
sebjulliand: Done!

sebjulliand:
>     * I think the new tree view should be collapsed by default.

It is now.

>     * As noted, the tree view should never be visible when in managed state.

I made it so the actions to start/stop the debugger and setup local/remote certificates are disabled when managed. But I don't think seeing the debugger status when managed will hurt 😁

> 
>     * The text in this window doesn't make sense anymore as it doesn't generate and download - it solely generates - there is then another step to download the certificate.

It would actually make sense to download the client certificate afterwards so I added that.

> 
> Now for the big ones:
> 
>     1. I think we can remove the webview totally in place of the tree view.

The webview is gone. Goodbye, sweet angel 😋

> 
>     2. The button in the hover menu to open the web view should instead focus on the tree with, which in return will also expand it.
> 

Done!

Back at you @worksofliam !

sebjulliand:
> I love it!! It's so nice and super clean. Code looks good too - fab job!
Yay!
> There is one major UX bug: I can spam all the item buttons, which can incur many requests to the server:
> image
> 
> I think, somehow, these buttons need to be able be disabled while an action is happening.

Good finding @worksofliam , thanks! I took care of that; tell me what you think!

worksofliam (approved):
Works like a charm! I absolutely **love** your `withContext` implementation! Got to use that in a few other places.

sebjulliand:
Brilliant, thank you @worksofliam !
@chrjorgensen can I request your expertise for translating the `da` messages? Thanks in advance!

chrjorgensen:
@sebjulliand Here you are... 🇩🇰 😃

sebjulliand:
> @sebjulliand Here you are... 🇩🇰 😃

Mange tak! 🥰

PR #1962, opened 2024-03-27
Added support for .deployignore

sebjulliand:
### Changes
Resolves https://github.com/codefori/vscode-ibmi/issues/1951

This adds the support for a `.deployignore` file that can be put at the root of a workspace to define patterns for files that should not be deployed by Code for IBM i when deploying a workspace. Files ignored during deployment can also be defined in `.gitignore`, but if `.deployignore` exists, it replaces `.gitignore` for this.

`.deployignore` supports the exact same format as `.gitignore`.

### How to test this PR
1. Run the test case `Test .deployignore` under `Deploy Tools API tests`
2. Create a `.deployignore` file in the workspace
3. Add a pattern to ignore something different from `.gitignore`
4. Deploy
5. Check the rules from `.deployignore` superseds `.gitignore`'s

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/local/deployTools.ts
+++ b/src/api/local/deployTools.ts
@@ -2,10 +2,9 @@ import createIgnore, { Ignore } from 'ignore';
 import path, { basename } from 'path';
 import vscode, { Uri, WorkspaceFolder } from 'vscode';
 import { instance } from '../../instantiate';
-import { LocalLanguageActions } from './LocalLanguageActions';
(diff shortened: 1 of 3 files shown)


worksofliam (approved):
1. I tested manually by creating my own `.deployignore` file combination with expecting a compile failure since files were not uploaded ✅ 
2. The test case you provided makes sense and is passing ✅

PR #1975, opened 2024-04-04
Cleanup/connection_manager_api

worksofliam:
### Changes

Replaces raw access of `connections` setting and password context with a `ConnectionManager` namespace.

### How to test this PR

1. Ensure the same test cases pass
2. Try out the extension as normal
3. Confirm debug still works
4. When editing connection settings or login settings, ensure the sort button is disabled.

### Checklist

* [ ] have tested my change


src/api/Configuration.ts
```diff
@@ -30,6 +30,79 @@ export namespace GlobalConfiguration {
...
 
+export interface StoredConnection {
+  index: number,
+  data: ConnectionData
+};
+
+export namespace ConnectionManager {
+  export function getByName(name: string): StoredConnection | undefined {
+    const connections = getAll();
+    if (connections) {
```
sebjulliand, line 41: `connections` is never falsy, no need to test it.

src/api/Configuration.ts
```diff
@@ -30,6 +30,79 @@ export namespace GlobalConfiguration {
...
+  export function getAll(): ConnectionData[] {
+    return GlobalConfiguration.get<ConnectionData[]>(`connections`) || [];
+  }
+
+  function setAll(connections: ConnectionData[]) {
+    return GlobalConfiguration.set(`connections`, connections);
+  }
+
+  export async function storeNew(data: ConnectionData): Promise<StoredConnection> {
+    let connections = getAll();
```
sebjulliand, line 64: Never use `let` unless it's absolutely necessary 😉
```suggestion
    const connections = getAll();
```

src/api/Configuration.ts
```diff
@@ -30,6 +30,79 @@ export namespace GlobalConfiguration {
...
+    connections[index] = data;
+
+    // Remove possible password from any connection
+    connections.forEach(conn => delete conn.password);
+
+    return GlobalConfiguration.set(`connections`, connections);
+  }
+
+  export function getStoredPassword(context: vscode.ExtensionContext, connectionName: string) {
+    const connectionKey = `${connectionName}_password`;
```
sebjulliand, line 91: Since `${connectionName}_password` is repeated three times, create a small function to encompass the logic.

src/webviews/login/index.ts
```diff
@@ -141,25 +136,25 @@ export class Login {
...
-    const connection = instance.getConnection();
-    if (connection) {
+    const existingConnection = instance.getConnection();
+    if (existingConnection) {
       // If the user is already connected and trying to connect to a different system, disconnect them first
-      if (name !== connection.currentConnectionName) {
-        vscode.window.showInformationMessage(`Disconnecting from ${connection.currentHost}.`);
+      if (name !== existingConnection.currentConnectionName) {
+        vscode.window.showInformationMessage(`Disconnecting from ${existingConnection.currentHost}.`);
         if (!disconnect()) return false;
```
sebjulliand, line 144: We're missing an `await` here.
```suggestion
        if (!await disconnect()) return false;
```

src/views/ConnectionBrowser.ts
```diff
@@ -95,12 +95,12 @@ export function initializeConnectionBrowser(context: vscode.ExtensionContext) {
 
         if (newName) {
           try {
-            let index;
             // First rename the connection details
-            const connections = GlobalConfiguration.get<ConnectionData[]>(`connections`) || [];
-            index = connections.findIndex(connection => connection.name === server.name);
+            const existingConnection = ConnectionManager.getByName(server.name)!;
+            let { index, data } = existingConnection!
```
sebjulliand, line 100: Since we don't reuse `existingConnection`...
```suggestion
            let { index, data } = ConnectionManager.getByName(server.name)!;
```

src/views/ConnectionBrowser.ts
```diff
@@ -268,7 +261,7 @@ class ConnectionBrowser implements vscode.TreeDataProvider<ServerItem> {
 
   async getChildren(): Promise<ServerItem[]> {
     const lastConnection = GlobalStorage.get().getLastConnections()?.[0];
-    return (GlobalConfiguration.get<ConnectionData[]>(`connections`) || [])
+    return (ConnectionManager.getAll() || [])
```
sebjulliand, line 264: `getAll` already returns an array no matter what.
```suggestion
    return ConnectionManager.getAll()
```

src/extension.ts
```diff
@@ -36,7 +36,7 @@ export async function activate(context: ExtensionContext): Promise<CodeForIBMi>
 
   await loadAllofExtension(context);
   const checkLastConnections = () => {
-    const connections = (GlobalConfiguration.get<ConnectionData[]>(`connections`) || []);
+    const connections = (ConnectionManager.getAll() || []);
```
sebjulliand, line 39: `getAll` always returns an array.
```suggestion
    const connections = ConnectionManager.getAll();
```

src/webviews/settings/index.ts
```diff
@@ -381,8 +380,8 @@ export class SettingsUI {
                 delete data.password;
                 delete data.buttons;
 
-                connections[connectionIdx] = Object.assign(connection, data);
-                await GlobalConfiguration.set(`connections`, connections);
+                stored = Object.assign(stored, data);
+                await ConnectionManager.updateByIndex(index, stored);
```
sebjulliand, line 384: We should refresh the browser here to update the connection node name/description.
```suggestion
                await ConnectionManager.updateByIndex(index, stored);
                vscode.commands.executeCommand(`code-for-ibmi.refreshConnections`);
```

sebjulliand (changes requested):
Nice cleanup @worksofliam ! Way cleaner than the raw accesses we used to do. I love it!
See the few comments I made, mostly code cleanup.

One thing to consider: since we update by index, there is a use case where the data can be saved on the wrong index:
- Open the settings of a connection and keep it opened
- Sort the connections browser
- Save the opened settings

Since the indexes were changed while sorting, the index kept by the settings editor is not the right one anymore.
I admit it's an edge case...but if I thought of it, someone else will.
So maybe we should set a context value while the webview is opened to prevent the sort and rename actions from being executed.

worksofliam:
@sebjulliand Sorry this took SO long. I went ahead and used `withContext` around the UI logic to disable the buttons. Check it out and let me know - thanks!

src/webviews/settings/index.ts
```diff
@@ -243,169 +245,157 @@ export class SettingsUI {
...
+          const connection = ConnectionManager.getByName(name);
+          if (connection) {
+            const storedPassword = await ConnectionManager.getStoredPassword(context, name);
+            let { data: stored, index } = connection;
 
-            const page = await new CustomUI()
-              .addInput(`host`, t(`login.host`), undefined, { default: connection.host, minlength: 1 })
-              .addInput(`port`, t(`login.port`), undefined, { default: String(connection.port), minlength: 1, maxlength: 5, regexTest: `^\\d+$` })
-              .addInput(`username`, t(`username`), undefined, { default: connection.username, minlength: 1 })
+            const ui = await new CustomUI()
```
sebjulliand, line 342: No need to `await`
```suggestion
            const ui = new CustomUI()
```

sebjulliand:
@worksofliam I'll fix a small issue here with `withContext`; in the meantime, how about adding that brilliant use of it to disable the buttons and disable `code-for-ibmi.connectTo`, `code-for-ibmi.connectToAndReload`, `code-for-ibmi.renameConnection`, `code-for-ibmi.deleteConnection` as well? WDYT?

sebjulliand (changes requested):
@worksofliam works like a charm! Brilliant cleanup!
Just see my last comment about disabling more commands to avoid potential issues and then we'll be good to go!

worksofliam (commented):
@sebjulliand I went ahead with your suggestion, and some other commands also. Thanks!

sebjulliand (changes requested):
Not sure about disabling "code-for-ibmi.showAdditionalSettings" and "code-for-ibmi.showLoginSettings"; it doesn't seem to be risky to open multiple editors (sorry...)

sebjulliand (approved):
Approved and merged! 😎

PR #1986, opened 2024-04-11
CompletionItemProvider for local actions variables

sebjulliand:
### Changes
The available variables for local actions json were provided as snippets, but they would appear in any `.json` file, not only the `.vscode/actions.js` file.

image

This PR removes these snippets and adds a CompletionItemProvider registered only for the `.vscode/actions.js` file. It allows to show the available variables using VSCode completion API.

The content assist will show up when `ctrl+space` it hit or when `&` is typed, only on the `"command"` lines of the document.
image

Triggering the content assist on any other line will not display the available variables.

@chrjorgensen I did my best for the missing DA translations, I'll let you check. Thanks 😅

### How to test this PR
1. Open or create a `.vscode/actions.js` in the current workspace
2. Place the cursor on a `"command"` line and open the list of available variables with
  a. Press `ctrl+space`
  b. Press `&`
4. Place the cursor on any other line and try to open the list of available variables. It must not show up.

### Checklist

* [x] have tested my change


worksofliam (approved):
Lovely! I love that you added some logic so it only shows up on the `command` prop. The descriptions are a nice touch. Bravo!

image

image

PR #1990, opened 2024-04-16
Debug service configuration and management improvement

sebjulliand:
### Changes
This PR fixes a number of potential issues related to the debug service and the way certificate and startup is managed.

- The certificate generated by the extension used the connection settings' hostname as the certificate password. The password is now randomly generated.
- Since the password used to be deterministic, it was re-encrypted every time the debug service was started. Re-encrypting it was mandatory since starting the debug service would destroy the password's encryption key (as designed by IBM). To avoid providing the password every time the service starts and to allow it to be started outside of VSCode, the encrypted password is now saved in the service's configuration and the encryption is backed up so it can be reused every time the `startDebugService.sh` is executed.
- The method above prevents an issue that could arise if two users with a different hostname for the same LPAR would try to start the debug service (only the user who generated the certificate could start the service).
- A new `DebugConfiguration` class has been introduced to manage the debug service configuration.
- Previously generated certificate are migrated when the connection is established (legacy certificates included)
- The `debug_service.crt` file is not deleted anymore. It is easier to keep it for clients to download it than generating it (as it requires the certificate's password). The `debug_service.crt` is re-generated when existing certificate are migrated or imported.
- The feature designed to import a certificate for the service has been fixed.
- Importing a certificate is now offered to the user if no certificate exists for the service.
- There is now a single method to check the client certificate state: it checks if it exists and if it matches the remote certificate.
- The Debugger section of the settings no longer allows to specify a certificate path and debug ports. These values are read from the service configuration.
image
- If debug is not managed, a new inline action located on the debug service node in the `IBM i debugger` view allows to open the service configuration file.
image
- Starting the Debug Service now uses `SBMJOB` to run the `startDebugService.sh` script. Starting the service will prompt for SBMJOB parameters. The prompt is prefilled with `JOBQ(QSYS/QUSRNOMAX) JOBD(QSYS/QSYSJOBD) USER(*CURRENT)`. `CMD` and `JOB` parameters are hard-coded. The user set in the USER parameter is checked as it must have `*ALLOBJ` special authority.
- With all of the above, the debug service can be started from Code for IBM i or from anywhere else just by running the `startDebugService.sh` script. It requires no external parameters or environment variables anymore. For example, it can be started using this command:
```
SBMJOB CMD(STRQSH CMD('/QOpenSys/pkgs/bin/bash -c /QIBM/ProdData/IBMiDebugService/bin/startDebugService.sh')) JOB(DBGSVCE) JOBQ(QSYS/QUSRNOMAX) JOBD(QSYS/QSYSJOBD)`
```

### How to test this PR
Before you start testing this PR, backup your existing debug service certificate so you can:
- restore it if something goes wrong 😅
- use it to test the Import feature

#### Certificate migration
1. Connect to a system with an existing debug service certificate
2. Check the certificate is migrated to the path defined in the debug service configuration file
3. Check the client certificate is generated
4. Check the configuration file is updated with the `DEBUG_SERVICE_KEYSTORE_FILE`, `DEBUG_SERVICE_KEYSTORE_PASSWORD` and `CODE4IDEBUG` values.

#### Certificate import
1. Delete the existing certificate for the service
2. Click on the action to fix the debug service in the debugger view
3. Select the import action
4. Follow the instructions on screen
5. Check if everything is ok (steps 2 to 4 above)

#### New certificate
1. Delete the existing certificate for the service
2. Click on the action to fix the debug service in the debugger view
3. Select the generate action
5. Check if everything is ok (same as above)

### Checklist
* [x] have tested my change


src/api/IBMiContent.ts
```diff
@@ -906,6 +906,10 @@ export default class IBMiContent {
     }
   }
 
+  async streamFileExists(path: string) {
```
worksofliam, line 909: Is this really any different from our `testStreamFile` API? Should we use that instead?

src/api/debug/certificates.ts
```diff
@@ -1,220 +1,268 @@
...
-export async function setup(connection: IBMi) {
-  const pw = getPasswordForHost(connection);
-  const extFileContent = await getExtFileContent(pw, connection);
-
-  if (!connection.usingBash()) {
-    if (connection.remoteFeatures[`bash`]) {
-      throw new Error(`Bash is installed on the IBM i, but it is not your default shell. Please switch to bash to setup the debug service.`);
-    } else {
-      throw new Error(`The debug service setup requires bash to be installed on the IBM i. Please install bash and try again.`);
+export async function setup(connection: IBMi, imported?: ImportedCertificate) {
```
worksofliam, line 71: Additional comments on setup would be nice. I know we're going to have to change this in the future.

It seems to me like this is entry point and what does all the setup for the debug service certificates.

src/api/debug/index.ts
```diff
@@ -267,46 +263,52 @@ export async function initialize(context: ExtensionContext) {
...
-              remoteCertsOk = true;
+            const remoteCertsExists = await certificates.remoteCertificatesExists();
+            if (remoteCertsExists) {
+              await certificates.downloadClientCert(connection);
+              vscode.window.showInformationMessage(`Debug Service Certificate already exist on the server. The client certificate has been downloaded to enable secure debugging.`);
             }
+            else {
+              const doSetup = await vscode.window.showInformationMessage(`Debug setup`, {
+                modal: true,
+                detail: `Debug service certificate does not exist on the system or it needs to be re-created. You can either import an existing certificate or generate one. Then the client certificate will be downloaded to your device.`
```
worksofliam, line 279: Should we be more explicit here about which debug certificate? E.g. server/client certificate.

src/api/debug/index.ts
```diff
@@ -267,46 +263,52 @@ export async function initialize(context: ExtensionContext) {
...
+
+              if (doSetup) {
+                try {
+                  let imported: certificates.ImportedCertificate | undefined;
+                  if (doSetup === "Import") {
+                    const localFile = (await vscode.window.showOpenDialog({
+                      canSelectFiles: true,
+                      canSelectFolders: false,
+                      canSelectMany: false,
+                      title: `Select debug service certificate`,
```
worksofliam, line 290: Perhaps 'Select debug service server certificate'?

worksofliam (changes requested):
Check out my initial questions/comments. Initial testing looks ok, but need to dive deeper into your provided scenarios.

src/api/debug/certificates.ts
sebjulliand: Agreed

src/api/debug/index.ts
sebjulliand: 👍🏻

worksofliam:
Got this issue: 

image

sebjulliand:
> Got this issue:
> ...

It's still working fine here on a couple of LPAR.
@worksofliam Is this with debug v2 installed?
The `key.properties` file is generated by the `encryptKeystorePassword.sh` script after it has been executed. `cp` is run in the debug service working directory, where the file is supposed to be found.
Can you share your output after you get that error?

worksofliam:
When trying to start the debug service as a *PGMR:

image

```
/QIBM/UserData/IBMiDebugService: export MY_JAVA_HOME="/QOpenSys/QIBM/ProdData/JavaVM/jdk80/64bit" && /QOpenSys/usr/bin/nohup "/QIBM/ProdData/IBMiDebugService/bin/startDebugService.sh"
/home/JOEUSER: LC_ALL=EN_US.UTF-8 system "call QSYS/QZDFMDB2 PARM('-d' '-i' '-t')"
Call QSYS2.QCMDEXC('CHGJOB CCSID(37)');
select distinct job_name, local_port from qsys2.netstat_job_info j where job_name = (select job_name from qsys2.netstat_job_info j where local_port = 8005 and remote_address = '0.0.0.0' fetch first row only)
{
    "code": 1,
    "signal": null,
    "stdout": "Environment variable DEBUG_SERVICE_KEYSTORE_FILE is not set.",
    "stderr": "java version \"1.8.0_401\"\nJava(TM) SE Runtime Environment (build 8.0.8.20 - pap6480sr8fp20-20240112_01(SR8 FP20))\nIBM J9 VM (build 2.9, JRE 1.8.0 OS/400 ppc64-64-Bit Compressed References 20231219_62619 (JIT enabled, AOT enabled)\nOpenJ9   - 08d2144\nOMR      - 7a5005b\nIBM      - 7394519)\nJCL - 20231221_01 based on Oracle jdk8u401-b10"
}
```

worksofliam:
@sebjulliand I am getting this in the 2.9.2 pre-release when I open the new IBM i Debug view we added. Is this something we should take care of in this PR?

image

```
/QIBM/UserData/IBMiDebugService/certs: openssl pkcs12 -in /QIBM/UserData/IBMiDebugService/certs/debug_service.pfx -passin pass:UT201P18.RCH.STGLABS.IBM.COM -info -nokeys -clcerts 2>/dev/null | openssl x509 -outform PEM
{
    "code": 1,
    "signal": null,
    "stdout": "",
    "stderr": "unable to load certificate\n4294967296:error:0909006C:PEM routines:get_name:no start line:crypto/pem/pem_lib.c:745:Expecting: TRUSTED CERTIFICATE"
}
```

worksofliam:
Here is the key issue I was getting the other day. This is using a v2 PTF:

<details>
<summary>Click me</summary>

```
/QIBM/UserData/IBMiDebugService/certs: openssl genrsa -out debug_service.key 2048 && openssl req -new -key debug_service.key -out debug_service.csr -subj '/CN=system' && openssl x509 -req -in debug_service.csr -signkey debug_service.key -out debug_service.crt -days 1095 -sha256 -req -extfile <(printf "subjectAltName=DNS:system,IP:9.5.12.241") && openssl pkcs12 -export -out debug_service.pfx -inkey debug_service.key -in debug_service.crt -password pass:O_FYmyevZLd8ae4LRb6LDfqDUtZKeUVokpZo8VcrXo15nsAnoYvW && rm debug_service.key debug_service.csr && chmod 444 debug_service.*
{
    "code": 0,
    "signal": null,
    "stdout": "",
    "stderr": "Certificate request self-signature ok\nsubject=CN = system"
}

/home/LIAM: export DEBUG_SERVICE_KEYSTORE_PASSWORD="O_FYmyevZLd8ae4LRb6LDfqDUtZKeUVokpZo8VcrXo15nsAnoYvW" && /QIBM/ProdData/IBMiDebugService/bin/encryptKeystorePassword.sh | /usr/bin/tail -n 1
{
    "code": 0,
    "signal": null,
    "stdout": "Incorrect Java version. Java 11 is required.",
    "stderr": ""
}

/QIBM/UserData/IBMiDebugService: cp key.properties .code4i.debug
{
    "code": 1,
    "signal": null,
    "stdout": "",
    "stderr": "cp: cannot stat 'key.properties': A file or directory in the path name does not exist."
}
```

</details>

sebjulliand:
> Here is the key issue I was getting the other day. This is using a v2 PTF:
> ...

@worksofliam I found the problem: I forgot to add java home to the env when running the encryption script. Since it failed, no key was generated. It's funny the script returned a code 0 and the error in stdout though...

sebjulliand:
@worksofliam I pushed a fix for your encryption issue on v2. It should be fine if you test after you pull the branch.
Also, the user that starts the service now needs to have `*ALLOBJ` to start it.

src/api/IBMi.ts
```diff
@@ -1308,5 +1327,21 @@ export default class IBMi {
       runtimeCcsid: this.runtimeCcsid,
       userDefaultCCSID: this.userDefaultCCSID,
     };
-  } 
+  }
+
+  async checkUserSpecialAuthorities(authorities: ("*ALLOBJ" | "*AUDIT" | "*IOSYSCFG" | "*JOBCTL" | "*SAVSYS" | "*SECADM" | "*SERVICE" | "*SPLCTL")[], user?: string) {
```
worksofliam, line 1332: I don't think this API should throw an error. I do think that it should instead return a valid result.

Perhaps `{valid: boolean, missing: string[]}`. I also think we need a type for the list of authorities.

worksofliam (changes requested):
Code is looking good other than one comment I left above. Check it out and let me know what you think. The last remaining thing is some additional test cases we need to ensure:

worksofliam:
Test cases:

## Ensure it works with v1 debug:

* [ ] Generating certificates working
* [ ] Using existing certificates working
* [ ] Batch debug works

## Ensure it works with v2 debug:

* [ ] Generating certificates working
* [ ] Using existing certificates working
* [ ] Batch debug works
* [ ] SEP debug works

src/api/IBMi.ts
sebjulliand: Sounds reasonable. Let's do this.

sebjulliand:
> Code is looking good other than one comment I left above. Check it out and let me know what you think. The last remaining thing is some additional test cases we need to ensure:

Thanks a lot @worksofliam !
Done as asked: the checkAuth API now returns a result. I also straightened the rights on the certificates related files.
We should be fine now 😄

worksofliam (approved):

PR #2005, opened 2024-04-24
Hide the connection browser when connecting

sebjulliand:
### Changes
This PR hides the connection browser when connecting to an IBM i. It prevents the user from clicking on the connections or doing any changes to the connections when connecting (that would be bad).

### How to test this PR
1. Connect to an IBM i
2. Connection browser must be hidden when connecting
3. It must show when disconnected

### Checklist
* [x] have tested my change


worksofliam (approved):
Looking and working good!

PR #2013, opened 2024-04-25
Fix remove shortcut text and add additional protection

chrjorgensen:
### Changes

This PR will remove the dots from the "Remove Shortcut..." men item text, since there is no additional dialog.
It will also add protection for /QSR and /QTCPTMM directories.

### How to test this PR

1. Expand the IFS browser and right-click a shortcut. The menu item should read `Remove Shortcut` without any dots.
2. Expand `/` in the IFS browser and right-click `QSR` and `QTCPTMM`. Verify no menu option `Delete...` is available.

### Checklist

* [x] have tested my change

### DIFF
--- a/package.json
+++ b/package.json
@@ -1364,7 +1364,7 @@
 			{
 				"command": "code-for-ibmi.removeIFSShortcut",
 				"enablement": "code-for-ibmi:connected",
-				"title": "Remove Shortcut...",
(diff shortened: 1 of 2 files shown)


worksofliam (approved):
Works as expected. Merge away @chrjorgensen!

PR #2015, opened 2024-04-26
Set CCSID of deployed files to 1208 if needed

sebjulliand:
### Changes
Resolves https://github.com/codefori/vscode-ibmi/issues/1995

On systems <= 7.3, the deployed files are created with a CCSID 819 (or at least different from 1208).
This PR checks (once per connection, during the first deployment - the check result is cached) if the CCSID needs to be set to 1208 after a successful deployment and runs the `setccsid` command if needed.

### How to test this PR
Here is a query to list files with CCSID != 1208:
```sql
select path_name, ccsid
from table(qsys2.ifs_object_statistics( 
           start_path_name => '/deployment/path',
           subtree_directories => 'yes'))
where ccsid != 1208
```

This should be tested on two systems:
* [x] On a system where stream files are created with CCSID 819 (IBM i OS <= 7.3)
1. Deploy the local workspace
2. Check the progress says `Fixing files CCSID...`
3. If running an action, check the following log appear in the task log: `Deployed files' CCSID set to 1208`
4. The SQL query above should not return any result

* [x] On a system where stream files are created with CCSID 1208
1. Deploy the local workspace
2. Nothing should be logged about fixing the CCSID
3. The SQL query above should not return any result

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/local/deployment.ts
+++ b/src/api/local/deployment.ts
@@ -23,6 +23,8 @@ export namespace Deployment {
   export const button = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Left, 0);
   export const workspaceChanges: Map<vscode.WorkspaceFolder, Map<string, vscode.Uri>> = new Map;
 
+  let fixCCSID: boolean | undefined;
+
   export function initialize(context: vscode.ExtensionContext) {
     button.command = {
       command: `code-for-ibmi.launchDeploy`,
@@ -89,6 +91,7 @@ export namespace Deployment {
     });
 
     instance.onEvent("disconnected", () => {
+      fixCCSID = undefined;
       button.hide();
     })
   }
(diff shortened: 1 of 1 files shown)


edmundreinhardt:
I have tested and this works!

edmundreinhardt:
Is there a reason why this is targetting 2.11 instead of 2.10?

worksofliam:
@edmundreinhardt 2.10.0 is basically finalized and we only need to do doc changes now.

edmundreinhardt:
Just a shame because this is important to supporting IFS deployment.  We were hitting this quite a bit on our test IBM i machines.  For Code for i to support local development on 7.3 for example, this is critical.  It is a small PR and should hopefully not be a big risk.

worksofliam:
@edmundreinhardt I am going to test it today 😄

worksofliam:
Tested on three systems and it works as expected. Good work @sebjulliand!

worksofliam (approved):

PR #2032, opened 2024-05-06
Don't use library list to validate current library change

sebjulliand:
### Changes
Fixes https://github.com/codefori/vscode-ibmi/issues/2031

When changing the current library, `CHGCURLIB` is run to validate the new current library. If for some reason, the current `current library` cannot be accessed, `CHGCURLIB` will fail as the library list is wrong (i.e. it contains a library that cannot be accessed). This prevents the current library from being changed to fix the library list.

This PR changes the way `CHGCURLIB` is called so it doesn't use the library list.

### How to test this PR

Examples:
1. Create a temporary library
2. Set it as the current library
3. Delete the temporary library
4. Reconnect
5. Change the current library

Before the PR, step 5 would fail. With the PR, step 5 runs fine.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/views/LibraryListView.ts
+++ b/src/views/LibraryListView.ts
@@ -4,7 +4,7 @@ import { instance } from "../instantiate";
 import { t } from "../locale";
 import { WithLibrary } from "../typings";
 
-export class LibraryListProvider implements vscode.TreeDataProvider<LibraryListNode>{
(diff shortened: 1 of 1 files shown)


chrjorgensen (approved):
@sebjulliand Simple and effective fix - good catch!! 👍

PR #2038, opened 2024-05-07
Open path from URI

worksofliam:
### Changes

* [x] Add new url for opening files at a specific path
* [x] Add improved error handling in the new URI
* [x] When the host is provided, but the user is not connected, then automatically connect if the connection is defined
    * [x] Depends on #1975 
* [x] Nice to have: make sure Sandbox mode still works

### How to test this PR

* `vscode://halcyontechltd.code-for-ibmi/open?path=/home/liama/xyz`
* `vscode://halcyontechltd.code-for-ibmi/open?host=1.1.1.1&path=/home/liama/xyz`

Examples:

1. Run the test cases

### Checklist

* [ ] have tested my change


julesyan:
If they have no current connection would it beneficial to allow the user to pick one? Instead of a warning message. 

Also I am unsure the purpose of the host param as the below two scenarios arent covered:
- **Scenario 1**: if you arent connected it doesnt auto connect to the given host, isntead it just gives a warning it cannot open the file. I think it should ask to auto connect if the given host is on the list. If the host is listed more than once then give me the option to pick which connection to initiate
- **Scenario 2**: if you are connected to the wrong host, it just prompts to open with the current connection. We should also have the option to switch to the correct connection, again if there is more that one listed conneciton to the same host then allow me to pick which connection to initiate

worksofliam:
@julesyan Thanks for your great testing.

1. Yes, we should automatically connect to the host given it is in the list.
2. Right now we don't even have the internal logic to disconnect and reconnect to a new system automatically. Let me see if this is even possible (since we do have APIs to do both separately.)

julesyan:
Tested locally and found no more issues, looks good!

src/uri/index.ts
```diff
@@ -0,0 +1,153 @@
...
+
+                const paths = Array.isArray(queryData.path) ? queryData.path : [queryData.path];
+                for (const path of paths) {
+                  commands.executeCommand(`code-for-ibmi.openEditable`, path);
+                }
+              } else {
+                window.showWarningMessage(t(`uriOpen.missingPath`));
+              }
+            } else {
+              window.showWarningMessage(t(`uriOpen.noConnection`));
```
SanjulaGanepola, line 69:
```suggestion
              window.showWarningMessage(t(`uriOpen.missingPath`));
```

src/uri/index.ts
```diff
@@ -0,0 +1,153 @@
...
+                    return; 
+                  }
+                }
+
+                const paths = Array.isArray(queryData.path) ? queryData.path : [queryData.path];
+                for (const path of paths) {
+                  commands.executeCommand(`code-for-ibmi.openEditable`, path);
+                }
+              } else {
+                window.showWarningMessage(t(`uriOpen.missingPath`));
```
SanjulaGanepola, line 66: Given the example you gave in this PR:
```
vscode://halcyontechltd.code-for-ibmi/open?path=/home/liama/xyz
```
I assume this means that the file will be opened based on the active connection. If so, is this else block a mistake because this will not open any file and just give this warning even though I am connected. In the case there is no connection, shouldn't the message be:
```ts
window.showWarningMessage(t(`uriOpen.noConnection`));
```

SanjulaGanepola (changes requested):
@worksofliam I did some testing with this change and noticed a few issues. Refer to my comments below.

varunrao1991:
Can you create an artefact for this PR?

sebjulliand:
> This works wonderfully with vscode.open command but fails with vscode.diff when you hand in 2 uris. "open <-> open" error.
> 
> I have not put a debugged on the code to try to understand why.

What URIs are you using with `vscode.diff`?

sebjulliand:
Well, the `/open` URI is not meant to be called like this, I'm afraid. Its main purpose it to be able to open an editor on a member/IFS File, and connect to a system if needed, in one single call. It could be used to open a source member from an external Web application, for example.

`vscode.open` and `vscode.diff` expect URis that are provided by a VS Code File System. Code for IBM i provides two file systems: `member://` and `streamfile://`. You can refer to [this documentation](https://codefori.github.io/docs/dev/examples/#get-members-and-streamfiles) to see how these URI are built.

If your goal is to copy streamfiles/members locally, without opening an editor, then you could:
- Do it without using the exported API, just by using the `member` or `streamfile` URI described above in conjunction with VS Code's `vscode.workspace.fs.copy` method.
- With [Code for i exported API](https://codefori.github.io/docs/dev/api/), once you get the reference to `instance` as described in the documentation, use `instance.getConnection()?.getContent().downloadStreamfileRaw` or `instance.getConnection()?.getContent().downloadMemberContent`. You'll need a bit of plumbing using VS Code API to save the content locally, but it should be fairly easy.

Hope this helps!

PR #2041, opened 2024-05-08
Get asp name when listing objects

sebjulliand:
### Changes
Fixes https://github.com/codefori/vscode-ibmi/issues/2039

This PR updates the queries used when listing objects so the IASP name can be resolved if needed.
It also adds the IASP name in the tooltip, if not undefined:
image

### How to test this PR
1. Connect to a system with an IASP
2. Expand a filter that lists physical source files
3. Display the physical source files tooltip.
4. The members count and CCSID must be OK

With the PR, the members count is wrong and CCSID shows `?`

### Checklist
* [x] have tested my change


worksofliam:
@sebjulliand Sadly I don't have a system with an iASP setup. 

@chrjorgensen Is this one you can help me with?

PR #2042, opened 2024-05-08
Refresh file

julesyan:
### Changes
Added the ability to refresh an IFS file. Button is located in the top right of the editor, similar to Run Actions. Making use of the pre-existing command `workbench.action.files.revert` from VSCode

### How to test this PR
1. Open an IFS file
2. Make/Trigger changes to the file elsewhere 
3. Click refresh button, file should update with latest content

### Checklist
* [x] have tested my change

### DIFF
--- a/package.json
+++ b/package.json
@@ -2334,7 +2334,7 @@
 				},
 				{
 					"command": "workbench.action.files.revert",
-					"when": "code-for-ibmi:connected && resourceScheme == streamfile",
+					"when": "code-for-ibmi:connected && (resourceScheme == streamfile || resourceScheme == member)",
 					"group": "navigation@4"
 				}
 			],
--- a/package.json
+++ b/package.json
@@ -1631,11 +1631,11 @@
 				"icon": "$(settings-gear)",
 				"enablement": "!code-for-ibmi:debugWorking"
 			},
-            {
-                "command": "code-for-ibmi.refreshFile",
-                "title": "Refresh File",
-                "category": "IBM i",
-                "icon": "$(refresh)"
+			{
+				"command": "code-for-ibmi.refreshFile",
+				"title": "Refresh File",
+				"category": "IBM i",
+				"icon": "$(refresh)"
 			},
 			{
 				"command": "code-for-ibmi.searchObjectBrowser",
@@ -1650,7 +1650,7 @@
 				"category": "IBM i",
 				"icon": "$(search-fuzzy)",
 				"enablement": "code-for-ibmi:connected"
-            }
+			}
 		],
 		"keybindings": [
 			{
--- a/package.json
+++ b/package.json
@@ -1630,7 +1630,13 @@
 				"category": "IBM i",
 				"icon": "$(settings-gear)",
 				"enablement": "!code-for-ibmi:debugWorking"
-			}
+			},
+            {
+                "command": "workbench.action.files.revert",
+                "title": "Refresh File",
+                "category": "IBM i",
+                "icon": "$(refresh)"
+            }
 		],
 		"keybindings": [
 			{
@@ -2325,6 +2331,11 @@
 					"command": "code-for-ibmi.clearDiagnostics",
 					"when": "code-for-ibmi:connected",
 					"group": "navigation@3"
+				},
+				{
+					"command": "workbench.action.files.revert",
+					"when": "code-for-ibmi:connected && resourceScheme == streamfile",
+					"group": "navigation@4"
 				}
 			],
 			"view/item/context": [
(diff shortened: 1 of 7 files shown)


chrjorgensen (changes requested):
@julesyan Thank you very much for your first contribution - a great idea and implementation! 😍 

One small addition to make it work for members also - and I think we're good to go... 🙏

chrjorgensen (changes requested):
Excellent - works for members now also. 👍 

One caveat that needs to be addressed before we can send it out: All changes in a member or streamfile is lost without warning, when the refresh icon is pressed!

Is there a way to warn the user about this? And save or discard the changes before refreshing?

worksofliam:
@chrjorgensen @julesyan 

> Is there a way to warn the user about this? And save or discard the changes before refreshing?

This is a great point. I am guessing we might have to define our own command, which wraps a warning if the document is dirty before executing the command programmatically.

Julia, take a stab at that if you want - and feel free to reach out and I can offer my help whenever!

src/instantiate.ts
```diff
@@ -700,7 +700,20 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
       vscode.commands.executeCommand(`code-for-ibmi.openEditable`, item.path, { readonly });
     }),
     vscode.commands.registerCommand("code-for-ibmi.updateConnectedBar", updateConnectedBar),
-  );
+    
+    vscode.commands.registerCommand("code-for-ibmi.refreshFile", async () => {
```
worksofliam, line 704: I believe there is a parameter here, of type `vscode.Uri` because it is being called from `editor/title`.

src/instantiate.ts
```diff
@@ -700,7 +700,20 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
...
+        const editor = vscode.window.activeTextEditor;
```
worksofliam, line 705: I think instead of using `activeTextEditor`, you should get the editor based on the provided URI parameter. 

We have `Tools.findExistingDocumentUri` to get a document based on URI.

src/instantiate.ts
julesyan: So I also let it stay in the command palette, I forgot to mention that in the PR. Does that method also give in the same params?

src/instantiate.ts
```diff
@@ -700,7 +700,29 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
...
+        doc = Tools.findExistingDocument(uri);
+      } else {
+        const editor = vscode.window.activeTextEditor;
+        doc = editor?.document;
+      }
+
+      if (doc?.isDirty) {
+        vscode.window
+          .showWarningMessage(`Your changes will be discarded`, `Continue`, `Cancel`)
+          .then(result => {
```
worksofliam, line 716: I believe this should be a modal instead of a traditional warning message.

The second parameter can be a `MessageOptions` object.

package.json
```diff
@@ -2357,6 +2363,11 @@
 					"command": "code-for-ibmi.clearDiagnostics",
 					"when": "code-for-ibmi:connected",
 					"group": "navigation@3"
+				},
+				{
+					"command": "code-for-ibmi.refreshFile",
+					"when": "code-for-ibmi:connected && (resourceScheme == streamfile || resourceScheme == member)",
```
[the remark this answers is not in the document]
worksofliam: Yup, you're right. Ignore this!

src/instantiate.ts
```diff
@@ -700,7 +700,32 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
...
+        const editor = vscode.window.activeTextEditor;
+        doc = editor?.document;
+      }
+
+      if (doc?.isDirty) {
+        vscode.window
+          .showWarningMessage(
+            `Your changes will be discarded`, 
+            { modal: true }, 
+            `Continue`)
```
worksofliam, line 718: I think your string literals here need to use our translate API. I believe 'Continue' has an id, but might need a new one for the label.

* See `en.json` and add a prop for the label
* Use the `t` API with the IDs in place of the string literals. (`t` comes from `locale.ts`)

worksofliam (changes requested):
Functionality is perfect. Last thing we need to do is fix up the locale and then you're done!

sebjulliand:
> @chrjorgensen @sebjulliand Any chance you can add `discard.changes` for FR and DA when you get a chance? Thanks!

Sure:
```
"discard.changes": "Vos modifications vont être perdues",
```

chrjorgensen (approved):
@julesyan Finally added the danish translation - sorry for the delay. 😞 

I have tested and reviewed this PR and it works great! Let's get this merged! 😍

PR #2048, opened 2024-05-10
Change where manage debug context is set

worksofliam:
### Changes

On the edge case when other extensions are setting the `DEBUG_MANAGED` variable, we now set the context when the connection is made.

### How to test this PR

Examples:

1. Set the environment variable `DEBUG_MANAGED` before connecting to a system and expect the debug setup view to be hidden.

### Checklist

* [x] have tested my change


SanjulaGanepola (approved):
Works as expected. Thanks for the fix

Issue #2069, opened 2024-05-23
Connection to IBM i keeps disconnecting. Error in dev tools console.


KerimG:
After successfully connecting to the IBM i, I get a pop-up saying "Connection lost" with the following error in the console. Not sure if the error is related, though. 

```
mainThreadExtensionService.ts:82 Error: No such file
	at 101 (c:\Users\VRPG\.vscode\extensions\halcyontechltd.code-for-ibmi-2.10.1\dist\extension.js:2:232673)
	at SFTP.push (c:\Users\VRPG\.vscode\extensions\halcyontechltd.code-for-ibmi-2.10.1\dist\extension.js:2:249360)
	at CHANNEL_DATA (c:\Users\VRPG\.vscode\extensions\halcyontechltd.code-for-ibmi-2.10.1\dist\extension.js:2:167608)
	at 94 (c:\Users\VRPG\.vscode\extensions\halcyontechltd.code-for-ibmi-2.10.1\dist\extension.js:2:332785)
	at e.exports.U (c:\Users\VRPG\.vscode\extensions\halcyontechltd.code-for-ibmi-2.10.1\dist\extension.js:2:193609)
	at F.decrypt (c:\Users\VRPG\.vscode\extensions\halcyontechltd.code-for-ibmi-2.10.1\dist\extension.js:2:289945)
	at e.exports.$ [as _parse] (c:\Users\VRPG\.vscode\extensions\halcyontechltd.code-for-ibmi-2.10.1\dist\extension.js:2:193133)
	at e.exports.parse (c:\Users\VRPG\.vscode\extensions\halcyontechltd.code-for-ibmi-2.10.1\dist\extension.js:2:197520)
	at Socket.<anonymous> (c:\Users\VRPG\.vscode\extensions\halcyontechltd.code-for-ibmi-2.10.1\dist\extension.js:2:169786)
	at Socket.emit (node:events:517:28)
	at addChunk (node:internal/streams/readable:335:12)
	at readableAddChunk (node:internal/streams/readable:308:9)
	at Readable.push (node:internal/streams/readable:245:10)
	at TCP.onStreamRead (node:internal/stream_base_commons:190:23)
```

**Expected behavior**
I expected the connection to maintain.

**Screenshots**
![image](https://github.com/codefori/vscode-ibmi/assets/9061713/ce27348e-590c-4462-bfbc-b11d3ca6b389)

**Environment (please complete the following information):**
 - Extension version 2.10.1
[...]

PR #2077, opened 2024-05-27
Added Keepalive interval login settings

sebjulliand:
### Changes
Resolves https://github.com/codefori/vscode-ibmi/issues/2074

This PR adds a new input to the Logins settings to let the user define a custom SSH `Keepalive interval`.
image

### How to test this PR
Set a breakpoint in src/api/IBMi.ts on line 126.

1. Open an existing connection
2. Define a custom Keepalive interval
3. Connect
4. Check the `connectionObject.keepaliveInterval` field is set with the custom value.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -123,7 +123,7 @@ export default class IBMi {
   async connect(connectionObject: ConnectionData, reconnecting?: boolean, reloadServerSettings: boolean = false): Promise<{ success: boolean, error?: any }> {
     return await Tools.withContext("code-for-ibmi:connecting", async () => {
       try {
-        connectionObject.keepaliveInterval = 35000;
+        connectionObject.keepaliveInterval = connectionObject.keepaliveInterval === undefined ? 35000 : connectionObject.keepaliveInterval;
 
         configVars.replaceAll(connectionObject);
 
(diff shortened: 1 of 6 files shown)


chrjorgensen:
@sebjulliand Code works as expected - but the setting does not prevent the disconnect, unfortunately... 😞 

I'm not sure how or if we can fix #2069... would have thought the keepalive setting was the solution, but apparantly not...

Maybe you should break this PR into two PR and isolate the other fix for private key path in its own PR?

sebjulliand:
> Maybe you should break this PR into two PR and isolate the other fix for private key path in its own PR?
@chrjorgensen absolutely. I removed the private key fix from here and put it there: https://github.com/codefori/vscode-ibmi/pull/2082

chrjorgensen:
> absolutely. I removed the private key fix from here and put it there: #2082

@sebjulliand Great! I changed this PR to draft... when we find a solution we can continue here if necessary...

worksofliam (changes requested):
Like I said, not very good. Work on this more.

worksofliam:
@sebjulliand is this actually ready to be reviewed?

sebjulliand:
@worksofliam welp, it was changed to draft since it doesn't seem to fix the original issue, even though the keepalive is adjusted (checkout @chrjorgensen comment above).

worksofliam:
@sebjulliand I suspect it's also related to the timeout as set by the OS. @chrjorgensen What do you see with `CHGTCPA`? I believe there is a timeout option there too which might be similar.

PR #2082, opened 2024-05-29
Fixed privatekeypath update

sebjulliand:
### Changes
This PR fixes an issue where the settings' `privateKeyPath` would be set to `''` instead of `undefined` after an update, crashing the connection process afterwards with this message:
image

### How to test this PR
Before the PR, making any change not related to the password or private key in the settings would make the next password connection attempts to crash.

After the PR, connection attempts are OK after a change in the settings.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/webviews/settings/index.ts
+++ b/src/webviews/settings/index.ts
@@ -368,7 +368,7 @@ export class SettingsUI {
 
                   default:
                     if (data.password) {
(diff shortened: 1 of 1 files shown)


worksofliam (approved):
Followed your steps to recreate the error.

image

PR solved the issue. Thanks!

PR #2085, opened 2024-05-29
Fix clearing of multiple diagnostics

### LINKED ISSUE #2084: `Clear Diagnostic on Edit` setting fails to clear multiple diagnostics
I noticed a bug with the `Clear Diagnostic on Edit` setting where if there were multiple diagnostics and a change was made in a range that impacted multiple of them, only the first diagnostic was cleared. Refer to the example below:

Two diagnostics:
image

Still one diagnostic after making a fix in the text range of both diagnostics:
image

SanjulaGanepola:
### Changes

Fixes #2084

With this PR, whenever the `Clear Diagnostic on Edit` setting is enabled and a text document is changed, **all** (instead of just the first one) diagnostics are cleared when the diagnostics range contains the text document's change range.

### How to test this PR

1. Enable the `Clear Diagnostic on Edit` setting
2. Git clone [ibmi-company_system](https://github.com/IBM/ibmi-company_system)
3. Create an error on line 7 (refer to the linked issue)
4. Compile the file and observe 2 diagnostics in this range.
5. Fix the error and the 2 diagnostics should be cleared.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/errors/diagnostics.ts
+++ b/src/api/errors/diagnostics.ts
@@ -52,13 +52,8 @@ export function clearDiagnostic(uri: vscode.Uri, changeRange: vscode.Range) {
   const currentList = ileDiagnostics.get(uri);
 
   if (currentList) {
-    const existing = currentList.findIndex(d => d.range.contains(changeRange));
-
-    if (existing >= 0) {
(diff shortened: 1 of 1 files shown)


worksofliam (approved):
Works as expected. Thanks for a good and simple fix!

PR #2096, opened 2024-06-03
Delete stale cached server settings

SanjulaGanepola:
### Changes

- Fixed bullet 3 of https://github.com/codefori/vscode-ibmi/issues/2088 by updating the current listener on `code-for-ibmi.connections` to detect and delete stale cached server settings. Also, refresh the connection browser to remove any deleted connections
- Delete cached server settings when login settings are changed 

### How to test this PR

Examples:

1. Create a new connection and connect so there exists some cached server settings.
2. Manually delete the connection from the `settings.json` file.
3. Observe that the connection is removed from the connection browser view.
4. Create a new connection with the same name.
5. Debug the extension and observe that connecting to the new connection does not reuse the old cached server settings. 
6. Disconnect and use the `Login Settings` action to change to another host or user profile.
7. Repeat step 5.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/Storage.ts
+++ b/src/api/Storage.ts
@@ -1,11 +1,13 @@
 import vscode from 'vscode';
+import { ConnectionData } from '../typings';
 
 const PREVIOUS_CUR_LIBS_KEY = `prevCurLibs`;
 const LAST_PROFILE_KEY = `currentProfile`;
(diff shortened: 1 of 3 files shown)


chrjorgensen (approved):
@SanjulaGanepola Works as expected - thanks for the fix, good catch! 👍

PR #2102, opened 2024-06-07
Only show server ports in Debug Service tooltip

sebjulliand:
### Changes
In its current state, the `Debug Service` item tooltip in the `IBM i Debugger` view shows too many ports:
toomany

The ports besides 8001 and 8005 are those used by the Service when it connects to the local TCP servers; they only appear once the Debug Service has been used at least once to debug a program:
image

This PR fixes the query used to fetch the Service ports to only get the server ports.
Code_Z7F1bRUJgk

### How to test this PR
1. Connect to an IBM i where the Debug Service runs and has been used at least once
2. Go to the `IBM i Debugger` view
3. Hover over the `Debug Service` item
4. Only two ports must be shown at the top of the tooltip

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/debug/server.ts
+++ b/src/api/debug/server.ts
@@ -112,7 +112,7 @@ export async function stopService(connection: IBMi) {
 export async function getDebugServiceJob() {
   const connection = instance.getConnection();
   if (connection) {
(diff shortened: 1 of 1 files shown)


worksofliam (approved):

sebjulliand:
@worksofliam oof, I forgot about it 😅

PR #2107, opened 2024-06-10
Fix object library being overwritten to current library for debug

SanjulaGanepola:
### Changes

@sebjulliand In relation to the changes in https://github.com/codefori/vscode-ibmi/pull/2079, we noticed in Project Explorer (https://github.com/IBM/vscode-ibmi-projectexplorer/issues/486) that the current library was always being used to debug a program even though it belonged to another library.

### How to test this PR

1. Run debug from Project Explorer.
2. Run debug from Object Browser.
3. Run debug from active editor (local file, stream file, member).

### Checklist

* [x] have tested my change


sebjulliand (approved):
Good catch! Thanks!

PR #2121, opened 2024-06-17
Add 5250 keyboard reset button

worksofliam:
### Changes

Adds a button to the terminal view that is displayed when the user attempts to type in an area where they cannot in 5250.

Not many users know the tn5250 shortcut (Control+R) to do a reset, so I added a button to the UI for it.

### How to test this PR

Examples:

1. Connect and launch a 5250
2. Sign in
3. Attempt to write outside an input field
4. See the button appear in the terminal tab
5. Click the button and see the cursor position reset

### Checklist

* [x] have tested my change

### DIFF
--- a/package.json
+++ b/package.json
@@ -1666,7 +1666,7 @@
 			},
 			{
 				"command": "code-for-ibmi.term5250.systemAttention",
-				"title": "5250 Attention Request",
+				"title": "Reset keyboard",
 				"category": "IBM i",
 				"icon": "$(warning)",
 				"enablement": "code-for-ibmi:connected"
--- a/package.json
+++ b/package.json
@@ -1665,10 +1665,10 @@
 				"enablement": "code-for-ibmi:connected"
 			},
 			{
-				"command": "code-for-ibmi.term5250.systemAttention",
+				"command": "code-for-ibmi.term5250.resetPosition",
 				"title": "Reset keyboard",
 				"category": "IBM i",
-				"icon": "$(warning)",
+				"icon": "$(keyboard)",
 				"enablement": "code-for-ibmi:connected"
 			}
 		],
@@ -2363,7 +2363,7 @@
 					"when": "view == ibmiDebugBrowser"
 				},
 				{
-					"command": "code-for-ibmi.term5250.systemAttention",
+					"command": "code-for-ibmi.term5250.resetPosition",
 					"group": "navigation",
 					"when": "code-for-ibmi:term5250Halted"
 				}
--- a/package.json
+++ b/package.json
@@ -1663,6 +1663,13 @@
 				"category": "IBM i",
 				"icon": "$(search-fuzzy)",
 				"enablement": "code-for-ibmi:connected"
+			},
+			{
+				"command": "code-for-ibmi.term5250.systemRequest",
+				"title": "5250 System Request",
+				"category": "IBM i",
+				"icon": "$(warning)",
+				"enablement": "code-for-ibmi:connected"
 			}
 		],
 		"keybindings": [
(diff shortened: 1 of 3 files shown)


worksofliam:
Fixed the wording: 

image

src/api/Terminal.ts
```diff
@@ -31,6 +32,35 @@ export namespace Terminal {
...
+          const ifsPath = (await content.isDirectory(ifsNode.path)) ? ifsNode.path : path.dirname(ifsNode.path);
+          const terminal = await selectAndOpen(instance, TerminalType.PASE);
+          terminal?.sendText(`cd ${ifsPath}`);
+        }
+      }),
+
+      vscode.commands.registerCommand(`code-for-ibmi.term5250.systemAttention`, () => {
+        const term = vscode.window.activeTerminal;
+        if (term) {
+          term.sendText(Buffer.from([1, 9]).toString(), false);
```
chrjorgensen, line 57: Where does `Buffer` come from?

package.json
```diff
@@ -1663,6 +1663,13 @@
 				"category": "IBM i",
 				"icon": "$(search-fuzzy)",
 				"enablement": "code-for-ibmi:connected"
+			},
+			{
+				"command": "code-for-ibmi.term5250.systemAttention",
+				"title": "5250 Attention Request",
```
chrjorgensen, line 1669:
```suggestion
				"title": "Reset keyboard",
```

src/api/Terminal.ts
```diff
@@ -71,13 +101,20 @@ export namespace Terminal {
     }
   }
 
+  const HALT_START = `[24;2H(B[0;1m[37m[40m`;
```
chrjorgensen, line 104: Why does these lines look like this in VS Code?

billede

chrjorgensen (changes requested):
@worksofliam The code works (I don't know why - see my comment on `Buffer`), but the button is only available on the login screen, not any subsequent panels.

I also think there are still some misunderstandings / wrong wordings here:

- The Attn (attention) key _does_ reset the keyboard when locked - but it does more than that: It calls the program set by the SETATNPGM command in the job.
- What you're after is the Reset button (Ctrl-R) - it resets the keyboard and nothing more.
- The title of the VS Code command should be changed as commented.
- The names used for VS Code commands and TypeScript functions should reflect this.
- The title of this PR should reflect this.

👍

src/api/Terminal.ts
worksofliam: It's a Node.js built-in! https://nodejs.org/api/buffer.html

chrjorgensen: Ah - so you're sending bytes 0x01 and 0x09 to the terminal, or how should I interpret it? I'm not jetlagged but I don't get it... it's not obvious to me! 😆

chrjorgensen: Maybe some constants to explain what these codes do would help on my understanding... 🙄 😄

src/api/Terminal.ts
worksofliam: Because these are bytes which don't have a character representation.

src/api/Terminal.ts
worksofliam: Done!

chrjorgensen (changes requested):
@worksofliam Almost there... 😉

It is named correctly and works on all panels. 👍 
Only remaining issues are
- the Halt status is not reset if user press Ctrl-R and unlocks the keyboard
- you test for `II` - so pressing `Enter` on any panel containing `II` wil cause the Halt status:
billede
Maybe better to check for `5250(5blanks)X II`?
billede

chrjorgensen (approved):
@worksofliam Approved and merged. 👍

PR #2122, opened 2024-06-18
Add tooltips to libraries in the user library list view

SanjulaGanepola:
### Changes

- Add tooltips for libraries in the user library list view (just like the object browser)
  image
- Add tooltip related APIs in `IBMiContent` to be used by Project Explorer:
  - `objectToToolTip`
  - `sourcePhysicalFileToToolTip`
  - `memberToToolTip`
  - `ifsFileToToolTip`

### How to test this PR

Check the following tooltips are rendering correctly

1. User Library List View - Libraries
2. Object Browser - Objects, Source Physical Files, Members
3. IFS Browser - IFS files/directories

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -3,7 +3,7 @@ import fs from 'fs';
 import path from 'path';
 import tmp from 'tmp';
 import util from 'util';
-import { window } from 'vscode';
+import { MarkdownString, window } from 'vscode';
 import { ObjectTypes } from '../filesystems/qsys/Objects';
 import { AttrOperands, CommandResult, IBMiError, IBMiMember, IBMiObject, IFSFile, QsysPath } from '../typings';
 import { ConnectionConfiguration } from './Configuration';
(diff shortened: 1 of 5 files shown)


chrjorgensen (changes requested):
@SanjulaGanepola Some issue were found:

- iasp info is shown in tootip even when undefined:
billede

- ASP info is shown in object tooltip - should not be there:
billede

- the text `iasp` is not found in the locale files. Please add this.
EDIT: the text `iasp` is indeed in the locale files... don't know why it is shown as `!iasp!` then...?

SanjulaGanepola:
@chrjorgensen Can you take a look at the following comments:

> - iasp info is shown in tootip even when 

That is very odd since it doesn't show up for me when it is `undefined` as the `generateTooltipHtmlTable` function filters out any `undefined` fields. Can you check if the result you are getting back is maybe a string somehow (`"undefined"`)? FYI, I also noticed a similar issue where the `Created by` field had values of `""` so I filtered those out -> [4bb5955](https://github.com/codefori/vscode-ibmi/pull/2122/commits/4bb59558be292f6dc994d5d269d0e78d35728d5c)
image
image

> - ASP info is shown in object tooltip - should not be there:

For the same reason as above, I do not see this as well unless it is an object in a library on the ASP. When you say "should not be there", do you mean that `IASP` information should not be on any object unless the type is `*LIB` or did you mean don't show it when `undefined` which would be related to the above issue.

> - the text iasp is not found in the locale files. Please add this.
EDIT: the text iasp is indeed in the locale files... don't know why it is shown as !iasp! then...?

I am not able to reproduce this behavior either :confused:
image

chrjorgensen (approved):
@SanjulaGanepola Looking good to me, all tooltips are rendered as expected! 👍

PR #2135, opened 2024-06-21
Add find of filenames to IFS browser

chrjorgensen:
### Changes

This PR will add a `Find...` option to IFS shortcuts and directories to allow searching for patterns in filenames in the IFS.

### How to test this PR

Examples:

1. Right-click a directory and click `Find...`
2. Enter a term to look for in filenames, press `Enter`
3. Search will start in the chosen directory. If any filenames are found, a new view `IBM I FIND: RESULTS` will open with a list of the files having matching name.

### Checklist

* [x] have tested my change


chrjorgensen:
@sebjulliand Nice to finally see it in action! 😄 

It's all your for further refinement - view merge etc. Maybe a `Show in IFS browser` context menu option to easily go back to the IFS browser and work on the found file?

When your work is done, we also need some translation work for the new texts.

Thanks for your help with this!! ❤️

sebjulliand:
@chrjorgensen here we go!
I merged both views, refactored some pieces of code here and there and added a result message for each use case (find and search).

image
image

Give it a go and let me know what you think!

chrjorgensen:
@sebjulliand Looks good! 👍 

I added Danish translations and fixed a bug retrieving previous find terms...

...and while on the subject of previous find and search terms, they don't seem to work - and haven't for a long time,
Anything I did wrong in `getPreviousSearchTerms` and `getPreviousFindTerms`? I put them in global storage to be shared between connections, and I had it working at one time - but now they don't work anymore... more help is needed here, I guess... 🙄

sebjulliand:
@chrjorgensen you're up! It should be OK now.
I also mentioned in the error message that `findutils` is required to run this feature (since it's the actual package that needs to be installed).

PR #2142, opened 2024-06-26
Fix `countFiles` when directory does not exist

SanjulaGanepola:
### Changes

The `countFiles` function ends up counting the number of files in the home directory when the directory that it is given does not exist. Although I have not noticed any issue in Code4i because of this, it causes an issue in Project Explorer (as it calls `getDeployCompareFiles`) where it shows that it is going to delete the contents of my home directory:

image

Without this fix, the issue can also be observed in the updated test where the `countFiles` result ends up counting the number of files in my home directory rather than the directory provided:

image

### How to test this PR

1. Use the test case provided
2. Use Project Explorer:
  - Deploy your workspace
  - Delete the deploy location using the IFS browser
  - Refresh the source in the Project Explorer and observe the contents of the home directory is not visible unlike the first image above

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -947,6 +947,6 @@ export default class IBMiContent {
   }
 
   async countFiles(directory: string) {
-    return Number((await this.ibmi.sendCommand({ command: `ls | wc -l`, directory })).stdout.trim());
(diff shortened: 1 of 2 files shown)


worksofliam (approved):
Code looks good and test is passing. Thanks!

PR #2146, opened 2024-06-30
Use default open mode in IFS search and find results

chrjorgensen:
### Changes

This PR will add a `readonly` value to IFS search and find results, to honor the default open mode set in the configurations.
When default open mode is `browse`, clicking on a search/find result will open the file in readonly mode.

### How to test this PR

Examples:

1. Set the default open mode to `browse` in the extension configuration
2. Create a search or find result
3. Click on one of the search/find results
4. Confirm the file is opened on readonly mode
5. Change the default open mode to `edit` in the extension configuration
6. Re-run steps 2 and 3
7. Confirm the file is opened in edit mode

### Checklist

* [x] have tested my change

### DIFF
--- a/src/instantiate.ts
+++ b/src/instantiate.ts
@@ -672,12 +672,12 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
       return vscode.commands.executeCommand("code-for-ibmi.openWithDefaultMode", item, "edit" as DefaultOpenMode);
     }),
 
-    vscode.commands.registerCommand("code-for-ibmi.openWithDefaultMode", (item: WithPath, overrideMode?: DefaultOpenMode) => {
+    vscode.commands.registerCommand("code-for-ibmi.openWithDefaultMode", (item: WithPath, overrideMode?: DefaultOpenMode, position?: vscode.Range) => {
       const readonly = (overrideMode || GlobalConfiguration.get<DefaultOpenMode>("defaultOpenMode")) === "browse";
-      vscode.commands.executeCommand(`code-for-ibmi.openEditable`, item.path, { readonly });
+      vscode.commands.executeCommand(`code-for-ibmi.openEditable`, item.path, { readonly, position } as OpenEditableOptions);
     }),
     vscode.commands.registerCommand("code-for-ibmi.updateConnectedBar", updateConnectedBar),
-    
+
     vscode.commands.registerCommand("code-for-ibmi.refreshFile", async (uri?: vscode.Uri) => {
       let doc: vscode.TextDocument | undefined;
       if (uri) {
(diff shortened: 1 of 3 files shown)


sebjulliand (changes requested):
Good catch!
The implementation is correct, but it would be better to simply call the `code-for-ibmi.openWithDefaultMode` command with just the path as the first argument and no other options,instead of `code-for-ibmi.openEditable` here: https://github.com/codefori/vscode-ibmi/blob/ab4d38c4cc20e7139726850d17edd465dd48013c/src/views/searchView.ts#L90

It does what you did here, but it keeps the logic in one single place :)

chrjorgensen:
> but it would be better to simply call the `code-for-ibmi.openWithDefaultMode` command

That was exactly my first attempt... but then the `readonly` attribute from the result view would be ignored, and we probably would like to propagate the value when the server or directory being searched is protected (would require further changes...), so I opted for this intermediate solution.

WDYT?

sebjulliand:
Hmmm...`code-for-ibmi.openWithDefaultMode` calls `code-for-ibmi.openEditable` with the default open mode.
`code-for-ibmi.openEditable` also checks if the path is protected, or even writable, etc. 

Actually, just swapping `code-for-ibmi.openEditable` with `code-for-ibmi.openWithDefaultMode` and keeping both arguments should do the trick. That will open using the default mode BUT honor the readonly flag in priority if it is `true`.

chrjorgensen:
> Actually, just swapping `code-for-ibmi.openEditable` with `code-for-ibmi.openWithDefaultMode` and keeping both arguments should do the trick. That will open using the default mode BUT honor the readonly flag in priority if it is `true`.

But that didn't work for me... `openWithDefaultMode` wouldn't call the `GlobalConfiguration.get` function, even though `overrideMode` was `undefined`...?!

I can give it another shot...?

sebjulliand:
I think I understand: it will check the configuration if there is no `overrideMode`.
https://github.com/codefori/vscode-ibmi/blob/ab4d38c4cc20e7139726850d17edd465dd48013c/src/instantiate.ts#L676

So I think the correct way to call the command from the search view would be:
```
this.command = {
  command: `code-for-ibmi.openWithDefaultMode`,
  title: `Open`,
  arguments: [this._path, this._readonly  ? 'browse' as DefaultOpenMode : undefined]
};
```

Forcing `browse` only when the view is in read-only mode will leave it up to the command to decide how to open the editor otherwise.

chrjorgensen:
@sebjulliand Seems to work now - for find results only!

Search results have the additional option of range, using type `OpenEditableOptions`, which is not supported by `openWithDefaultMode`... 😢 

This is a case for our TypeScript guru! 🙏

sebjulliand (approved):
It looks good to me, and works like a charm.
I pushed the required changes to use `openWithDefaultMode` from the Search results as well. I added support for a Range to be passed onto `openWithDefaultMode`. Nothing too fancy, but it seems to work well!
I'll let you have a final lokk @chrjorgensen and merge when you're OK with it 😃

PR #2151, opened 2024-07-03
Minor fixes and enhancements to debugger support

sebjulliand:
### Changes
- Do not perform Sanity Check (i.e. migration of certificate generated prior to v2.9) if current user doesn't have `*ALLOBJ` authority
- Added a `regenerate certificate` action in debug service's right click menu
- Moved `open configuration` action to debug service's right click menu
image
- Changed the logic of the method checking the debug service job status when it starts. It did not perform well on slow CPU and/or bad connection
- Debug service details cache is now cleared when disconnecting instead of connecting. That clearing could happen after the `IBM i Debugger` view title was loaded when connecting, leading to the wring version being displayed when switching between connections.
- IBM i debugger view loads slightly faster thanks to `Promise.all`
- Added an `Open output` action on the error notification showed when the debug service fails to start. Clicking on it will load the job's QPRINT spooled file that contains the output from the service startup script.
image
image
- Force usage of `/QOpenSys/usr/bin/openssl` to generate certificates

### How to test this PR
- Connect to an IBM i with debug PTF installed
- Try to regenerate the service certificate
- Try to stop and start the service; the success notification should show up only once
- Restart the debug server and service to ensure the view still loads fine

### Checklist
* [x] have tested my change

### DIFF
--- a/package.json
+++ b/package.json
@@ -939,6 +939,13 @@
 				"icon": "$(file-binary)"
 			},
 			{
+				"command": "code-for-ibmi.debug.redo.setup",
+				"title": "Regenerate Service Certificate",
+				"category": "IBM i Debug",
+				"enablement": "code-for-ibmi:connected && !code-for-ibmi:debugManaged && !code-for-ibmi:debugWorking",
+				"icon": "$(file-binary)"
+			},
+			{
 				"command": "code-for-ibmi.debug.setup.local",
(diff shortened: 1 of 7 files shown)


chrjorgensen (approved):
@sebjulliand LGTM - works as expected. 👍

PR #2154, opened 2024-07-09
Await on event subscribers functions

sebjulliand:
### Changes
This PR fixes an issue where too many requests are made at the same time on the SSH connection when connecting to an IBM i, resulting in this error showing up:
image

This was due to the functions being triggered by the `Connected` event making too many requests in parallel.
Waiting for each function to end solves this problem.

### How to test this PR
- Connect to a system
- Make sure everything loads correctly (debug status, library list view, ...)
- The error message in the description must not show up

### Checklist
* [x] have tested my change


chrjorgensen (commented):
@sebjulliand LGTM - works as before, no issues...

However, I'm not able to recreate the error message, even in master branch, so I forward the review to @worksofliam for additional eyes and tests and final approval... 👀

sebjulliand:
@chrjorgensen yep, it's hard to reproduce; I can because i've got more extensions doing things during `connect`. The operation that takes the longer to process is the update of the debugger browser and its status (5 seconds over the VPN here). One of my extension does something that takes around 2 seconds. It gives plenty of time to create collisions!

worksofliam:
Like @chrjorgensen, I am also not able to recreate this error message. But, the implementation looks good. The extension also starts up as expected.

worksofliam (approved):

PR #2157, opened 2024-07-09
Always debug securely

sebjulliand:
### Changes
This PR removes the `debugIsSecure` parameter from the connection settings.
Since the Service Entry Points requires a client certificate, there is no point in keeping the option to turn off secure debugging.

If the client certificate is not found locally or if there is a mismatch, the `IBM i Debugger` view will offer to download and update it locally.

A notification is shown in the status bar when the client certificate is being downloaded.

### How to test this PR
1. Connect
2. Go to the `IBM i Debugger` view
3. Check if the client certificate needs to be downloaded and click on the action to download it
4. Try debugging

### Checklist
* [x] have tested my change

### DIFF
--- a/package.json
+++ b/package.json
@@ -369,11 +369,6 @@
 								"default": "8008",
 								"description": "Port to connect to IBM i Debug Service for SEP."
 							},
-							"debugIsSecure": {
-								"type": "boolean",
-								"default": true,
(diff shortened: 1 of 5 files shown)


chrjorgensen (approved):
@sebjulliand Works as previously, no issues found. 👍

PR #2175, opened 2024-07-17
Compile non-IFS sources from local

worksofliam:
### Changes

Enables local actions to compile sources from source members. This enables the ability to use `CRTDSPF`, `CRTPF`, `CRTLF`, etc, on local workspace sources.

### How to test this PR

Examples:

1. Create a valid display file
2. Create an action in `actions.json` for `CRTDSPF`
3. Use the Action runner to use the newly action
4. Put an intential error in the display file
5. Run the action again and expect problems to appear

Also, run the test cases for Actions.

### Nice to have

* New option in the Actions setup quickpick ✅
* Update the documentation on Actions
* Ability to define record length and CCSID of temp source file

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/errors/diagnostics.ts
+++ b/src/api/errors/diagnostics.ts
@@ -169,6 +169,18 @@ export function handleEvfeventLines(lines: string[], instance: Instance, evfeven
             }
             continue;
           }
+
+          // If we get there, that means that even though we compiled from local, we likely had to use a temp member.
+          // We should try to find the file in the workspace. Since we can use findFile (it's async), then we look for open
+          // tabs like we do below.
+          if (evfeventInfo.extension) {
+            const baseName = file.split(`/`).pop();
+            const openFile = Tools.findExistingDocumentByName(`${baseName}.${evfeventInfo.extension}`);
+            if (openFile) {
+              ileDiagnostics.set(openFile, diagnostics);
+              continue;
+            }
+          }
         }
       }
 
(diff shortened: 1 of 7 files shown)


src/api/errors/diagnostics.ts
```diff
@@ -169,6 +169,18 @@ export function handleEvfeventLines(lines: string[], instance: Instance, evfeven
             }
             continue;
           }
+
+          // If we get there, that means that even though we compiled from local, we likely had to use a temp member.
+          // We should try to find the file in the workspace. Since we can use findFile (it's async), then we look for open
+          // tabs like we do below.
+          if (evfeventInfo.extension) {
+            const baseName = file.split(`/`).pop();
+            const openFile = Tools.findExistingDocumentByName(`${baseName}.${evfeventInfo.extension}`);
```
SanjulaGanepola, line 178: Instead of having to do any work here to attempt to find the `openFile`, can't we pass the `uri` down from the action stuff to this diagnostic handling by doing something like -> https://github.com/codefori/vscode-ibmi/compare/feature/local_temp_member...SanjulaGanepola:vscode-ibmi:feature/local_temp_member

Based on your comment, maybe this `findExistingDocumentByName` logic won't even be needed if it only gets to this point if its compile from local as your comments mention in which case `uri` should always be good to use?

worksofliam: @SanjulaGanepola The uri makes sense in only some instances. But, if the error list has more than one file, then always using the uri doesn't make sense - which is why I opted for this method.

SanjulaGanepola: Good point, forgot that the error list can point to multiple files

SanjulaGanepola (approved):

PR #2182, opened 2024-07-19
Fix event functions chain crash

sebjulliand:
### Changes
Resolves https://github.com/codefori/vscode-ibmi/issues/2181
This PR prevents a failing event function from crashing the whole chain (i.e. preventing other functions from being run when connecting/disconnecting).

### How to test this PR
1. Connect
2. Make sure every view loads correctly

Connecting to PUB400 should show a warning about the failing function.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/Instance.ts
+++ b/src/api/Instance.ts
@@ -86,6 +86,7 @@ export default class Instance {
    */
   onEvent(event: IBMiEvent, func: Function): void {
     this.getSubscribers(event).set(`deprecated - ${func.name || "unknown"}_${this.deprecationCount++}`, { func });
+    console.warn("[Code for IBM i] Deprecation warning: you are using Instance::onEvent which is deprecated and will be removed in v3.0.0. Please use Instance::subscribe instead.");
   }
 
   fire(event: IBMiEvent) {
--- a/src/api/Instance.ts
+++ b/src/api/Instance.ts
@@ -94,6 +94,7 @@ export default class Instance {
 
   async processEvent(event: IBMiEvent) {
     const eventSubscribers = this.getSubscribers(event)
+    console.time(event);
     for (const [identity, callable] of eventSubscribers.entries()) {
       try {
         console.time(identity);
@@ -101,7 +102,7 @@ export default class Instance {
         console.timeEnd(identity);
       }
       catch (error) {
-        console.log(`Event function ${identity} failed`, error);
+        console.log(`${event} event function ${identity} failed`, error);
       }
       finally {
         if (callable.transient) {
@@ -109,5 +110,6 @@ export default class Instance {
         }
       }
     }
+    console.timeEnd(event);
   }
 }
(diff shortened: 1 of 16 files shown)


src/api/Instance.ts
```diff
@@ -66,7 +66,13 @@ export default class Instance {
 
   async processEvent(event: IBMiEvent) {
     for (const subscriber of this.subscribers.filter(s => s.event === event)) {
-      await subscriber.func();
+      try{
+        await subscriber.func();
+      }
+      catch(error){
+        const name = subscriber.func.name ? `Event function ${subscriber.func.name}` : `Anonymous event function`;
+        vscode.window.showWarningMessage(`${name} failed: ${error}`)
```
worksofliam, line 74: I don't know if the user cares. I think we should `console.error` here and let the using-extension handle the error in their function (e.g. show any warnings/issues to the user.)

worksofliam (changes requested):

src/api/Instance.ts
sebjulliand: Fair enough!

sebjulliand:
@worksofliam there you go.
I added a `subscribe` function that supersedes `onEvent` and replaced every occurrence of it.
I also added some time tracing around these function calls...it won't hurt, in case of a bottleneck situation.
And I also removed a few redundant view refreshes that were called twice.

worksofliam:
@sebjulliand Initial review of the code looks good. Glad to see `onEvent` marked as depreciated and replaced with `subscribe`. I will have a go at testing it tomorrow. I believe after this has merged, and with the release of 3.0.0, we will need to update the [documentation on our API](https://github.com/codefori/vscode-ibmi/pull/2182/files). After merging, and release of the next version, I can go ahead and try out `subscribe` in the Db2 for i extension.

src/api/Instance.ts
```diff
@@ -56,17 +59,57 @@ export default class Instance {
...
+    }
+    return eventSubscribers;
+  }
+
+  /**
+   * @deprecated Will be removed in `v3.0.0`; use {@link subscribe} instead
+   */
   onEvent(event: IBMiEvent, func: Function): void {
-    this.subscribers.push({ event, func });
+    this.getSubscribers(event).set(`deprecated - ${func.name || "unknown"}_${this.deprecationCount++}`, { func });
```
worksofliam, line 88: I think an additional `console.log` in here so extension developers can see `onEvent` is going away in 3.0.0.

Something like `you are using onEvent, which is deprecated and will be removed in 3.0.0. Please use `subscribe`

worksofliam (approved):
I think one last change and we should be good to go here.

src/api/Instance.ts
sebjulliand: Good idea!

sebjulliand:
Tgabjs for the review @worksofliam . Give it a last whirl and I guess we'll be good to go!

worksofliam:
@sebjulliand I can confirm this PR fixed a startup bug with the database extension as the event error handler is working as expected.

image

worksofliam (approved):

worksofliam:
@sebjulliand Merge when you are ready! Thanks for your work on this.

PR #2191, opened 2024-07-23
Fix and test case for supporting members with dollar signs

### LINKED ISSUE #2177: Opening files with $ sign in name
We can't open files/members any more when name of it has a $ sign in it for example ren$sign.cblle. This is not the case when I went back to 2.9.0. 
After that I tried every version and did not work any more. What did change? And I feel like depending on versions readonly and position attributes don't work the way expected.
Help appreciated.
Thank you.

worksofliam:
### Changes

Fixes #2177 by not escaping at the right time.

* A new test case with members with dollar signs in the name has been added
* Renamed `sanitise` to `noEscape` to make more sense based on the existing code logic

### How to test this PR

Examples:

1. Run the test cases
2. Try creating and editing a member with a dollar sign

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -149,7 +149,7 @@ export default class IBMiContent {
     member = this.ibmi.upperCaseName(member);
 
     let retry = false;
-    let path = Tools.qualifyPath(library, sourceFile, member, asp);
+    let path = Tools.qualifyPath(library, sourceFile, member, asp, true);
     const tempRmt = this.getTempRemote(path);
     while (true) {
       let copyResult: CommandResult;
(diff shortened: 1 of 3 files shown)


worksofliam:
It seems like there might be some contention between where the encoding actually happens. I am passing `true` to `qualifyPath` so it doesn't escape:

https://github.com/codefori/vscode-ibmi/blob/fix/dollars_in_names/src/api/IBMiContent.ts#L152

As I am expecting it to be escaped in the `runCommand` function: https://github.com/codefori/vscode-ibmi/blob/fix/dollars_in_names/src/api/CompileTools.ts#L681

src/api/IBMiContent.ts
```diff
@@ -261,7 +261,6 @@ export default class IBMiContent {
               //The member may be located on SYSBAS
               if (asp) {
                 path = Tools.qualifyPath(library, sourceFile, member);
-                retry = true;
```
sebjulliand, line 264: Why have you removed `retry = true` here?

sebjulliand (changes requested):
Code looks good to me. Replacing `sanitize` with `noEscape` is great...makes a lot more sense!
See my remark about that line being removed though.

worksofliam:
@sebjulliand Not sure why I did that. Re-requested your review.

sebjulliand (approved):
Approved and good to go!

PR #2213, opened 2024-08-03
Fix `Search Source File` crashing when used with a complex filter

sebjulliand:
### Changes
Fixes https://github.com/codefori/vscode-ibmi/issues/2207

Searching for text in members of a source file would crash if run on a "complex" filter. By complex, I mean a filter whose member filtering term is either a Regex or not a single generic name (`XXX*`).

This PR fixes this issue, making sure the search runs fine in all cases.
It also fixes some minor issues:
- The search progress message was still `Listing members...` even after the listing was done and `grep` was running
- The position used when opening a search hit could be wrong if the line started with one or more whitespaces
- The interval set for displaying different messages during the search was not cleared immediately when the search was over

### How to test this PR
1. Run a search on a simple filter
2. Run a search on a complex filter

In both cases, the results must be as expected and only the members actually listed by the filter must be shown in the search result.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -86,6 +86,7 @@ export default class IBMi {
 
   commandsExecuted = 0;
 
+  //Maximum admited length for command's argument - any command whose arguments are longer than this won't be executed by the shell
   maximumArgsLength = 0;
 
   dangerousVariants = false;
@@ -913,8 +914,8 @@ export default class IBMi {
           }
 
           if ((!quickConnect || !cachedServerSettings?.maximumArgsLength)) {
-            //Compute the maximum length of arguments. Source: Googling and https://www.in-ulm.de/~mascheck/various/argmax/#effectively_usable
-            this.maximumArgsLength = Number((await this.sendCommand({ command: "expr `getconf ARG_MAX` - `env|wc -c` - `env|wc -l` \\* 4 - 2048" })).stdout);
+            //Compute the maximum admited length of a command's arguments. Source: Googling and https://www.in-ulm.de/~mascheck/various/argmax/#effectively_usable
+            this.maximumArgsLength = Number((await this.sendCommand({ command: "/QOpenSys/usr/bin/expr `/QOpenSys/usr/bin/getconf ARG_MAX` - `env|wc -c` - `env|wc -l` \\* 4 - 2048" })).stdout);
           }
           else{
             this.maximumArgsLength = cachedServerSettings.maximumArgsLength;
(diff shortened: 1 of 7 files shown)


worksofliam:
I will review over the next day or two. Is it worth creating or updating any test cases for this fix?

src/api/IBMi.ts
```diff
@@ -911,6 +912,14 @@ export default class IBMi {
             }
           }
 
+          if ((!quickConnect || !cachedServerSettings?.maximumArgsLength)) {
+            //Compute the maximum length of arguments. Source: Googling and https://www.in-ulm.de/~mascheck/various/argmax/#effectively_usable
+            this.maximumArgsLength = Number((await this.sendCommand({ command: "expr `getconf ARG_MAX` - `env|wc -c` - `env|wc -l` \\* 4 - 2048" })).stdout);
```
worksofliam, line 917: Some thoughts here:

1. Is this assuming bash is the default?
2. Should we use explicit paths to `expr` and `getconf`?

src/api/IBMi.ts
```diff
@@ -85,7 +84,9 @@ export default class IBMi {
   content = new IBMiContent(this);
   shell?: string;
 
-  commandsExecuted: number = 0;
+  commandsExecuted = 0;
+
+  maximumArgsLength = 0;
```
worksofliam, line 89: Some commands on the purpose of this new property might be good.

worksofliam (changes requested):
Left some comments. I personally need more context to understand the purpose of this additional property (sorry!)

src/api/IBMi.ts
sebjulliand: 1. No, it seems to work from QSH as well.
2. Both are found under `/QOpenSys/usr/bin` and I assume they are always here. But making their path absolute won't hurt 😊

src/api/IBMi.ts
sebjulliand: Yep. Basically, my biggest concern with the change I made to the search API was that there was a risk to issue a command that would be too long for the shell to process. So I looked if an actual limit existed and how to find it. And there it was.
So this property is the maximum admitted length for a command's arguments.

I use it in the search API to prevent an error in case the `grep` built to search is too long to be run. I put a comment in the search API but I'll add one here too.

sebjulliand:
> I will review over the next day or two. Is it worth creating or updating any test cases for this fix?

Here, have a few tests 😊

worksofliam (approved):

PR #2215, opened 2024-08-05
Fixed object/members listing failing when name contains variant chars

sebjulliand:
### Changes
Listing objects and/or members with names containing variant characters using a profile with CCSID different from 37 would result in members being displayed with American variant characters and physical source files not being displayed or being displayed as regular PF.

Example, with CCSID 297 variants characters:

Before this PR, when listing members
image

After this PR
image

Before this PR, when listing source physical files:
With Object type = *ALL
image
With Object type = *SRCPF
image

After this PR
image
image

Turns out the system tables contains the names in CCSID 37 and the OBJECT_STATISTICS table function does translates the names to the connection's CCSID before returning the results.

### How to test this PR
1. Connect with a profile with CCSID other than 37
2. Create a member with one or more variant characters
3. It must be displayed correctly and can be opened.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -656,7 +656,7 @@ export default class IBMiContent {
         asp,
         library,
         file: String(result.SOURCE_FILE),
-        name: String(result.NAME),
+        name: this.ibmi.sysNameInLocal(String(result.NAME)),
         extension: String(result.TYPE),
         recordLength: Number(result.RECORD_LENGTH) - 12,
         text: `${result.TEXT || ``}${sourceFile === `*ALL` ? ` (${result.SOURCE_FILE})` : ``}`.trim(),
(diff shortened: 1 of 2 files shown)


worksofliam (approved):
Approved with manual testing and running the existing test cases.

I am not pressed for this PR, but is it at-all possible to have a test case that runs when the user CCSID is not 37 to automate this testing>?

sebjulliand:
> Approved with manual testing and running the existing test cases.
Thanks @worksofliam !

> I am not pressed for this PR, but is it at-all possible to have a test case that runs when the user CCSID is not 37 to automate this testing>?
There you go, I added a test that will run if CCSID is not 37. It can run on PUB400 as well, the library test part is simply skipped in this case.
Good thing you had me add this test since I eventually found a bug that remained (filtering on source files name would not work if the filter had variant characters).

Give it a last try and let me know what you think!

PR #2219, opened 2024-08-19
Source Dates: compute a unique alias for each member + mode

sebjulliand:
### Changes
Fixes https://github.com/ARCAD-Software/elias-vscode/issues/35
How to reproduce: [Dates_changed.pdf](https://github.com/user-attachments/files/16662022/Dates_changed.pdf)

The source dates handling relies on an alias that was based on the member's library, file and name. When the same member is opened in both `edit` and `browse` mode, the same alias is used for both editor. When the editor opened in browse mode is closed before the one in edit mode, the bug described in https://github.com/ARCAD-Software/elias-vscode/issues/35 happens.

The bug can happen with any extension that uses the `member://` File System. It virtually cannot happen from Code for i UI since everywhere a member is opened, it is through the `code-for-ibmi.openWithDefaultMode` command.

### How to test this PR
1. Make sure Member Date Handling is turned on
2. Open a member in both browse and edit mode.
    - Using another extension (like ARCAD Elias)
    - Using a command you'll code just for the sake of testing this   
3. Close the browse mode editor
4. Make a change in the edit mode editor
5. Save and close the editor
6. Re-open the member
7. Dates must be correctly updated

### Checklist
* [x] have tested my change


julesyan (approved):
BTW I was able to replicate this issue with the Code for i UI by opening a file in Browse mode then Edit mode while browse mode is still loading. 
Seems to have also fixed the date not updating at all sometimes or the dates all 0ing out. 
Looks good!

sebjulliand:
Awesome, thanks a lot @julesyan !

PR #2231, opened 2024-08-27
Change logic for triggering disconnect

worksofliam:
### Changes

We were not triggering the disconnect when the connection timed out. This was causing the database extension issues because the jobs tied to the connection were not going away because the extension was not signaled of the disconnect.

Real changes:

* Trigger `disconnect` and cleanup client when timeout happens
* Trigger `connected` no matter if connecting for the first time or if reconnecting.

### How to test this PR

Examples:

#### Example 1

1. Use release version of the extension
2. Connect to server, spin up SQL job with database extension
3. Trigger the timeout somehow (disconnect from VPN in my case)
4. Use the reconnect option in the UI
5. See that the SQL jobs haven't ended, but are instead just handing

#### Example 2

1. Use this branch of the extension
2. Connect to server, spin up SQL job with database extension
3. Trigger the timeout somehow (disconnect from VPN in my case)
4. Use the reconnect option in the UI
5. See the previous jobs are gone, and new ones must be started - no hang ups

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -159,20 +159,24 @@ export default class IBMi {
 
           let tempLibrarySet = false;
 
-          const disconnected = async () => {
-            const choice = await vscode.window.showWarningMessage(`Connection lost`, {
-              modal: true,
-              detail: `Connection to ${this.currentConnectionName} has dropped. Would you like to reconnect?`
-            }, `Yes`);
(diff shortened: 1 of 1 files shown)


sebjulliand (changes requested):
Works as expected for me; it cleans up everything thanks to the `disconnect` event being triggered.
Fix that small typo and merge away!

PR #2232, opened 2024-08-28
Extend the base API to allow other extensions to add their own components.

worksofliam:
### Changes

Adds brand new export to the base API which allows other extensions to add their own installable components.

Components are checked on connection to ensure they are installed on the OS.

### How to test this PR

Examples:

1. Run the test cases
2. Ensure extension starts up with existing components as usual.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/components/component.ts
+++ b/src/components/component.ts
@@ -1,7 +1,4 @@
 import IBMi from "../api/IBMi";
-import { CopyToImport } from "./copyToImport";
-import { GetMemberInfo } from "./getMemberInfo";
-import { GetNewLibl } from "./getNewLibl";
 
 export enum ComponentState {
   NotChecked = `NotChecked`,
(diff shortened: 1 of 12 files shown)


sebjulliand:
@worksofliam I tried to make the component's workflow as clear as possible for whoever would like to register a component.
Let me know what you think.

Question: would we want to keep track of which extension registers which component and be able to display a list of all the registered components (in the settings view, for example)?

sebjulliand:
> Question: would we want to keep track of which extension registers which component and be able to display a list of all the registered components (in the settings view, for example)?

Forget it: yes we do.
image

@worksofliam give it a whirl 😊

src/components/component.ts
```diff
@@ -1,61 +1,93 @@
 import IBMi from "../api/IBMi";
-import { CopyToImport } from "./copyToImport";
-import { GetMemberInfo } from "./getMemberInfo";
-import { GetNewLibl } from "./getNewLibl";
 
-export enum ComponentState {
-  NotChecked = `NotChecked`,
-  NotInstalled = `NotInstalled`,
+export const enum ComponentState {
```
worksofliam, line 3: Should we convert this to a `type` so it's easier to export?

worksofliam:
@sebjulliand Okay, I think we need one final test each and then we're good!

worksofliam:
Working good for me - let's merge away!

PR #2248, opened 2024-09-03
Support workspace folder parameter on `runAction`

SanjulaGanepola:
### Changes

This PR simply adds the `workspaceFolder` as an optional parameter on `runAction`. The purpose of this change is so that when `runAction` is used on an object or member in Project Explorer, the local project's `LIBL` and `CURLIB` variables are used.

Associated change in PE: https://github.com/IBM/vscode-ibmi-projectexplorer/pull/575

### How to test this PR

1. Actions unrelated to PE should work as before.
2. Actions from PE should use the project's `LIBL` and `CURLIB`.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/CompileTools.ts
+++ b/src/api/CompileTools.ts
@@ -79,15 +79,17 @@ export namespace CompileTools {
     }
   }
 
-  export async function runAction(instance: Instance, uri: vscode.Uri, customAction?: Action, method?: DeploymentMethod, browserItem?: BrowserItem): Promise<boolean> {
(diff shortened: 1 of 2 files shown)


worksofliam:
Code is fine, but I think this is the last parameter we accept before changing the entire signature to use an object instead.

SanjulaGanepola:
@worksofliam @sebjulliand Could we get this PR merged and released? Need it for a PE release 😃

worksofliam (approved):

PR #2250, opened 2024-09-06
Implement remove filter command in object browser

worksofliam:
This pull request adds a new command, "Remove Filter", to the object browser in the IBM i category. The command allows users to remove a filter from the object browser. The implementation includes changes to the command registration and the initialization of the object browser.

This change was added to seperate the difference between deleting an object vs removing the filter (just like we do in the IFS Browser)

### DIFF
--- a/src/views/objectBrowser.ts
+++ b/src/views/objectBrowser.ts
@@ -1234,6 +1234,15 @@ export function initializeObjectBrowser(context: vscode.ExtensionContext) {
         }
       } while (newLibrary && !newLibraryOK)
     }),
+    vscode.commands.registerCommand(`code-for-ibmi.objectBrowser.removeFilter`, async (node?: ObjectBrowserItem) => {
+      if (node && node instanceof ObjectBrowserFilterItem) {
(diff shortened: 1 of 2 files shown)


src/views/objectBrowser.ts
```diff
@@ -1234,6 +1234,15 @@ export function initializeObjectBrowser(context: vscode.ExtensionContext) {
         }
       } while (newLibrary && !newLibraryOK)
     }),
+    vscode.commands.registerCommand(`code-for-ibmi.objectBrowser.removeFilter`, async (node?: ObjectBrowserItem) => {
+      if (node && node instanceof ObjectBrowserFilterItem) {
+        const message = t('objectBrowser.delete.confirm', node.toString());
+        if (await vscode.window.showWarningMessage(message, { modal: true }, t(`Yes`))) {
```
julesyan, line 1240: The 'Remove Shortcut' in the IFS browser does not have a warning, would we want to match that functionality more?

worksofliam:
Swapping back to draft and will reopen based on suggestion from 2252

PR #2252, opened 2024-09-10
Allow filters to be deleted even if protected

sebjulliand:
### Changes
Fixes https://github.com/codefori/vscode-ibmi/issues/2251
The logic for deleting Object Browser items would prevent filters from being deleted if they were protected.

This PR makes filter item deletable, not mater what.

### How to test this PR
1. Try to delete a protected filter
2. The confirmation dialog must show up
3. If confirmed, the filter must be removed

### Checklist
* [x] have tested my change

### DIFF
--- a/src/views/objectBrowser.ts
+++ b/src/views/objectBrowser.ts
@@ -1246,7 +1246,7 @@ export function initializeObjectBrowser(context: vscode.ExtensionContext) {
         candidates.push(...objectTreeViewer.selection.filter(i => i instanceof ObjectBrowserItem) as ObjectBrowserItem[]);
       }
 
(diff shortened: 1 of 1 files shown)


worksofliam (approved):
Works as expected. Maybe there is some overlap between this and #2250?

sebjulliand:
> Works as expected. Maybe there is some overlap between this and #2250?

A little bit...I guess the command in #2250 should simply be:
```ts
vscode.commands.registerCommand(`code-for-ibmi.objectBrowser.removeFilter`, (node:  ObjectBrowserItem, nodes:  ObjectBrowserItem[]) => vscode.commands.executeCommand(`code-for-ibmi.objectBrowser.delete`, node, nodes));
```

PR #2277, opened 2024-10-02
Simplify IBMi class 

krethan:
### Changes

Introducing new classes to help simplify logic in IBMiSettings class.

### How to test this PR

Examples:

1. Run the test cases

### Checklist

* [* ] have tested my change
* [* ] have added myself to the contributors' list in [CONTRIBUTING.md](https://github.com/codefori/vscode-ibmi/blob/master/CONTRIBUTING.md)

### DIFF
--- a/src/api/IBMiApps.ts
+++ b/src/api/IBMiApps.ts
@@ -1,0 +1,87 @@
+import IBMi from "./IBMi";
+import { RemoteApp, RemoteApps, RemoteFeatures } from "../typings";
+
+export default class IBMiApps {
+
+    private remoteApps: RemoteApps;
+    private remoteFeatures: RemoteFeatures;
+
+    constructor() {
+        this.remoteApps = [
+            {
+                path: `/usr/bin/`,
+                names: [`setccsid`, `iconv`, `attr`, `tar`, `ls`]
+            },
+            {
+                path: `/QOpenSys/pkgs/bin/`,
+                names: [`git`, `grep`, `tn5250`, `md5sum`, `bash`, `chsh`, `stat`, `sort`, `tar`, `ls`, `find`]
+            },
+            {
+                path: `/QSYS.LIB/`,
+                // In the future, we may use a generic specific.
+                // Right now we only need one program
+                // specific: `*.PGM`,
+                specific: `QZDFMDB2.PGM`,
+                names: [`QZDFMDB2.PGM`]
+            },
+            {
+                path: `/QIBM/ProdData/IBMiDebugService/bin/`,
+                specific: `startDebugService.sh`,
+                names: [`startDebugService.sh`]
+            }
+        ];
+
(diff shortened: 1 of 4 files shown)


src/api/IBMiSettings.ts
```diff
@@ -0,0 +1,513 @@
...
+
+        return Promise.resolve(clearMsg);
+
+    }
+
+    async clearTempDirectory(tempDir: string): Promise<string> {
+
+        let clearMsg = '';
+
+        this.connection.sendCommand({
```
worksofliam, line 240: You are better off awaiting here and using `try`/`catch` because it's going to return before it finishes below.

src/api/IBMiApps.ts
```diff
@@ -0,0 +1,87 @@
...
+            this.remoteFeatures[name] = undefined;
+        }
+
+    }
+
+    getRemoteApps(): RemoteApps {
+        return this.remoteApps;
+    }
+
+    setRemoteFeatures() {
```
worksofliam, line 55: I think this needs an improved name since it's setting all the apps to undefined.

src/api/IBMiSettings.ts
```diff
@@ -0,0 +1,513 @@
...
+            .catch(e => {
+                // CPF2125: No objects deleted.
+                clearMsg = `Temporary data not cleared from ${tempDir}.`;
+            });
+
+        return Promise.resolve(clearMsg);
+
+    }
+
+    async checkObjectExists(library: string, object: string, type: string): Promise<boolean> {
```
worksofliam, line 255: This API doesn't belong here. I think we have an 'object exists' API already. (See `checkObject` under `IBMiContent`)

src/api/IBMiSettings.ts
```diff
@@ -0,0 +1,513 @@
...
+        const objResult = await this.connection.runCommand({
+            command: `CHKOBJ OBJ(${library}/${object}) OBJTYPE(${type})`,
+            noLibList: true
+        });
+
+        return Promise.resolve(objResult.code === 0);
+
+    }
+
+    async deleteObject(library: string, object: string, type: string): Promise<boolean> {
```
worksofliam, line 266: It might be worth moving this to `IBMiContent` for re-usability.

worksofliam (changes requested):
Left some general comments. Also, we typically use an indent of 2 instead of 4. Looks good so far!

krethan:
Okay sending for another review.

worksofliam:
@krethan I should give a warning that there is another PR being worked on which separates the UI logic from the backend logic in the IBM i class. That may have to come before this.

krethan:
Thanks for the heads up

worksofliam:
@krethan That happened and the frontend logic is no longer tied into the backend. Sorry for the trouble!

PR #2284, opened 2024-10-11
Migrate to `vscode.l10n` API

sebjulliand:
### Changes
Migrated code from our custom translation tooling to VS Code standard API.
Used a script to migrate most of the code and manually fixed remaining calls to` l10n.t` that couldn't be changed automatically.

### How to test this PR
1. Change VS Code display language to one of the supported locale
2. Look around for translated strings (debug view, IFS browser, ...)

### Checklist
* [x] have tested my change


src/webviews/settings/index.ts
```diff
@@ -355,19 +354,19 @@ export class SettingsUI {
...
-              .addFile(`privateKeyPath`, `${t(`privateKey`)}${stored.privateKeyPath ? ` (${t(`current`)}: ${stored.privateKeyPath})` : ``}`, t(`login.privateKey.label`) + ' ' + t(`login.privateKey.support`))
+              .addInput(`host`, vscode.l10n.t(`Host or IP Address`), undefined, { default: stored.host, minlength: 1 })
+              .addInput(`port`, vscode.l10n.t(`Port (SSH)`), undefined, { default: String(stored.port), minlength: 1, maxlength: 5, regexTest: `^\\d+$` })
+              .addInput(`username`, vscode.l10n.t(`Username`), undefined, { default: stored.username, minlength: 1 })
+              .addParagraph(vscode.l10n.t(`Only provide either the password or a private key - not both.`))
+              .addPassword(`password`, `${vscode.l10n.t(`Password`)}${storedPassword ? ` (${vscode.l10n.t(`Password`)})` : ``}`, vscode.l10n.t(`Password`))
+              .addFile(`privateKeyPath`, `${vscode.l10n.t(`Private Key`)}${stored.privateKeyPath ? ` (${vscode.l10n.t(`Private Key`)}: ${stored.privateKeyPath})` : ``}`, vscode.l10n.t(`Private Key`) + ' ' + vscode.l10n.t(`Private Key`))
```
worksofliam, line 362: See the diff here. It's missing the `support` part that used to be there, which had clearing wording.

worksofliam (changes requested):
Only one small thing to change and then we're good to merge. Nice work!

src/webviews/settings/index.ts
sebjulliand: Well spoted; my update script wasn't that good. It must be OK for both the password and private key field now.

PR #2293, opened 2024-10-16
Fix QSysFS iasp handling

### LINKED ISSUE #2289: Issue with opening members with new version 2.13.4
I often use the Go to File commands and open from the search view.  I noticed that when I open using that method the tool does not want to open.   

image
image

I have to revert to the prior version to open members not using the standard filter item.

sebjulliand:
### Changes
Fixes #2289

Opening member located on an iasp from the member search result or from the `Go to` action failed because the URI would not contain the member's asp.

This PR fixes this issue by looking for the member on each known iasp if if fails to find it on `*SYSBAS`.
If the member is found on an iasp, the information is cached so any other member opened for the same library will automatically be done using the iasp previously found.

If a member is opened from the object browser, the library's iasp is also cached immediately to be reused if needed.

The library => iasp cache is cleared after disconnecting.

### How to test this PR
1. Search for members in a library located on an iasp
2. Open a member from the results
3. Use the `Go to` action to open a member located on an iasp

### Checklist
* [x] have tested my change

### DIFF
--- a/src/filesystems/qsys/QSysFs.ts
+++ b/src/filesystems/qsys/QSysFs.ts
@@ -2,9 +2,10 @@ import { parse as parsePath } from "path";
 import { ParsedUrlQueryInput, parse, stringify } from "querystring";
 import vscode, { FilePermission, FileSystemError } from "vscode";
 import { onCodeForIBMiConfigurationChange } from "../../api/Configuration";
+import IBMi from "../../api/IBMi";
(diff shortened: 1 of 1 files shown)


chrjorgensen (approved):
@sebjulliand Works perfectly - as usual...! 🥇 

Approved - merge when appropriate.

sebjulliand:
Thanks @chrjorgensen, much obliged 🥰

PR #2301, opened 2024-10-19
Feature/download members with correct EOL format based on OS

preetpalsinghmehra:
### Changes

This PR addresses an issue with line endings in the code repository. Currently, when the member is downloaded from the object browser to the local repository on a Windows machine, files are downloaded with LF (Line Feed) endings instead of the expected CRLF (Carriage Return Line Feed) format. As a result, when code is pushed to the remote repository, Git detects changes across all lines due to inconsistent line endings, leading to unnecessary diff outputs and complicating code reviews. As a fix CPYTOSTMF is changed based on the type of Operating System.

### How to test this PR
Download the member from object browser on Windows machine it should be downloaded with CRLF (Carriage Return Line Feed) and on Linux/Mac OS it should be downloaded with LF (Line Feed). It can be verified by checking the format on the bottom right of the VS Code.

image

Examples:

1. Run the test cases

### Checklist

* [x] have tested my change

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,3 +59,4 @@ Thanks so much to everyone [who has contributed](https://github.com/codefori/vsc
 * [@NicolasSchindler](https://github.com/NicolasSchindler)
 * [@marcin-ogon](https://github.com/marcin-ogon)
 * [@Detrytus59](https://github.com/Detrytus59)
+* [@preetpalsinghmehra](https://github.com/preetpalsinghmehra)
--- a/src/views/objectBrowser.ts
+++ b/src/views/objectBrowser.ts
@@ -888,10 +888,18 @@ export function initializeObjectBrowser(context: vscode.ExtensionContext) {
           try {
             await connection.withTempDirectory(async directory => {
               task.report({ message: t('objectBrowser.downloadMemberContent.download.cpytostmf'), increment: 33 })
-              const copyToStreamFiles = toBeDownloaded
+              let copyToStreamFiles;
+              if(os.platform() === 'win32') {       //Windows --> CRLF
+                copyToStreamFiles = toBeDownloaded
+                .filter(member => member.copy)
+                .map(member => `@CPYTOSTMF FROMMBR('${member.path}') TOSTMF('${directory}/${member.name.toLocaleLowerCase()}') STMFOPT(*REPLACE) STMFCCSID(1208) DBFCCSID(${config.sourceFileCCSID}) ENDLINFMT(*CRLF);`)
+                .join("\n");
+              } else {                              //Other OS --> LF
+                copyToStreamFiles = toBeDownloaded
                 .filter(member => member.copy)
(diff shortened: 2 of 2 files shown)


chrjorgensen:
@preetpalsinghmehra Thank you for your contribution to this project - very much appreciated. 👍 

However, it seems like your change is made to fix some line ending issues you have with git on Windows - and that you're now changing Code for IBM i in order to remove the issue with git.

Git has always recognized that it could be used on different platforms with different line endings in files and have provided a mechanism to avoid this issue. It involves telling git how to handle line endings by changing some settings in git (globally or in the repository) and have git handling the line endings when reading from and writing to the repository.

I found this page to have the best description of how this works and how to set it up:
https://adaptivepatchwork.com/2012/03/01/mind-the-end-of-your-line

Especially this paragraph is important:
> Git’s primary solution to all this is to specify that **LF** is the best way to store line endings for text files in a Git repository’s object database. It doesn’t force this on you but most developers using Git and GitHub have adopted this as a convention and even our own [help](http://help.github.com/line-endings/) recommends setting up your config to do this.

Other useful links for git and line endings could be these:
https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes
https://git-scm.com/book/en/v2/Getting-Started-First-Time-Git-Setup
https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

Code for IBM i is built and used on both Windows, Mac and Linux, and we prefer to avoid platform-specific code in the code base, since this would make it harder to do testing, which would require access to multiple platforms.

Also note that you can easily have VS Code change the line endings in the source files, if you click in the status bar on the line ending symbol (or press F1 and search for `Change End of Line Sequence`). And that you can [configure VS Code](https://blog.boot.dev/clean-code/line-breaks-vs-code-lf-vs-crlf/) to use the LF line ending for new files globally or per project (open the settings (Ctrl-,) and search for `files.eol`).

preetpalsinghmehra:
Thanks @chrjorgensen  for you valuable comments and sharing the information.

chrjorgensen:
@preetpalsinghmehra I just checked my git configuration on Windows, and I have `core.autocrlf = input` - meaning, git will change any CRLF to LF during commit and keep the LF during checkout, clone etc. I work solely with LF as line endings, and should I by mistake commit a file using CRLF, it will be converted to LF by git.

Hope you fix your issues. Line endings and git is tricky, and no matter how you configure git, you always risk to get in to this situation, if the original committer or repository creator did not think about this.

Should we close this PR - or do you have other reasons for this change of member download behavior?

jkyeung:
Not that I am in a position to make the decision, but in my opinion, it can be closed, for all the reasons @chrjorgensen stated.

It has been getting easier and easier over the years to do things "the Unix way" on Windows, and if you use VS Code, you can easily configure it so that you work entirely with LF-only line endings, even for files destined to spend their entire existence on Windows. For example, the Python interpreter works equally well on source with LF or CRLF line endings, regardless of platform, so it's common to simply use LF for all Python source, no matter where it lives.

preetpalsinghmehra:
@chrjorgensen Yes we can close the pull request thanks.

PR #2325, opened 2024-11-01
Ensure type is a string for translation

worksofliam:
SImple fix. TS compiler knew it was a string but our strict type generation wanted to be safe.

We don't need a release for this.


chrjorgensen (approved):
@worksofliam Not tested, just approved...

PR #2326, opened 2024-11-03
Add Auto Reconnect setting

### LINKED ISSUE #2300: Auto-connect to server configuration
If I open VS Code with an RPG file left open (or open a local RPG file) I don't always want to connect to an IBM i.
I think auto-connect could be a really handy feature, though I think it needs a bit of configuration.

A global setting for auto-connect would be helpful (yes/no/ask me).
"Yes" would auto-connect to the last connection
"No" would never connect
"Ask Me"pop up a dialog box and ask me which system I want to connect to and give me the option to not connect to any of them.

sebjulliand:
### Changes
Resolves #2300 

Based on @p-behr suggestion, this PR adds a new global settings that sets the action taken out by Code for i when a member/streamfile is left open when VS Code is closed and re-opened.

image

Supported values are:
* `ask`: opens a notification asking whether or not the user wants to reconnect
image
* `always`: always reconnect
* `never`: never reconnect

If the user choose to reconnect, the member/streamfile is reloaded. Otherwise, the editor that was left open is closed.

### How to test this PR
1. Open a member/streamfile
2. Leave it open and close VS Code
3. Re-open VS Code
4. Based on the reconnect setting, code for i will reconnect, close the editor or ask the user if they want to reconnect

### Checklist
* [x] have tested my change

### DIFF
--- a/src/filesystems/qsys/FSUtils.ts
+++ b/src/filesystems/qsys/FSUtils.ts
@@ -10,7 +10,7 @@ import { Tools } from "../../api/Tools";
  * @returns `true` if the user choses to reconnect, `false` otherwise.
  */
 export async function reconnectFS(uri: vscode.Uri) {
-  const reconnect = GlobalConfiguration.get<ReconnectMode>("autoReconnect");
+  const reconnect = GlobalConfiguration.get<ReconnectMode>("autoReconnect") || "ask";
   let doReconnect = false;
   switch (reconnect) {
     case "always":
--- a/src/filesystems/qsys/FSUtils.ts
+++ b/src/filesystems/qsys/FSUtils.ts
@@ -1,11 +1,13 @@
-import vscode, { FileSystemError, l10n } from "vscode";
+import vscode, { l10n } from "vscode";
 import { GlobalConfiguration, ReconnectMode } from "../../api/Configuration";
+import { Tools } from "../../api/Tools";
 
 /**
  * Called when a member/streamfile is left open when VS Code is closed and re-opened to reconnect (or not) to the previous IBM i, based on the `autoReconnect` global configuration value.
+ * If the user choses not to reconnect, the editor tab will be closed.
  * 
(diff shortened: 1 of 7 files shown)


src/filesystems/qsys/FSUtils.ts
```diff
@@ -0,0 +1,39 @@
...
+export async function reconnectFS(uri: vscode.Uri) {
+  const reconnect = GlobalConfiguration.get<ReconnectMode>("autoReconnect") || "ask";
+  let doReconnect = false;
+  switch (reconnect) {
+    case "always":
+      doReconnect = true;
+      break;
+
+    case "ask":
+      if (await vscode.window.showInformationMessage(l10n.t("Do you want to reconnect and open {0}?", uri.path.split('/').reverse()?.[0]), l10n.t("Reconnect"))) {
```
worksofliam, line 21: Perhaps `path.basename(uri.path)` might be more suited?

worksofliam (changes requested):
I very much like this PR, but I think the message should show what system they're reconnecting to.

> Do you want to reconnect to SYSTEM and open FILE?

src/filesystems/qsys/FSUtils.ts
sebjulliand: Yes, that's way less clunky 😅

sebjulliand:
There you go @worksofliam. It's all yours for a final review.
image

src/filesystems/qsys/FSUtils.ts
sebjulliand: Yes, that's way less clunky 😅

worksofliam (approved):
Working good! I will let you merge :)

p-behr:
This is terrific, thanks @sebjulliand

sebjulliand:
> This is terrific, thanks @sebjulliand

My pleasure 🙏🏻

PR #2339, opened 2024-11-09
Refactor ASP handling to use current iASP name and fetch additional information

worksofliam:
Update the handling of ASPs to utilize the current iASP name instead of the deprecated sourceASP property, and enhance the fetching of ASP information by RDB name.

This means that Code for IBM i will always use the ASP that is defined as part of their user profile / job description.

The downside about this PR is I actually have no systems to test it on. So I kind of buggered up there.

### Checklist

* [ ] We need test cases for the new ASP methods in the `IBMi` class
* [ ] Ensure that when an ASP is set that the user can browse and open members in *SYSBAS when they have an ASP configured
   * [ ] Object browser checked
   * [ ] Members checked
* [ ] have tested my change


worksofliam:
Ping @sebjulliand - please take a look at this in the next couple of weeks if you can.

src/api/IBMi.ts
```diff
@@ -1279,6 +1302,63 @@ export default class IBMi {
...
+    let asp: AspInfo|undefined;
+    if (typeof by === 'string') {
+      asp = this.iAspInfo.find(asp => asp.name === by);
+    } else {
+      asp = this.iAspInfo.find(asp => asp.id === by);
+    }
+
+    if (asp) {
+      return asp.name;
+    }
```
sebjulliand, line 1332: This looks a lot like `getIAspDetail`'s implementation. Reuse it then!
```suggestion
    return getIAspDetail(by)?.name;
```

src/filesystems/qsys/QSysFs.ts
```diff
@@ -131,40 +129,23 @@ export class QSysFS implements vscode.FileSystemProvider {
     async getMemberAttributes(connection: IBMi, path: QsysPath & { member?: string }) {
         const loadAttributes = async () => await connection.content.getAttributes(path, "CREATE_TIME", "MODIFY_TIME", "DATA_SIZE");
```
sebjulliand, line 130: Use not deprecated `getContent()`.
```suggestion
        const loadAttributes = async () => await connection.getContent().getAttributes(path, "CREATE_TIME", "MODIFY_TIME", "DATA_SIZE");
```

src/api/IBMi.ts
worksofliam: Smart!

worksofliam:
@sebjulliand The merge here too complex to make it worth it. I will need to do it. This is a worthy PR I think.

PR #2345, opened 2024-11-12
Fix invalid options for `Upload...` command

SanjulaGanepola:
### Changes

Fix invalid options for `Upload...` command


sebjulliand (approved):
Good catch, thanks @SanjulaGanepola !

PR #2353, opened 2024-11-18
Fixed incorrect messages + open members download location

sebjulliand:
### Changes
Some messages were placed incorrectly, because of mistakes made after the `vscode.l10n` migration.
This PR fixes this.

It also fixes the `Open` action on the notification shown after downloading a physical source file. The action would run on Windows but not on *nix OSes, because of the file system's case sensitivity.

### How to test this PR
**Only on Linux/Mac os**
1. Download a physical source file
2. Click on the `Open` action in the `Download complete` notification
3. The download location must open

### Checklist
* [x] have tested my change


sebjulliand:
Gracias!

PR #2369, opened 2024-11-26
Testing: basic coverage support

worksofliam:
Adds ability to capture how many times functions or instance methods were called in an object or class, allowing us to get coverage information for our test suite.

The 'Test Cases' UI now has two groups:

* **Test Suites** which contains the nodes that was what the original Test Cases UI had, and
* **Coverage** which shows a list of registered classes or objects inside of it, and when they are expanded, shows the method or function name with a count of how many times it has been called. Additionally, each class or object shows the percentage of methods/functions that have been covered inside of it.

image


worksofliam:
New build now includes the ability to see where methods are used!

PR #2374, opened 2024-11-27
Refactor connection methods and improve encapsulation

worksofliam:
Make several fields private, deprecate old methods, and replace them with new ones to enhance code clarity and maintainability. Reorganize command handling for better structure.

The biggest changes here:

* `instance.getConfig` and `instance.getContent` deprecated in favor of `IBMi#getConfig` and `IBMi#getContent`
* `instantiate.ts` has been refactored to have almost all command definitions moved into their own folders to declutter.
* Changed `DebugConfiguration` to make use of a constructor instance of using `instance.getConnection`
* `Instance` now has a connect method which handles the callback timeout 

### Todo:

* [x] Move all connection logic from `/Users/barry/Repos/vscode-ibmi/src/webviews/login/index.ts` to separate UI and backend logic.
* [x] Stop using `instance.setConnection` in `IBMi#connect`.
* [x] Instead of `login.ts` creating a new instance of `IBMi`, create call `instance.connect(...)` which manages the new connection

--- 

I am not looking for a code review. Reviewers, please just focus on ensuring the UI/UX experience is not different.


worksofliam:
All but one test case (the tab one as always) is passing. Seems like UX is not hindered by this change. I did additional user testing on:

* Cancelling during the connect process ✅
* Timeout during the connect process ✅

image

SanjulaGanepola (changes requested):
> I am not looking for a code review. Reviewers, please just focus on ensuring the UI/UX experience is not different.

Did a quick run through and found some issues:

1. Getting `Not connected to an IBM i` error after connecting:
image

2. Cannot open source members in Object Browser.

3. Cannot open stream files in IFS Browser.

4. (I haven't confirmed if this bug exists in the current release outside this PR) - Click a stream file in the IFS Browser. After clicking `Change Working Directory` command, I would expect it to fail but it doesn't. Since it doesn't fail, it updates the working directory and then breaks expanding any IFS directory.
image

worksofliam:
@SanjulaGanepola Items 1, 2 and 3 have been fixed. Please have another go while I run the test cases again.

worksofliam:
@SanjulaGanepola I looked at number 4 but I am not able to recreate. Do you have more detailed steps I could try?

image

SanjulaGanepola:
1/2/3 - Good now

4 - The issue is reproducible if you are able to get it to set the working directory to either a path that does not exists or if you set it to a path to a file. To get it to be a path that does not exist, use the `IBM i: Change Working Directory` from the command palette and enter some wrong path. It sets it even though it fails (first screen cap below). To get it to be a path to a file, click on a stream file in the IFS Browser and click `Change Working Directory` on the `view/title` of the IFS Browser which gives a similar error (second screen cap below). In both cases, then using the IFS Browser to expand folders and do other stuff is broken because the working directory is invalid.
image
image

5 - Open a source member. Breadcrumb navigation when clicking on a library or object seems to be broken. Seems to work fine in the IFS.

6 - NL string issue when using `IBM i: Search Source File` from the command palette:
image

worksofliam:
@SanjulaGanepola checked 5 and 6. Same issues on master, so I wouldn't consider them as part of this PR.

worksofliam:
Regarding 4, it's the same issue on master - and I think this has been an issue for sometime.

image

SanjulaGanepola:
Your right, all of them are indeed on master 😅. In that case, I didn't find any UI issues with this PR.

worksofliam:
@SanjulaGanepola 

A newer, smaller PR will be opened. This is a mess.

PR #2380, opened 2024-11-27
Fixed Component manager exportation

sebjulliand:
### Changes
The Component Manager could not be used by other extensions because of the way types were exported.
This PR fixes this, allowing other extension to actually contribute components to Code for i.

It also adds an `.npmrc` file to avoid issues when `npm` is ran with a proxy npm repo defined locally.

### How to test this PR
1. Run the test cases
2. Open connection settings
3. Check all Components are listed correctly

### Checklist
* [x] have tested my change


src/components/cqsh/index.ts
```diff
@@ -1,33 +1,32 @@
 
 import { stat } from "fs/promises";
-import { ComponentState, IBMiComponent } from "../component";
 import path from "path";
 import { extensions } from "vscode";
+import IBMi from "../../api/IBMi";
+import { ComponentState, IBMiComponent } from "../component";
+
+export class CustomQSh implements IBMiComponent {
+  static ID = "CQSh";
```
worksofliam, line 9: Please reset this `id` back to `cqsh`

worksofliam (changes requested):
Please see my only comment, and other than that, it's all good.

src/components/cqsh/index.ts
sebjulliand: Sure!

sebjulliand:
Thanks @worksofliam !

worksofliam:
Please see my alternative PR: #2381

worksofliam (approved):
Looks like components are still loading good and the test cases are passing. Merge away!

PR #2381, opened 2024-11-27
Refactor IBMiComponent for easier export

worksofliam:
Refactor methods in the IBMiComponent class to improve export functionality. This includes changing access modifiers and return types for better clarity and consistency.

It also enables `IBMiComponent` to be implemented, and not only extended. This is particularly important for other extensions, so they can add their own components.

```ts
  class a implements IBMiComponent {
    state: ComponentState = `NotChecked`;
    cachedInstallDirectory: string | undefined;
    constructor(readonly connection: IBMi) {}
    async getInstallDirectory(): Promise<string> {
      return `BOOP`;
    }
    getState(): ComponentState {
      return `NotInstalled`;
    }
    async check(): Promise<ComponentState> {
      return `NotInstalled`;
    }
    toString(): string {
      throw new Error("Method not implemented.");
    }
    getIdentification(): ComponentIdentification {
      return {name: `test`, version: 1};
    }
    getRemoteState(): ComponentState | Promise<ComponentState> {
      throw new Error("Method not implemented.");
    }
    async update(): Promise<ComponentState> {
      return `Installed`;
    }
  }
  
  extensionComponentRegistry.registerComponent(context, a);
```

### DIFF
--- a/src/components/component.ts
+++ b/src/components/component.ts
@@ -34,10 +34,10 @@ export type IBMiComponentType<T extends IBMiComponent> = new (c: IBMi) => T;
  */
 export abstract class IBMiComponent {
   public static readonly InstallDirectory = `$HOME/.vscode/`;
-  private state: ComponentState = `NotChecked`;
-  private cachedInstallDirectory: string | undefined;
+  state: ComponentState = `NotChecked`;
+  cachedInstallDirectory: string | undefined;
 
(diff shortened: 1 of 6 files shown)


worksofliam:
This screenshot shows that with this PR, we can implement the `IBMiComponent` to add components.

image

PR #2382, opened 2024-11-27
Deprecate legacy APIs and update references

worksofliam:
Begin the process of deprecating legacy APIs/getters by updating references throughout the codebase to use new connection methods. This change enhances code maintainability and prepares for future improvements.

* Creates new generic `connect` and `disconnect` methods in `instance`, which is safer than calling `new IBMi()` to set the instance in the `connect` method.
* Split up `instantiate` so it doesn't hard code all commands inside of it.
* Move timeout handler and reconnect logic out of IBMi class.
* Remove all uses of `instance` in the IBMi class.

### DIFF
--- a/src/api/CompileTools.ts
+++ b/src/api/CompileTools.ts
@@ -59,42 +59,40 @@ export namespace CompileTools {
     return inputValue;
   }
 
-  function applyDefaultVariables(instance: Instance, variables: Variable) {
-    const connection = instance.getConnection();
-    const config = instance.getConfig();
-    if (connection && config) {
-      variables[`&BUILDLIB`] = variables[`CURLIB`] || config.currentLibrary;
-      if (!variables[`&CURLIB`]) variables[`&CURLIB`] = config.currentLibrary;
-      if (!variables[`\\*CURLIB`]) variables[`\\*CURLIB`] = config.currentLibrary;
-      variables[`&USERNAME`] = connection.currentUser;
-      variables[`{usrprf}`] = connection.currentUser;
-      variables[`&HOST`] = connection.currentHost;
-      variables[`{host}`] = connection.currentHost;
-      variables[`&HOME`] = config.homeDirectory;
-      variables[`&WORKDIR`] = config.homeDirectory;
+  function applyDefaultVariables(connection: IBMi, variables: Variable) {
+    const config = connection.getConfig();
+    variables[`&BUILDLIB`] = variables[`CURLIB`] || config.currentLibrary;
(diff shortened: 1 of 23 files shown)


SanjulaGanepola (changes requested):
Changes look good. Only one minor issue found so far where the output channel is not disposed properly in the case the connection was dropped. This can lead to multiple output channels.

image

Steps to reproduce:
1. Connect
2. Once connected, disconnect VPN or use some other means to drop the connection.
3. Click `Yes` to reconnect.
4. Click `Cancel` to try again.
5. Notice the `Code for IBM i` output channel still exists

worksofliam:
@SanjulaGanepola Cleaned it up and tested it. Let me know if it continues.

SanjulaGanepola:
Just gave it a try and still getting the same issue. I see in the latest commit you added a call to [dispose](https://github.com/codefori/vscode-ibmi/pull/2382/commits/2f87b30fbde49b742644f287edd83eafa23f4165#diff-af91aaa8436f73c4457121b35da47e46d53459b47f044b7dbdb78ea785cb1824R1098). But I think the issue is that the second time we are dealing with a new `IBMi` object so we aren't disposing the output channel of the original connection.

worksofliam:
@SanjulaGanepola I had to rewrite some of the logic for this for it to work better and also improve how we handle timeouts and disconnects. Give it a try now.

worksofliam:
@sebjulliand can I get a review on this whenever you have a moment? 🙏

SanjulaGanepola:
@worksofliam Noticing some issues now with view enablement. Drop the connection and then hit `No, get logs` or `Cancel`. The standard 3 views are still shown instead of the `Servers` view even though there is no connection.

worksofliam:
@SanjulaGanepola More changes, this time a small rework on the `connected` and `disconnected` events get fired. An improvement I think, as we have now removed all calls to `instance` in the `IBMi` class :)

SanjulaGanepola (changes requested):
All but one scenario is good 😅

Connect -> Drop connection (dc VPN, etc.) ->Click `Yes` to reconnect -> Press `Cancel` on `Connecting: Connecting via SSH.`

Even though we cancelled, it only shows the Server view after a couple seconds (it is still connecting in the background it seems). Also shouldn't then be getting the `Timed out while waiting for handshake` message after that either.

worksofliam:
@SanjulaGanepola I actually believe this to be the current scenario on master too. I will double check this week.

SanjulaGanepola (approved):
Just gave it a try on `master` and your right. Everything else looks good, so its approved!

SanjulaGanepola (approved):

PR #2342, opened 2024-11-12
Updated `vscode-elements` and fixed Private Key storage

sebjulliand:
### Changes
This PR bumps `vscode-elements` to `1.8.0` (and updates the package name from `@bendera/vscode-webview-elements` to `@vscode-elements/elements`.

The CustomUI code has been updated according to the changed made since version `0.12.0`.

It also fixes an issue that prevented connections private key path from being saved (after saving, the path would be `undefined`).
image

#### Test the private key setting
1. Edit a connection
2. Set a private key
3. Save
4. Re-open the connection for edition: the private path must be stored correctly and shown in the settings

#### Check the UI
- Open the `Actions` management screen
- Open the connection settings
- Open the custom variables editor

### Checklist
* [x] have tested my change

PR #2386, opened 2024-11-28
Implement custom file picker functionality in CustomUI

worksofliam:
Introduce the native file picker feature that allows users to select files through a button in the CustomUI. This change enhances user interaction by enabling file selection and updating the corresponding field with the selected file's path.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,8 +9,8 @@
 			"version": "2.14.4-dev.0",
 			"license": "MIT",
 			"dependencies": {
-				"@bendera/vscode-webview-elements": "^0.12.0",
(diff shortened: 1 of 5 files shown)


sebjulliand (changes requested):
Very nice change; works like a charm!

I'd have one request: do you think you can bump the UI framework version and replace the `vsc-select` event by `vsc-tree-select`, like I did in #2342 ?

If so, we can kill two bird with one stone and delete #2342.

worksofliam:
@sebjulliand I have done:

1. Updated the package
2. Copied over your additional changes
3. Tested with the updated package and it's working as expected still.

sebjulliand (approved):
Perfect, very good change. Thanks mate!

PR #2390, opened 2024-11-29
Fix getAttributes for files with spaces and enhance tests

worksofliam:
Address issues with retrieving attributes for files that contain spaces in their names. Add additional test logic to ensure proper handling of such files and directories.


sebjulliand (approved):
Good one, thanks for the test!

PR #2393, opened 2024-12-02
Make private key path OS indendant

sebjulliand:
### Scenario
I use Code for i on several systems and I use VS Code settings synchronization so I my connection settings are shared across all my systems.

Since I may not be using the same OS and/or user name on these systems, the private key path set in the settings is different on all the systems, making it impossible to share that setting (the last path used will erase all the others because of VS Code synch).

### Changes
This PR changes the connection private key path before it is stored to make it OS agnostic:
- If the path starts with the user's hoe directory path, this part of the path is replaced by `~`
- The path is always stored with `/` as the path segment separator

When the private key path is displayed in the settings or retrieved to connect to an LPAR, it is transformed to a local path:
- `~` is replaced with the current user home directory
- `/` are converted back to `\` when running on Windows

This is how it looks:
- In the settings:
image

- Displayed/used on Linux
image

- Displayed/used on Windows
image

This was tested on both Windows and Linux thanks to the settings synchronization. Each system used a different ed25519 private key. I created the connection from one system and I could connect from the other after the settings got synched.

A warning is displayed in the settings if the private key doesn't exist on the system.
image

### How to test this PR
1. Set a private key to connect to a system
2. Check the user settings to see how it's stored
3. Connect to the system

### Checklist
* [x] have tested my change

### DIFF
--- a/src/webviews/settings/index.ts
+++ b/src/webviews/settings/index.ts
@@ -396,7 +396,6 @@ export class SettingsUI {
                       // then remove the password from the data and
                       // use the keypath instead
                       data.privateKeyPath = Tools.normalizePath(data.privateKeyPath);
-                      console.log(data.privateKeyPath)
                       await ConnectionManager.deleteStoredPassword(context, name);
                       vscode.window.showInformationMessage(vscode.l10n.t(`Private key updated and will be used for "{0}".`, name));
                     }


src/webviews/settings/index.ts
```diff
@@ -393,6 +395,8 @@ export class SettingsUI {
                       // If no password was entered, but a keypath exists
                       // then remove the password from the data and
                       // use the keypath instead
+                      data.privateKeyPath = Tools.normalizePath(data.privateKeyPath);
+                      console.log(data.privateKeyPath)
```
worksofliam, line 399: Let's remove this log.

worksofliam (changes requested):
Seems like just one log that needs to be removed and then it's good to go.

worksofliam (approved):
Very good @sebjulliand! Merge away :)

PR #2396, opened 2024-12-03
Fixed GETMBRINFO status fetch crashing on system with undefind Runtime CCSID

sebjulliand:
### Changes
`LONG_COMMENT` field in `qsys2.sysroutines` is VARGRAPHIC so the query used to fetch GETMBRINFO version crashes on systems with undefined Runtime CCSID since converting from CCSID 1200 to 65535 is not possible.

This PR casts `LONG_COMMENT` to `VARCHAR` to prevent the crash from happening.

### How to test this PR
1. Connect
2. Open the connection settings
3. Go to the Components tab
4. GetMemberInfo must show `(version 1): Installed` (before the PR, with Runtime CCSID 65535, it would show `(version 0): Error`

### Checklist
* [x] have tested my change

### DIFF
--- a/src/components/getMemberInfo.ts
+++ b/src/components/getMemberInfo.ts
@@ -19,7 +19,7 @@ export class GetMemberInfo implements IBMiComponent {
   }
 
   async getRemoteState(connection: IBMi): Promise<ComponentState> {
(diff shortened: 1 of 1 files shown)


worksofliam (approved):
Simply lovely as Max would say. Thanks!

PR #2429, opened 2024-12-18
Use pfgrep instead of QShell grep when possible

NattyNarwhal:
[pfgrep][pfgrep] is a fast alternative to QShell grep. It aims to be much faster and support the full spectrum of PCRE regular expression functionality.

In my testing, searching for the string `Qp0l` in `QSYSINC/H` shows:

- qshell grep: ~11 seconds
- pfgrep v0.2: ~8 seconds
- pfgrep v0.3: ~2 seconds

Quite a big improvement!

Questions include:

- Should VS Code offer to point you to where to download pfgrep? It's not yet in the IBM repository, but we do have our own repo, RPMs available, and it can be built from source.
- With faster search, doing things like regular expressions or searching across multiple files/libraries goes from tedious to quite possible. That's out of scope for this PR, but would be quite nice to add. Same deal for regular expressions instead of fixed strings.
- If you do need any features in pfgrep, let me know. I can prioritize things that VS Code needs.

[pfgrep]: https://github.com/SeidenGroup/pfgrep

### Changes

Description of change here.

### How to test this PR

Examples:

1. Run the test cases

### Checklist

* [X] have tested my change
* [X] have created one or more test cases


worksofliam:
> Should VS Code offer to point you to where to download pfgrep? It's not yet in the IBM repository, but we do have our own repo, RPMs available, and it can be built from source.

This is the main question. I suspect the best we can do is offer a page in our docs (under Tips) specifically for getting `pfgrep` and what the benefits are.

NattyNarwhal:
Might want to wait for next version, I'm working on `-f` flag support (saving us from having to escape) and inverting behaviour of `-t` (making it unnecessary for the sake of the unit test).

worksofliam:
@NattyNarwhal I don't know if you're interested, but we have a major codebase change in the last week, so it might be worth fixing up this PR.

NattyNarwhal:
I'll push out an update to pfgrep this week and rebase.

worksofliam:
@NattyNarwhal I think your test cases need to be moved over to the newer `search.test.ts` file, which can be tested without launching VS Code.

https://github.com/codefori/vscode-ibmi/blob/master/src/api/tests/suites/search.test.ts

NattyNarwhal:
I did move the search over; should I revert the ones in the non-API tests suite side?

NattyNarwhal:
I've rebased this for the ASP handling changes.

worksofliam:
@NattyNarwhal Good news - it's time to actually get this merged. I need the PR updated with the latest from `master`, then I will test it, and try the test cases, and we're good to go. Thanks and sorry for the delay.

src/api/Search.ts
```diff
@@ -42,10 +44,23 @@ export namespace Search {
...
+        // pfgrep vs. qshell grep difference: uses -r for recursion instead of -R
+        // (GNU/BSD grep treat them the same); we don't use recursion yet though...
+        // older versions before 0.4 need -t to trim whitespace, 0.4 inverts the flag
+        const command = `${pfgrep} -inHr -F "${sanitizeSearchTerm(searchTerm)}" ${memberFilter}`;
+        result = await connection.sendCommand({
+          command: command,
+          directory: connection.sysNameInAmerican(`${asp ? `/${asp}` : ``}/QSYS.LIB/${library}.LIB/${sourceFile}.FILE`)
+        });
+      } else {
+        const command = `/usr/bin/grep -inHR -F "${sanitizeSearchTerm(searchTerm)}" ${memberFilter}`;
```
worksofliam, line 58: Something we need to do in the future is not use the hardcoded grep here, but instead use grep from `remoteFeatures`.

worksofliam: Though looking at it, we might have a problem using `remote` features with two different versions of grep (since they share a name and key). Ignore this for now.

worksofliam (approved):
Tests are passing. Thank you and sorry for the wait.

PR #2437, opened 2024-12-27
Correctly check GetNewLibl component state

sebjulliand:
### Changes
`GetNewLibl` component state was not checked as it should. Instead of being checked in the corresponding component, it was added to the `IBMi::remoteApps` global array to be checked as a remote feature. This caused two issues:
- The actual version of `GetNewLibl` was not checked
- A new entry for checking `GetNewLibl` was added to the global array with each 'non-quick connect' connection attempts
image

This PR fixes this by:
- Not checking `GetNewLibl` as a remote feature anymore
- Handling `GetNewLibl` version in its component object

### How to test this PR
1. Connect
2. Check that `GetNewLibl` procedure get created in the connection temporary library with the version in its comment

### Checklist
* [x] have tested my change


worksofliam (approved):
Merge away!

PR #2443, opened 2025-01-02
New connection settings `Connection Timeout` field

### LINKED ISSUE #2442: Add a connection setting to change SSH `ready timeout` value
### Discussed in https://github.com/orgs/codefori/discussions/2440

<sup>Originally posted by **a4ginatl** January  2, 2025</sup>
My development system has a communications overload every morning as I transfer logs from the production system to the dev system.

When the files are being transferred, I cannot connect to the IBM i as the connection is talking longer than the timeout allows for. As a result, I have to wait until about 10:00am before I am able to connect.

TIA, Darryl Freinkel.
</div>

## Required change
A new connection setting should be added to be able to change SSH2 `readyTimeout` configuration.

sebjulliand:
### Changes
Resolves #2442 and #2440.

This PR adds a new `readyTimeout` field to the connection settings (i.e. to the `ConnectionData` type). This field is described [here](https://github.com/mscdex/ssh2#client-methods).

A new `number` input has been added to the Connection/Login Settings to manage this field:
image

The `CustomUI` class has been enhanced to support the `number` input type as well as its `min` and `max` attributes.
The SSH port field has been turned into a `number` input.

### How to test this PR
1. Create a new connection with wrong credentials
2. Change the `Connection Timeout` field to 1000 (i.e. 1 second)
3. Try to connect: the connection attempt must fail after 1 second
4. Change the `Connection Timeout` field to 5000
5. Try to connect: the connection attempt must fail after 5 seconds

### Checklist
* [x] have tested my change

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
 			"license": "MIT",
 			"dependencies": {
 				"@ibm/ibmi-eventf-parser": "^1.0.2",
-				"@vscode-elements/elements": "^1.9.0",
+				"@vscode-elements/elements": "^1.9.1",
 				"adm-zip": "^0.5.14",
 				"crc-32": "https://cdn.sheetjs.com/crc-32-latest/crc-32-latest.tgz",
 				"csv": "^6.2.1",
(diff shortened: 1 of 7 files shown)


julesyan:
There is some odd behaviour that I have been able to inconsistently recreate. I set the value to 2000 (or 20,000), double checked the value. Connected to the system successfully. Disconnected, then checked the value again and it seems to have changed the timeout value to 1999 (or was 19,998). I will see if i can consistently recreate it, but so far its sporadic behaviour. 
image

julesyan:
I can set the readyTimeout field to 0, even though it has min set to 1. It shows as red, but the save is still allowed. I checked the settings file and it does save as 0. It does not treat the timeout as 0ms, it seems to just ignore the timeout when it is 0. 
image

In regards to the other issue I mentioned with the timeout value changing, I have been unable to recreate it more than 2 more times after 200+ tries. So it seems to be a possible one off of the string to number conversion. I wouldn't be concerned about it.

sebjulliand:
Thanks @julesyan !
I fixed the numeric value not being entirely checked - setting a value lesser than 1 will not allow you to save the settings.

As for the odd behavior, I think I found what the "issue" is. Spoiler alert: it's not a bug, it's a feature. 😅
It turns out that if a numeric field has the focus and your mouse pointer it on it, the value will increment/decrement if you use your mouse scrollwheel!

To try to prevent this from happening, I moved the field to the bottom of the form. Since it used to be in the middle, it was natural to use the wheel to scroll down to the `Save` button, with the risk of changing the value while doing so.

Give it a try and let me know how it goes!

sebjulliand:
I changed the type of the SSH port number to a "number" input, so there's a better check on the value.

julesyan (changes requested):
Everything looks good aside from if i have an invalid value for the timeout, then use the scroll or up/down arrows on the field, it does not revalidate that data, it stays as an error

sebjulliand:
@julesyan it should be OK now. I used a more global event to catch changes in any `input` field. Let me know how it goes!

julesyan (approved):
Looks good!

sebjulliand:
Much obliged @julesyan 🙏🏻

PR #2457, opened 2025-01-15
Fix filtering objects when filter contains both source files and other object types

### LINKED ISSUE #2269: Cannot add multiple object filters when *PFSRC is in the list
When you add object filters you can add a comma separated list such as *PFSRC,*MSGF,*BNDDIR according to the documentation, however, when you add *SRCPF it will not show unless it is the only entry ie:

*PFSRC shows as expected
*PFSRC,*MSGF will only show the *MSGF objects.

Tried moving *SRCPF to other locations in the list with no change.

janfh:
Fixes #2269 

Objectbrowser filter needs to check if object is a source file if filter contains both *SRCPF and other object types - not only the object type.


chrjorgensen (changes requested):
@janfh Thank you for your contribution! 😍 

Your change works - but also stops object type filter `*ALL` from working. Can you fix this?

PR #2472, opened 2025-01-19
fix: extendend content handling srcdat field for unsupported srcdat field values

crisrivlop:
### Changes

Add support for Null dates when extendedContent.ts/uploadMemberContentWithDates function is called.

### How to test this PR

Examples:

1. Open a remote file with SRCDAT field in null if extendedMemberSupport property is true. It should open with this new change.

### Checklist

* [x] have tested my change


sebjulliand:
I like the idea, but `try_cast` has been introduced in 7.5 and 7.4 TR6 - and for now, Code for IBM i still supports 7.3 TR 5 at a minimum.
This won't make it's way into the extension until the minimum OS version we supports becomes 7.4 TR6 (which would prevent quite a lot of people from using the extension, sadly).

worksofliam:
@sebjulliand you are fully right. I also worry about the performance implications of this additional function call. This process is already a little bit slow.

PR #2480, opened 2025-01-25
Replace hardcoded 'diff' command with constant

worksofliam:
Replace the hardcoded 'diff' command with a constant for improved maintainability and consistency in the code.


p-behr:
Confirm that compare works :-)

PR #2481, opened 2025-01-25
Enhance runCommand with error handling and event callbacks

worksofliam:
Patrick B raised an issue with me directly about this issue. 

>When I use the default commands I get the variable names in the prompt, but the command runs with the proper values.
>
> I have a couple of custom actions that aren't working properly...the variable names appear in the form, and are also used in the command:

This new change fixes the issue where the variables were appearing in both confirmation UIs before they were expanded.

Include unit tests for variable expansion in ILE commands.


p-behr:
Confirming that variable replacement in actions works :-)

PR #2483, opened 2025-01-25
Update ASP configuration and tests

worksofliam:
Replace the old ASP configuration with a new API and implement tests for basic ASP functionality. This change enhances the handling of ASPs throughout the application.

This means that Code for IBM i will always use the ASP that is defined as part of their user profile / job description.

The downside about this PR is I actually have no systems to test it on. So I kind of buggered up there. But, I have started on some test cases which will only run when an ASP is defined.

* [ ] Finish off test cases


sebjulliand (changes requested):
Good job @worksofliam ! That's a very nice change that should make our lives easier 😁
I requested a few changes (nitpicking...) and I made a few changes myself to simplify calls to upload/download members (since we can leverage that iasp lookup you introduced!).

src/api/IBMi.ts
```diff
@@ -1434,6 +1449,68 @@ export default class IBMi {
...
+
+  getIAspName(by: string|number): string|undefined {
+    return this.getIAspDetail(by)?.name;
+  }
+
+  getCurrentIAspName() {
+    return this.currentAsp;
+  }
+
+  private libraryAsps: { [library: string]: number } = {};
```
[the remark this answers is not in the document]
worksofliam: Done in new commit.

src/api/IBMiContent.ts
```diff
@@ -137,15 +137,15 @@ export default class IBMiContent {
   /**
    * Download the contents of a source member
    */
-  async downloadMemberContent(asp: string | undefined, library: string, sourceFile: string, member: string, localPath?: string) {
-    asp = asp || this.config.sourceASP;
+  async downloadMemberContent(library: string, sourceFile: string, member: string, localPath?: string) {
```
worksofliam, line 140: @sebjulliand I am a little concerned about this signature change. If we merge this, we potentially break 3rd party extensions (and I was hoping this PR would just be a slip in).

Perhaps, for now, we keep the `asp` parameter, but rename it to `_` and add some jsdoc to indicate that it is no longer used.

Better yet, create a new function without the use of `asp` and then mark the old `downloadMemberContent` as deprecated.

src/api/IBMiContent.ts
```diff
@@ -205,66 +198,51 @@ export default class IBMiContent {
   /**
    * Upload to a member
    */
-  async uploadMemberContent(asp: string | undefined, library: string, sourceFile: string, member: string, content: string | Uint8Array) {
```
worksofliam, line 208: Same note about `asp` here.

worksofliam (commented):
@sebjulliand Please see my comments about parameter changes.

src/api/IBMiContent.ts
sebjulliand: Agreed...I had the deprecated function in mind too. I'll do that!

sebjulliand:
@worksofliam I turned the API changes to non-breaking breaking, using overloading. The test runner help me avoid a mistake, thanks again for that!

I saw what's wrong with Save As, I'll fix it tomorrow.

sebjulliand:
@worksofliam you're on. Last fix for Save As is in.

sebjulliand:
> @sebjulliand i have never used overloading as you did here. I learned something.
> 
> I will say, I am not a fan of the 'or' variables, but naturally I think that's because I would have implemented it differently when deprecating.

Honestly, I pondered both options, as the code is as bit uglier with the `or` variables. But since one signature is deprecated and it's only temporary, I chose overloading for this case. Had it been a permanent change, I'd have done as you suggested.

worksofliam:
@sebjulliand Assuming the ASP tests are running on a system where you have an ASP, then this can be merged. Thanks for finishing up this PR for me.

sebjulliand:
@worksofliam looks good to me:
image

And I played around with members on iasp: no issues so far (open, save, save as, compilation errors, ...).
Let's merge!

sebjulliand (approved):

PR #2485, opened 2025-01-25
Generate binder source from module or service program

### LINKED ISSUE #2484: Generate binder source from module
What about a new feature to generate binder source from module similar to RTVBNDSRC?

janfh:
Fixes #2484 

### Changes

* New command and object browser option to generate binder source from module or service program
* New functions in IBMiContent for retrieving module exports using DSPMOD *OUTFILE and service program exports using the PROGRAM_EXPORT_IMPORT_INFO view
* Generates binder source in a new editor

### How to test this PR

Examples:

1. Locate a module and a service program in object browser
2. Right click and select "Generate binder source"
3. Verify binder source generated in editor

### Checklist

* [x] have tested my change


worksofliam:
Honestly I love the idea!

There is some overlap with the FS extension (look up vscode-ibmi-fs on our org), but since it hasn't been released yet and isn't anywhere near being released, I believe it could live here until then.

Thoughts @sebjulliand?

worksofliam:
@janfh What are the chances you can write new test cases for the new APIs you added? Perhaps against objects in QSYS2?

janfh:
> @janfh What are the chances you can write new test cases for the new APIs you added? Perhaps against objects in QSYS2?

Sure :) But I'm not sure if there is any *MODULE to to test "getModuleExport". Do you know of any?

worksofliam:
@janfh You're right, I don't know of many modules to test with in either QSYS, QSYS2, or SYSTOOLS. If you have to skip one API, that's ok. But if you're feeling adventurous, perhaps your test could create a module with a simple export.

janfh:
@worksofliam : I have added a test for each of the two new methods in IBMiContent

src/ui/views/objectBrowser.ts
```diff
@@ -541,6 +541,25 @@ export function initializeObjectBrowser(context: vscode.ExtensionContext) {
       objectTreeViewer.reveal(item, options);
     }),
 
+    vscode.commands.registerCommand(`code-for-ibmi.generateBinderSource`, async (node: ObjectBrowserObjectItem) => {
+      const contentApi: IBMiContent = getContent();
```
worksofliam, line 545: We can leave this code as is, but in the future, we wouldn't usually define and import `IBMiContent` here as it is inferred from `getContent()`.

src/ui/views/objectBrowser.ts
```diff
@@ -541,6 +541,25 @@ export function initializeObjectBrowser(context: vscode.ExtensionContext) {
...
+      let exports: ProgramExportImportInfo[] | ModuleExport[] = [];
+      if (node.object.type === '*MODULE') {
+        exports = (await contentApi.getModuleExports(node.object)).filter(exp => exp.symbol_type === 'PROCEDURE');
+      } else {
+        exports = (await contentApi.getProgramExportImportInfo(node.object)).filter(info => info.symbol_usage === '*PROCEXP');
+      }
+      const content = [
+        `/*  Binder source generated from ${node}  */`,
+        ``,
```
worksofliam, line 554: I think `node` needs to be stringified here. Perhaps you could show do `node.object.name` instead?

src/api/tests/suites/content.test.ts
```diff
@@ -642,4 +643,56 @@ describe('Content Tests', {concurrent: true}, () => {
...
+  it('getProgramExportImportInfo', async () => {
+    const content = connection.getContent();
+    const info: ProgramExportImportInfo[] = (await content.getProgramExportImportInfo({
+      library: `QSYS2`,
+      name: `QCMDEXC2`,
+      type: `*SRVPGM`,
+      text: ``
+    })).filter(info => info.symbol_usage === '*PROCEXP');
+
+    expect(info.length).toBeGreaterThanOrEqual(1);
```
worksofliam, line 694: This test is failing on one of my systems! What do you need from me to help you debug this?

> expected 0 to be greater than or equal to 1

src/api/tests/suites/content.test.ts
```diff
@@ -642,4 +643,56 @@ describe('Content Tests', {concurrent: true}, () => {
...
+    });
+  });
+
+  it('getProgramExportImportInfo', async () => {
+    const content = connection.getContent();
+    const info: ProgramExportImportInfo[] = (await content.getProgramExportImportInfo({
+      library: `QSYS2`,
+      name: `QCMDEXC2`,
+      type: `*SRVPGM`,
+      text: ``
```
worksofliam, line 691: I see that `text` is not optional and therefore you are passing blank. I don't think we should do this.

I believe there are three options:

1. on `getProgramExportImportInfo`, use `Partial` like so: `async getProgramExportImportInfo(object: Partial<IBMiObject>)` - though be warned, this has other implications.
2. make `text` optional in the `IBMiObject` interface and see where the code breaks (try running `npm run typings` after this change) - I think I like this one the most, though not sure of the impact.
3. change the parameter type to be specific to your requirements: `async getProgramExportImportInfo(object: {library: string, name: string, type: "*SRVPGM"|"*MODULE"})`

worksofliam (approved):
Overall a nice change, but sadly I have a test case not passing and also am making a recommendation to the parameter type on your new method.

worksofliam (changes requested):
Sorry - meant to press request changes.

src/api/tests/suites/content.test.ts
worksofliam: The SQL statement returning 0 rows.

image

worksofliam: To be safe, we might also be better off creating a new object as part of the test case.

src/ui/views/objectBrowser.ts
janfh: ObjectBrowserObjectItem already have a toString-method:
https://github.com/codefori/vscode-ibmi/blob/ffbcbd6fffbc27d4a5e9ad74d142cfae1ea6294d/src/ui/views/objectBrowser.ts#L397

Example:
image

Do you still want me to change it?

src/api/tests/suites/content.test.ts
janfh: Looks like many other methods in IBMiContent is using option 3. Maybe it's better to have lib, name and type as parameters instead of IBMiObject?

src/ui/views/objectBrowser.ts
worksofliam: I did not see that. Looks good then!

janfh:
> Sorry - meant to press request changes.

Thanks for the feedback :)

src/api/tests/suites/content.test.ts
janfh: Several other methods in IBMiContent uses library and name as (string) parameters, and I did something similar. What do you think? Do you want me to try option 2?

src/api/types.ts
```diff
@@ -182,4 +182,23 @@ export type SearchHitLine = {
...
+}
+
+export interface ModuleExport {
+  module_library: string, 
+  module_name: string, 
+  module_attr: string, 
+  symbol_name: string, 
+  symbol_type: string,
+  argument_optimization: string,
+}
```
sebjulliand, line 202: Prefer camel case for variable names, to keep it consistent with the others.
```suggestion
export interface ProgramExportImportInfo {
  programLibrary: string
  programName: string
  objectType: string
  symbolName: string
  symbolUsage: string
  argumentOptimization: string
  dataItemSize: number
}

export interface ModuleExport {
  moduleLibrary: string
  moduleName: string
  moduleAttr: string
  symbolName: string
  symbolType: string
  argumentOptimization: string
}
```

sebjulliand (changes requested):
Nice feature, thanks @janfh !
I just have a few style/simplification suggestions related to the code. Thanks!

sebjulliand:
> Honestly I love the idea!
> 
> There is some overlap with the FS extension (look up vscode-ibmi-fs on our org), but since it hasn't been released yet and isn't anywhere near being released, I believe it could live here until then.
> 
> Thoughts @sebjulliand?

We'll definitely have to keep in mind to move it to FS in due time, since it's really an action targeted at objects 😄

package.json
```diff
@@ -2932,6 +2943,11 @@
 					"command": "code-for-ibmi.debug.setup.local",
 					"when": "!code-for-ibmi:debugManaged && view == ibmiDebugBrowser && viewItem =~ /^certificateIssue_localissue$/",
 					"group": "inline"
+				},
+				{
+					"command": "code-for-ibmi.generateBinderSource",
+					"when": "view == objectBrowser && viewItem =~ /^object.(module|srvpgm).*/",
```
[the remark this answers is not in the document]
janfh: Are you sure? The suggested change seems to be invalid syntax? Or am I missing something else here?

janfh:
> Nice feature, thanks @janfh ! I just have a few style/simplification suggestions related to the code. Thanks!

Thanks for the feedback! And sorry for the slow response :) Hope you will have another look at this

sebjulliand (approved):
Lovely! Looks good to me!
No worries, it's been busy around here too as you could see!

package.json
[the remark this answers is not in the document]
sebjulliand: Oh, right, that would be `object\\.`; the slash has to be escaped.

sebjulliand:
@worksofliam it's fine by me. You can have a final look and merge!

PR #2499, opened 2025-01-31
Fix/breadcrumb browse

sebjulliand:
### Changes
This fixes the breadcrumbs navigation for members.

image

The call to `getMemberAttributes` made by `stat` would fail for physical files, preventing the navigation from working (clicking on it would do nothing).

This PR fixes this by preventing stat from getting attributes except for members.
It also fixes the library listing in the breadcrumb (clicking on the library would only show QSYS and then it would fail to list anything relevant). QSYS FS now uses a dedicated call to `OBJECT_STATISTICS`, using `*ALLSIMPLE` since all we need is a list of names.

### How to test this PR
1. Open a member
2. Click on a path element in the breadcrumb and try to
* [ ] List libraries
* [ ] List files
* [ ] List members 

### Checklist
* [x] have tested my change

### DIFF
--- a/src/filesystems/qsys/QSysFs.ts
+++ b/src/filesystems/qsys/QSysFs.ts
@@ -1,9 +1,9 @@
 import { parse as parsePath } from "path";
 import { parse, ParsedUrlQueryInput, stringify } from "querystring";
 import vscode, { FilePermission, FileSystemError } from "vscode";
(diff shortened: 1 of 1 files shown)


chrjorgensen (approved):
@sebjulliand Works like a charm - now... 👍

sebjulliand:
Much obliged @chrjorgensen 🙏🏻

PR #2501, opened 2025-02-03
Return default CCSID when no result is found in runSQL query

worksofliam:
Ensure the function returns the default CCSID when the runSQL query yields no results.

I have found a system (`common1.iinthecloud.com`) which returns no rows when trying to get the CCSID of the SSHD job. If that is the case, then we just return the system QCCSID.

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -1,16 +1,16 @@
 import { parse } from 'csv-parse/sync';
 import fs from 'fs';
+import * as node_ssh from "node-ssh";
 import path from 'path';
 import tmp from 'tmp';
 import util from 'util';
-import * as node_ssh from "node-ssh";
-import { GetMemberInfo } from './components/getMemberInfo';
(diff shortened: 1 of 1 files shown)


sebjulliand:
@worksofliam are you able to connect to this system and check why it fails to return any result?
I game some thought to it and looking at the code, the query is not flawless:
- it's hard-coded to look for port 22. In most cases, it's fine, but if you run this on PUB400 where ssh runs on port 2222, it fails.
- `substring(job_name, locate('/', job_name, 8)+1, 15)` doesn't work on PUB400 either because the job id is `205625/QUSER/P0ZSPWT`. `Locate` returns `205625/QUS` and the subquery fails subsequently.

I pushed a commit that should fix that query (using the actual SSH port and calling locate twice to get a fool-proof result). Let me know if that makes sense to you and if you can test it on that `common1.iinthecloud.com`, see if it changes anything.

worksofliam:
@sebjulliand Your improvements are spot on. Honestly I was thinking I might do the port change in another PR, but it makes a lot of sense here. It did solve the problems with the `common1` system. With your approval, we will merge.

image

sebjulliand (approved):

sebjulliand:
Perfect! Merge away buddy! 🤩

PR #2502, opened 2025-02-03
Refactor component management and add state caching

worksofliam:
Improve code readability and maintainability by refactoring library and directory checks into separate methods. Introduce caching for component states to enhance connection performance.

Adds the component stats to the settings cache and validation each one individually. This will allow us to update one component without updating them all again.

Before:

slow

After:

fast

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -247,8 +247,6 @@ export default class IBMi {
 
       this.appendOutput(`Code for IBM i, version ${currentExtensionVersion}\n\n`);
 
-      let tempLibrarySet = false;
-
       callbacks.progress({
         message: `Loading configuration.`
       });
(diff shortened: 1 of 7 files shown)


jwoehr:
- Uninstalled Code4i and install .vsix
- First connection it was about the same (as one would expect)
- Disconnected and tried again, faster, obviously skipping some checks now that things are cached
- Assuming that "Connect and Reload Server Settings" would void the cache and load everything again

src/api/components/manager.ts
```diff
@@ -51,11 +53,38 @@ export class ComponentManager {
     });
   }
 
-  public async startup() {
+  public shouldUpdate(lastInstalled: ComponentInstallState[] = []) {
```
SanjulaGanepola, line 56: Where does `shouldUpdate` actually get used? Looks like there is some duplicate code here with `startup`. Is this intended?

src/api/IBMi.ts
```diff
@@ -939,6 +896,7 @@ export default class IBMi {
         qccsid: this.qccsid,
         jobCcsid: this.userJobCcsid,
         remoteFeatures: this.remoteFeatures,
+        installedComponents: this.componentManager.getInstallState(),
```
SanjulaGanepola, line 899:
```suggestion
        installedComponents: componentStates,
```

src/api/components/manager.ts
worksofliam: You're totally right. Left in there from my iterations. I will remove in a new commit.

src/api/IBMi.ts
worksofliam: Done

worksofliam:
@SanjulaGanepola I followed through with your recommendations. Please give another review.

worksofliam:
@SanjulaGanepola This PR is now ready for review again. The tests are correctly passing after I added a missing await.

SanjulaGanepola (approved):
Looks good now and connect is faster. Another great change!

PR #2504, opened 2025-02-05
Add removed APIs from `IBMi` class

worksofliam:
Deprecate the upload and download methods in the IBMi class, recommending the use of alternative methods for file operations.

This was needed as 2.14.5 has these APIs, and we can't remove them until 3.0.0.

Without this PR, the Db2 for i extension will break in the next release.


SanjulaGanepola (approved):
LGTM

Issue #2500, opened 2025-02-03
When running an action that doesn't use the *EVENTF, an error message shows in the log regarding *EVENTF being missing from command string.


richardschoen:
**Describe the bug**
When running a Run action that doesn't use the *EVENTF, an error message shows in the log regarding *EVENTF being missing from command string.

This message is not useful and I find it distracting.  Perhaps eliminate the false error message or let me turn that off as a setting ?

Example:
```*EVENTF not found in command string. Not fetching errors for LIB/SRCMBR001.```

**To Reproduce**
1. Create an  action or run any CL command that doesn't use *EVENTF in it's CL command string,
2. Run the action.
3. See error: ```*EVENTF not found in command string. Not fetching errors for LIB/SRCMBR001.```

**Expected behavior**
I would expect that if the *EVENTF is not used in my command line, I should not have to see this extraneous non-useful message. It's distracting.  

**Perhaps eliminate the false error message or let me turn that off as a setting ?**

I know what it's like to over message users. I've been doing that my entire career until someone points out sometimes that it's sometimes TMI and distracting. 

Thanks

**Screenshots**
<img width="895" alt="Image" src="https://github.com/user-attachments/assets/7e2f9bc7-9eaf-4642-9ec2-d6ce7ee03c42" />

**Environment (please complete the following information):**
 - Extension version [e.g. 2.14.7]
 - IBM i OS version [e.g. 7.5]

wimjongman:
I suggest to only fire this when the command is a CRT* command.

heymchri:
@richardschoen we've just worked around this by adding "/* OPTION(*EVENTF) */" at the end of those command strings. Since it's just a comment, it doesn't affect the command.

richardschoen:
@heymchri So it ignores the OPTION(*EVENTF) because it's in a comment ?

heymchri:
Hi @richardschoen That's correct. E.g.:

![Image](https://github.com/user-attachments/assets/4b694fa9-38d5-4dd0-8bc6-5ff14f3b9f94)

richardschoen:
Very clever. Thanks for sharing that.

richardschoen:
That's good to know.  I honestly hadn't tested recently :-)

PR #2514, opened 2025-02-13
Log regarding *EVENTF being missing from command string. #2500

wimjongman:
### Changes
* This message is now suppressed from commands not starting with CRT
* Complexity of the method was reduced

### How to test this PR

*    Create an action or run any CL command that doesn't use *EVENTF in it's CL command string,
*    Run the action.
*    See error: *EVENTF not found in command string. Not fetching errors for LIB/SRCMBR001.

### DIFF
--- a/src/ui/actions.ts
+++ b/src/ui/actions.ts
@@ -34,525 +34,524 @@ export async function runAction(instance: Instance, uri: vscode.Uri, customActio
   const connection = instance.getConnection();
 
   const uriOptions = parseFSOptions(uri);
-  if (connection) {
-    const config = connection.getConfig();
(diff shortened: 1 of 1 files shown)


wimjongman:
Review might be a bit messy because I reduced the complexity with a couple of levels. Please switch off whitespace compare

https://github.com/codefori/vscode-ibmi/pull/2514/files?diff=split&w=1

worksofliam:
@wimjongman A nice and simple change, but not sure it needed 400 line changes. Perhaps this should be better split into two pull requests: one for doing the cleanup, and another for the actual change. Esp. because everything is done in one commit.

PR #2517, opened 2025-02-17
Recover source date changes when editors are left open

### LINKED ISSUE #2346: Source Dates replaced with todays date
### Discussed in https://github.com/orgs/codefori/discussions/2275

<sup>Originally posted by **NextPage66** September 27, 2024</sup>
I always like to see the source date in the gutter.  It is very helpful and useful.    
Most of the time I see the dates but there have been times when i am moving between tabs and notice that the date is not visible.  I usually close the member and reopen it to get the source dates.  
However, I made some changes to a member yesterday and after I pressed the save button I noticed the dates were not there.  I closed and reopened the source and all the lines now have yesterdays date.   I think there is not an issue as long as the source dates are visible.  But if they ever disappear I cannot save my changes until they decide to show up again.  If I click around and come back to the source file they will sometimes reappear.  This is a big deal.  Not user friendly to have to monitor that so closely.  
Now I have to spend a few hours getting the source dates back to their original date.
What would cause the source dates to disappear and reappear in the gutter?

Example from today:
I opened a compare on 2 source files.
Pic #1 shows the first time I opened them and it is missing the source dates on the left side.
Pic #2 shows the second time I opened them and it is not missing the source dates.
All I did was close the compare and reopen it.
sourcedate2
sourcedate1

This issue is not just with compare function because it has happened on just regular source viewing also.

Thanks for any help.

 </div>

sebjulliand:
### Changes
Fixes #2346.

Two use cases that can break source date changes tracking have been identified. They both involve leaving one or more source member tabs opened and reconnecting after:
- Closing and re-opening VS Code
- Updating and reloading extensions

In both cases, source tracking is broken because the cached information used to detect line changes are lost.

This PR attempts to fix this issue by detecting a potential loss of the source date tracking information and trying to reload them before saving. The `diff` method is always used to perform this recovery as `edit` does not allow to recover which lines where changed.

It also forces the source member support to be reloaded right away when VS Code reconnects after being closed with source members opened.

The warning about source date support having been changed has been removed this PR allows source dates for being recovered and handled in this case too.

### How to test this PR
- Turn on Source dates support in the connection settings.
- Set `Auto Reconnect` to `always` in user settings

#### Use case 1
- Connect
- Open at least two source members
- Close VS Code
- Re-open VS Code
- Wait for VS Code to reconnect
- Make changes in both members and save
- Source dates must be kept

#### Use case 2
- Connect
- Open at least two source members
- Open the command palette and execute the `Restart extension host` command
image
- Wait for VS Code to reconnect
- Make changes in both members and save
- Source dates must be kept

#### Use case 3
- Connect
- Turn off Source dates support in the connection settings.
- Open a member
- Turn on Source dates support in the connection settings.
- Make some changes and save
- Source dates must be kept

### Checklist
* [x] have tested my change


worksofliam (approved):
Number two scared me the most! After you restart the extension host, the source dates gutter disappears as expected. I reconnected, made a change but the gutter doesn't reappear.

Good news is that it worked ok!

image

Great change. Merge when you can @sebjulliand :)

sebjulliand:
Lovely! Thanks @worksofliam !

themountinman:
Thank your for your time and effort addressing this. I have not had a recurrence of the issue since I reported (but I'm also not sure whether my version of the extension has been updated to include the version that includes this fix.

PR #2521, opened 2025-02-19
Default to using a diff when tracking source dates

worksofliam:
### Changes

Removes the ability to configure which mode will be used to track source date changes, and we now default to `diff`, which tracks changes based on the previous save of the file.

No functional changes here, only UX changes. Documentation may need to be updated after merge.

### How to test this PR

Examples:

1. Enable source dates, gutter optional but nice to have on
2. Make changes to a file (optionally see the gutter update)
3. Save the file and ensure all the source dates aren't set to today and only the changed lines are updated.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/filesystems/qsys/QSysFs.ts
+++ b/src/filesystems/qsys/QSysFs.ts
@@ -55,10 +55,6 @@ export class QSysFS implements vscode.FileSystemProvider {
         this.sourceDateHandler = new SourceDateHandler(context);
         this.extendedContent = new ExtendedIBMiContent(this.sourceDateHandler);
 
-        context.subscriptions.push(
-            onCodeForIBMiConfigurationChange(["connectionSettings", "showDateSearchButton"], () => this.updateMemberSupport()),
-        );
-
         instance.subscribe(
             context,
             'connected',
--- a/package.json
+++ b/package.json
@@ -481,11 +481,6 @@
 					"default": false,
 					"description": "If enabled, when the user tries to delete a directory, they will be asked to confirm the deletion by typing in the directory's name."
 				},
-				"code-for-ibmi.showDateSearchButton": {
-					"type": "boolean",
-					"default": true,
-					"description": "If enabled, will show the 'Date search' button in the status bar when source dates with diff mode is enabled."
-				},
 				"code-for-ibmi.terminals.5250.openInEditorArea": {
 					"type": "boolean",
 					"default": true,


src/filesystems/qsys/QSysFs.ts
```diff
@@ -82,7 +82,6 @@ export class QSysFS implements vscode.FileSystemProvider {
         if (connection && config?.enableSourceDates) {
             if (connection.sqlRunnerAvailable()) {
                 this.extendedMemberSupport = true;
-                this.sourceDateHandler.changeSourceDateMode(config.sourceDateMode);
```
SanjulaGanepola, line 85: I noticed that the `Date search` status bar item is not visible. Looks like it is because we removed this which called `this.sourceDateSearchBarItem.show()`.

SanjulaGanepola (changes requested):
Just a couple comments + we should update the description for `code-for-ibmi.showDateSearchButton`:

> If enabled, will show the 'Date search' button in the status bar when source dates with diff mode is enabled.

src/filesystems/qsys/QSysFs.ts
worksofliam: I am more surprised this was not caught by the build! I am going to remove this line.

src/filesystems/qsys/sourceDateHandler.ts
```diff
@@ -82,36 +79,17 @@ export class SourceDateHandler {
       this.enabled = enabled
       if (!this.enabled) {
         clearTimeout(this.timeout);
-        this.sourceDateSearchBarItem.hide();
+        this.sourceDateSearchBarItem.show();
```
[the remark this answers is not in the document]
worksofliam: Not sure! It should be hide here, but show if enabled.

src/filesystems/qsys/QSysFs.ts
worksofliam: Calling `setEnabled` below will now show it.

SanjulaGanepola (approved):
Changes look good and it works as expected now!

Going back to my previous comment, shouldn't we drop the "when source dates with diff mode is enabled." for `showDateSearchButton`?

worksofliam:
@SanjulaGanepola I am following now. I removed the `showDateSearchButton` config option.

SanjulaGanepola:
I was thinking just drop that part of the description, but your completely right in that we don't need that option all together. Looks good!

PR #2523, opened 2025-02-19
Add search item tooltip

chrjorgensen:
### Changes

This PR will add tooltips to the items in the search result view:

Members:
billede

Streamfiles:
billede

Streamfiles - find:
billede

Included in the PR is a fix for invalid dates from the `getMemberInfo` and `getMultipleMemberInfo` functions...

### How to test this PR

Examples:

1. Run the test cases
2. Search for some text in members and streamfile and check for tooltips in the result view.
3. Find some streamfiles and check for tooltips in the result view.

### Checklist

* [x] have tested my change

PR #2525, opened 2025-02-21
Refactor component handling and improve user input management

worksofliam:
* move `enableSQL` to the connection instance.
* enhance component manager checks for `NotChecked` state
    * Add missing away that was setting all cached states to `NotChecked`
* handle potential crashes when the `getMemberInfo` component is unavailable.
* fixed broken test introduced by another PR (where a date should have been returned correctly)

## How to test

1. Ensure Go to File works as expected for opening members:
    * `Try entering QSYSINC/*` to filter
    * `Then try `QSYSINC/H/*` to filter
    * Try opening `MATH`

### DIFF
--- a/src/api/components/getMemberInfo.ts
+++ b/src/api/components/getMemberInfo.ts
@@ -53,6 +53,24 @@ export class GetMemberInfo implements IBMiComponent {
     });
   }
 
+  private static parseDateString(tsString: string|undefined): Date | undefined {
+    if (!tsString) {
+      return undefined;
+    }
+    
+    const dateParts = tsString.split('-');
+    const timeParts = dateParts[3].split('.');
+
+    const year = parseInt(dateParts[0], 10);
+    const month = parseInt(dateParts[1], 10) - 1; // Months are zero-based in JavaScript
+    const day = parseInt(dateParts[2], 10);
+    const hours = parseInt(timeParts[0], 10);
+    const minutes = parseInt(timeParts[1], 10);
+    const seconds = parseInt(timeParts[2], 10);
+
+    return new Date(year, month, day, hours, minutes, seconds);
+  }
+
   async getMemberInfo(connection: IBMi, library: string, sourceFile: string, member: string): Promise<IBMiMember | undefined> {
     const config = connection.config!;
     const tempLib = config.tempLibrary;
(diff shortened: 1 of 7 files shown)


src/api/components/getMemberInfo.ts
```diff
@@ -78,8 +96,8 @@ export class GetMemberInfo implements IBMiComponent {
         name: result.MEMBER,
         extension: result.EXTENSION,
         text: result.DESCRIPTION,
-        created: new Date(result.CREATED ? Number(result.CREATED) : 0),
-        changed: new Date(result.CHANGED ? Number(result.CHANGED) : 0)
+        created: GetMemberInfo.parseDateString(String(result.CREATED)),
```
chrjorgensen, line 99: @worksofliam This issue was fixed in my PR #2523 in commit [128cbbf](https://github.com/codefori/vscode-ibmi/pull/2523/commits/128cbbf9a1dbf98839dee2280daa2107fd35a5cd) - which changed the SQL in `getMemberInfo` to act like the other places in C4i where we get and display item dates - as an EPOCH value...?

worksofliam: Sadly, it was not fixed. If you check out to the master branch and try and run the test, you'll see that they fail.

This change that you are commenting on is a specific fix for that.

worksofliam: Your timestamps are being returned to strings and we were trying to convert them to numbers before. Now I am passing the string that is returned instead.

worksofliam: @chrjorgensen I deleted my old comments. Yes, #2523 fixed this problem as well. So, when that PR has been merged, I will bring those changes into here as well.

src/commands/open.ts
```diff
@@ -408,6 +411,9 @@ export function registerOpenCommands(instance: Instance): Disposable[] {
             }
           }
         }
+
+        quickPick.busy = false;
+        quickPick.enabled = true;
```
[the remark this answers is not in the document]
worksofliam: @SanjulaGanepola I actually removed all the 'busy' logic in a new commit. Needs more work in a new PR.

chrjorgensen (approved):
@worksofliam Code changes for Go To file tested and approved. 👍 

Biggest issue during testing was that `GetMbrInfo` test failed - seems like the test creates a new, temp config for the host running the test? None of my connection profiles for that host uses the default temp library, so the function and the test sourcefile was not in the expected library. Something to be aware of!

worksofliam:
@chrjorgensen Thanks!!

odd about your tests not passing. I recommend deleting the `.json` files under the `tests` folder and then trying running your tests against.

image

Issue #2524, opened 2025-02-21
Sometimes `GetMemberInfo` throws an SQL error


worksofliam:
This input statement:

```sql
create table QTEMP.MEMBERINFO as (select * from table(ILEDITOR.GETMBRINFO('SNDBX3', 'QRPGLESRC', '#HELLO'))) with data;
select * from QTEMP.MEMBERINFO
```

Throws this result:

```
 **** CLI ERROR *****
         SQLSTATE: 22007
NATIVE ERROR CODE: -20448
Expression not valid using format string specified for TIMESTAMP_FORMAT. 

 **** CLI ERROR *****
         SQLSTATE: 42704
NATIVE ERROR CODE: -204
MEMBERINFO in QTEMP type *FILE not found.
```

worksofliam:
More info: `GETMBRINFO` returns no rows when the chosen member is empty.

chrjorgensen:
@worksofliam The error is easily reproduced:

1. Create a new sourcefile with an empty member: CRTSRCPF <library>/<file> <member>
2. Call table function GetMbrInfo: select * from GetMbrInfo( <library>, <file>, <member> )

The call will always throw the error.

I will investigate this...

PR #2528, opened 2025-02-21
Fix invalid dates from GetMbrInfo

chrjorgensen:
### Changes

This PR will fix the error reported in issue #2524 by returning the date `1970-01-01 00:00:00` when the source member date is not initialized, i.e. the date does not start with a '1' or '0'. This is the case when a member is created and not yet used: The QUSRMBRD API will return hexadecimal zeroes for the member change date if the member has never been refreshed.

A test for valid dates from GetMbrInfo has also been added.

### How to test this PR

Examples:

1. Run the test cases

### Checklist

* [x] have tested my change


worksofliam:
I mistakenly merged this PR. The tests didn't run here, and I mistook that as a pass. I ran them locally after I merged and they are failing (including the test case).

I have created a fix in #2525, and Christian has a change in #2523 which will also fix it.

PR #2548, opened 2025-03-04
Add deprecation warnings for outdated methods in IBMi and Instance classes

worksofliam:
Introduce deprecation warnings for outdated methods to guide users towards updated alternatives before their removal in future versions.


worksofliam:
@SanjulaGanepola Thanks. I updated it!

PR #2587, opened 2025-03-25
Fixed droping QSYS members into IFS folder

sebjulliand:
### Changes
For some reason, dropping members into the IFS browser now fails with this message.
image

Getting the members items from the `dataTransfer` object doesn't give the expected result (anymore?).
However, a URI list is automatically added to the `dataTransfer` when dragging items, so we now use that to perform the member copy to IFS operation.

### How to test this PR

1. Drag and drop a member from the Object Browser into an IFS folder in the IFS browser
2. The member must be copied to a streamfile in the target folder.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/ui/views/ifsBrowser.ts
+++ b/src/ui/views/ifsBrowser.ts
@@ -3,13 +3,13 @@ import path, { dirname, extname } from "path";
 import vscode, { FileType, l10n, window } from "vscode";
 
 import { existsSync, mkdirSync, rmdirSync } from "fs";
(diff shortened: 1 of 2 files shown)


chrjorgensen (approved):
@sebjulliand Good catch! I was just caught by this and happy to see your fix was already available. Works as expected - as usual. 👏

sebjulliand:
Thanks @chrjorgensen !
I still don't understand what happened to VS Code...this used to work just fine.

PR #2588, opened 2025-03-26
Added support for `/open` URI for opening members/streamfiles

sebjulliand:
### Changes
This PR refactors Code for IBM i URI handling and adds support for a new `/open` URI that allows to open a member or a streamfile.

The URI supports these query parameters:
| Name       | Required | Description                                                                                                                                  |
|:------------|:----------|:----------------------------------------------------------------------------------------------------------------------------------------------|
| `path`     | ✔        | A streamfile path (e.g. `/home/seb/todo.txt`) or a member path (e.g. `SEBLIB/QRPGLESRC/KABOOM.RPGLE`).                                       |
| `host`     | ❌        | The IBM i hosting the `path`. Can be an IP, a host name or a Code for IBM i connection name.<br/>Uses the current connection when not specified. |
| `user` | ❌        | If specified with `host`, a connection with the specified user must exist or the current connection must use this user.                                                                                              |
| `readonly` | ❌        | Forces the path to be opened in read-only mode.                                                                                              |

If the `host` parameter is not specified and Code for IBM i is not connected, an error will be displayed.
image

If `host` is specified and it doesn't match the current connection, a dialog will offer to switch the current connection.
image

### Examples:
- vscode://halcyontechltd.code-for-ibmi/open?path=SEBLIB/QRPGLESRC/KABOOM.RPGLE&readonly
- vscode://halcyontechltd.code-for-ibmi/open?host=PUB400.com&path=SEBLIB/QRPGLESRC/KABOOM.RPGLE
- vscode://halcyontechltd.code-for-ibmi/open?host=MY_PUB400_CONNECTION&path=/home/seb/todo.txt
- vscode://halcyontechltd.code-for-ibmi/open?host=MY_PUB400_CONNECTION&user=JOHNDOE&path=/home/seb/todo.txt
- vscode://halcyontechltd.code-for-ibmi/open?host=10.4.3.2&path=/home/seb/todo.txt

### Checklist
* [x] have tested my change


src/uri/handlers/open.ts
```diff
@@ -0,0 +1,122 @@
...
+  const host = await resolveHost(query.host as string | undefined);
+  let connect;
+  let cancel;
+
+  if (!connection && !host) {
+    throw l10n.t("Not connected to IBM i: 'host' query parameter is required");
+  }
+
+  if (connection && host) {
+    if (host.name !== connection.currentConnectionName) {
```
SanjulaGanepola, line 75: General question: Is it intended that `user` is not one of the query parameters (an optional one)?

If yes, this check would need to change as it is checking the connection name instead of the host name which can lead to this:
image

src/uri/handlers/open.ts
```diff
@@ -0,0 +1,122 @@
...
+}
+
+function toBoolean(value?: string) {
+  return value !== undefined && (value === "" || value.toLowerCase() === "true");
+}
+
+async function loadParameters(query: querystring.ParsedUrlQuery, connection?: IBMi): Promise<Parameters> {
+  const path = toPath(query.path as string | undefined);
+  const readonly: DefaultOpenMode | undefined = toBoolean(query.readonly as string | undefined) ? "browse" : undefined;
+  const host = await resolveHost(query.host as string | undefined);
```
SanjulaGanepola, line 66: For clarity, it might make more sense to rename this variable and function to something like: `const connectionData = await resolveConnectionData`...

SanjulaGanepola (changes requested):

src/uri/handlers/open.ts
sebjulliand: Good catch, that wasn't intended...but ovelooked 😅
I'll add `user` as an optional parameter.

src/uri/handlers/open.ts
sebjulliand: Agreed.

sebjulliand:
There you go @SanjulaGanepola : `user` can now be specified.

SanjulaGanepola (changes requested):
@sebjulliand I just realized I forgot to submit this 2nd review. Couple minor changes/suggestions below.

src/uri/handlers/open.ts
```diff
@@ -0,0 +1,132 @@
...
+  if (host) {
+    const userMatches = (connectionData: ConnectionData) => !user || connectionData.username.toLocaleUpperCase() === user;
+    const connectionByName = IBMi.connectionManager.getByName(host)?.data;
+    if (connectionByName && userMatches(connectionByName)) {
+      return connectionByName;
+    }
+
+    const ip = !/^((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)\.?\b){4}$/.test(host) ? (await dns.lookup(host)).address : host;
+    for (const connection of IBMi.connectionManager.getAll()) {
+      if ((await dns.lookup(connection.host)).address === ip && userMatches(connection)) {
```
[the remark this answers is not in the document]
sebjulliand: I don't think so. Looking up both IPs to compare them seems to me like the most reliable way of comparing two connection's host. It's reliable in the way that it can't take multiple forms (unless we bring IPv6 in the mix but we're not there yet). The host name, on the other hand, can be a short name (i.e. myibmi) or a long name (i.e. myibmi.mydomain), at least. So I guess we'd rather stick to comparing the IPs here 😉

sebjulliand:
Thank you @SanjulaGanepola ! Back at you 😉

src/uri/handlers/open.ts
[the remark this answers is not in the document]
SanjulaGanepola: Good point

SanjulaGanepola (approved):
Tested and works as expected. Great work!

SanjulaGanepola (approved):

sebjulliand:
Thank you for your review @SanjulaGanepola 😊

PR #2589, opened 2025-03-26
Added Compare with each other command

sebjulliand:
### Changes
This PR adds a `Compare with...each other` command that is enabled when exactly two members or streamfiles are selected in there respective browser:
image

The `Compare with...` submenu items have been renamed too, to avoid redundancy.
image

### How to test this PR
1. Select exactly two members or streamfiles
2. Select the `Each other` command in the `Compare with...` menu
3. Select just one member or streamfile: the menu must show 6 commands
4. Select more than two members/streamfiles: the menu must not be shown

### Checklist
* [x] have tested my change

### DIFF
--- a/package.json
+++ b/package.json
@@ -1071,9 +1071,15 @@
 				"category": "IBM i"
 			},
 			{
-				"command": "code-for-ibmi.compareWithSelected",
+				"command": "code-for-ibmi.compareWithSelected.short",
 				"enablement": "code-for-ibmi:connected",
 				"title": "Selected",
+				"category": "IBM i"
+			},
+			{
+				"command": "code-for-ibmi.compareWithSelected",
+				"enablement": "code-for-ibmi:connected",
+				"title": "Compare with selected",
 				"category": "IBM i"
 			},
 			{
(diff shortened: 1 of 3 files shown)


chrjorgensen (changes requested):
@sebjulliand Your change is great, works as expected and simplifies the UX.

I have some reservations, though:
1. You should not have `...` in the submenu title for the compare menu items - `...` is for additional UI to follow, which is not the case here (the submenu is not a new UI item). It should be the same as the `Sort by` submenu title: 
image

2. You renamed the titles for the compare commands and made the submenu items very slick 😍 
image
However, that also caused the titles to change in the Command Palette to change, and the commands are now not intuitive in the palette. Before: 
image
After: 
image
image
...and so on.

Could this be fixed to give a better UX?

sebjulliand:
Thanks @chrjorgensen !
Have another look, the compare actions in the palette will now have their full title whereas they have the shortened title in the menu.

I also made a modification to show the "Open file dialog" when comparing against a local file - I thought it would be more user-friendly than typing the local path in a prompt 😅

sebjulliand:
I just fixed the `compareWithActiveFile` command too 😊

PR #2603, opened 2025-04-07
Standard library list API

worksofliam:
Introduce a new instance API that fetches the library list and current library settings based on the connection and optional workspace folder.

This will be used in many places:

* RPGLE language server
* Project Explorer
* Database extension

### DIFF
--- a/src/Instance.ts
+++ b/src/Instance.ts
@@ -8,6 +8,8 @@ import { VsCodeConfig } from "./config/Configuration";
 import { EventEmitter } from "stream";
 import { ConnectionStorage } from "./api/configuration/storage/ConnectionStorage";
 import { VscodeTools } from "./ui/Tools";
+import { ILELibrarySettings } from "./api/CompileTools";
+import { getEnvConfig } from "./filesystems/local/env";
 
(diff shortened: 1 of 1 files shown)


src/Instance.ts
```diff
@@ -198,6 +200,19 @@ export default class Instance {
...
+  async getLibraryList(connection: IBMi, workspaceFolder?: vscode.WorkspaceFolder): Promise<ILELibrarySettings> {
+    const config = connection.getConfig();
+
+    const env = workspaceFolder ? (await getEnvConfig(workspaceFolder)) : {};
+
+    const librarySetup: ILELibrarySettings = {
+      currentLibrary: env[`&CURLIB`] || config.currentLibrary,
+      libraryList: env[`&LIBL`]?.split(` `) || config.libraryList,
+    };
+
```
SanjulaGanepola, line 212: Should we remove duplicates like we do in `CompileTools` or leave that up to the caller?
```suggestion
librarySetup.libraryList.filter(Tools.distinct)
```

worksofliam: Yup, let's leave it to the caller!

PR #2620, opened 2025-04-19
Enabled multi-selection support for several commands

sebjulliand:
### Changes
As discussed here: https://github.com/orgs/codefori/discussions/2612

This PR enables multi-selection support for the following commands:
- Generate binder source
- Browse/edit members
- Browse/edit IFS files
- Maintain filter
- Custom actions
- Search source file
- Search in IFS
- Find in IFS

#### Run multiple custom actions
Some noticeable changes were done to allow running custom actions on multiple items.
- All the selected items must share the same nature (i.e. URI scheme) (otherwise a warning is shown)
- If the command must be prompted ~~, it will do so for each item~~:
  - If the action is run on a single item, it will be prompted as usual
  - If the action is run on multiple items, it will be prompted once, keeping the variables placeholders
- The whole execution can now be cancelled. It will stop the process after the current item is done being processed
image

- The output webview had been enhanced to display multiple outputs. The left side displays items' path and the right side shows the selected item output. The icon in front of the item shows the execution result (ok, failed or not run).
image
- Running an action on a single item will only show the single output (no need for navigating through multiple results)

#### Search in members / Search/find in streamfiles
Finding or searching can be cancelled. The search will stop after the current path is done being processed.
Once a path has been processed, the results are added to the Search Results view, which means the view gets populated dynamically while the search is running.

### How to test this PR
1. Use `ctrl+click` or `shift+click` to select multiple items in the Object browser or IFS browser
2. Check that actions are consistently shown
3. The the actions mentioned in the description

### Checklist
* [x] have tested my change

### DIFF
--- a/src/commands/actions.ts
+++ b/src/commands/actions.ts
@@ -1,6 +1,7 @@
 import path from "path";
 import { commands, Disposable, l10n, TreeItem, Uri, window, WorkspaceFolder } from "vscode";
 import IBMi from "../api/IBMi";
+import { Tools } from "../api/Tools";
 import Instance from "../Instance";
 import { Action, DeploymentMethod } from "../typings";
 import { runAction } from "../ui/actions";
@@ -37,7 +38,7 @@ export function registerActionsCommands(instance: Instance): Disposable[] {
         const scheme = uris[0]?.scheme;
         if (scheme) {
           if (!uris.every(uri => uri.scheme === scheme)) {
-            window.showWarningMessage(l10n.t("Actions can't be run on multiple items of different natures."))
+            window.showWarningMessage(l10n.t("Actions can't be run on multiple items of different natures. ({0})", uris.map(uri => uri.scheme).filter(Tools.distinct).join(", ")));
             return false;
           }
 
--- a/src/commands/actions.ts
+++ b/src/commands/actions.ts
@@ -22,7 +22,7 @@ export function registerActionsCommands(instance: Instance): Disposable[] {
           }
         }
         else {
-          for (const target of (items || [item])) {
+          for (const target of (Array.isArray(items) ? items : [item])) {
             if (target instanceof Uri) {
               uris.push(target);
             }
--- a/src/commands/actions.ts
+++ b/src/commands/actions.ts
@@ -1,67 +1,81 @@
 import path from "path";
-import { commands, TreeItem, Uri, WorkspaceFolder, window, Disposable } from "vscode";
-import { refreshDiagnosticsFromServer } from "../ui/diagnostics";
+import { commands, Disposable, l10n, TreeItem, Uri, window, WorkspaceFolder } from "vscode";
+import IBMi from "../api/IBMi";
 import Instance from "../Instance";
 import { Action, DeploymentMethod } from "../typings";
 import { runAction } from "../ui/actions";
-import IBMi from "../api/IBMi";
+import { refreshDiagnosticsFromServer } from "../ui/diagnostics";
 import { BrowserItem } from "../ui/types";
 
 export function registerActionsCommands(instance: Instance): Disposable[] {
   return [
-    commands.registerCommand(`code-for-ibmi.runAction`, async (target: TreeItem | BrowserItem | Uri, group?: any, action?: Action, method?: DeploymentMethod, workspaceFolder?: WorkspaceFolder) => {
+    commands.registerCommand(`code-for-ibmi.runAction`, async (item?: (TreeItem | BrowserItem | Uri), items?: (TreeItem | BrowserItem | Uri)[], action?: Action, method?: DeploymentMethod, workspaceFolder?: WorkspaceFolder) => {
       const connection = instance.getConnection()!;
-      const editor = window.activeTextEditor;
-      let uri;
-      let browserItem;
-      if (target) {
-        if ("fsPath" in target) {
-          uri = target;
-        }
-        else {
-          uri = target?.resourceUri;
-          if ("refresh" in target) {
-            browserItem = target;
+      if (connection) {
+        const editor = window.activeTextEditor;
+        const browserItems: BrowserItem[] = [];
+        const uris: Uri[] = [];
+        if (!item) {
+          if (editor?.document.uri) {
+            uris.push(editor?.document.uri);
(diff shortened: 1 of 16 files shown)


sebjulliand:
Edit: on second thought, prompted custom actions should be prompted once and executed many times instead of prompting each item.

src/webviews/filters/index.ts
```diff
@@ -1,14 +1,14 @@
...
 import { instance } from "../../instantiate";
-import IBMi from "../../api/IBMi";
 import { ObjectFilters } from "../../typings";
+import { CustomUI } from "../CustomUI";
 
 export async function editFilter(filter?: ObjectFilters, copy = false) {
   const connection = instance.getConnection();
-  const config = instance.getConfig();
+  const config = connection?.getConfig();
   if (config) {
```
worksofliam, line 10: Instead of `connection?`, change your `if` statement to:

```
if (connection) {
```

and then define `config` inside of it.

src/ui/actions.ts
```diff
@@ -1,26 +1,38 @@
...
+import { CustomUI, TreeListItem } from '../webviews/CustomUI';
 import { BrowserItem } from './types';
 
-interface CommandObject {
+type CommandObject = {
   object: string
   library?: string
 }
 
+type ActionTarget = {
```
worksofliam, line 24: Any chance this type could just extend our existing `action` type or is it really not applicable here?

src/ui/actions.ts
```diff
@@ -30,22 +42,29 @@ export function registerActionTools(context: vscode.ExtensionContext) {
...
+      protected: parseFSOptions(uri).readonly || config?.readOnlyMode,
+      workspaceFolder: workspaceFolder || vscode.workspace.getWorkspaceFolder(uri),
+      executionOK: false,
+      hasRun: false,
+      processed: false,
+      output: []
+    }) as ActionTarget);
 
-    const isProtected = uriOptions.readonly || config?.readOnlyMode;
+    workspaceFolder = targets[0].workspaceFolder;
```
worksofliam, line 66: Do we need some logic here to make sure all the targets are from the same workspace folder?

src/ui/actions.ts
sebjulliand: No; since an `ActionTarget` is a recipient of the action execution, it would not make sense.

src/webviews/filters/index.ts
sebjulliand: Done!

sebjulliand:
Thanks @worksofliam ; see my comment and give it a go 😀

src/ui/actions.ts
worksofliam: @sebjulliand That's up to you. If it's easy to add as part of this PR, go for it!

worksofliam:
@sebjulliand Generally it's looking good. I want to also add at least one test case for each file system in `src/testing/action.ts` which run an action against multiple files. What do you think?

src/ui/actions.ts
sebjulliand: It's done 😁

sebjulliand:
@worksofliam tests are in, one for each scheme as requested.
I'll let you have a final look. Thanks!

worksofliam (approved):
It's looking great. This is a serious UX improvement for everyone! Fab work here. Please see my comment for peace of mind. Merge when ready.

image

worksofliam (changes requested):
Whoops, pressing the Run Action button on the editor UI causes this error.

image

src/commands/actions.ts
```diff
@@ -1,67 +1,82 @@
...
 import { Action, DeploymentMethod } from "../typings";
 import { runAction } from "../ui/actions";
-import IBMi from "../api/IBMi";
+import { refreshDiagnosticsFromServer } from "../ui/diagnostics";
 import { BrowserItem } from "../ui/types";
 
 export function registerActionsCommands(instance: Instance): Disposable[] {
   return [
-    commands.registerCommand(`code-for-ibmi.runAction`, async (target: TreeItem | BrowserItem | Uri, group?: any, action?: Action, method?: DeploymentMethod, workspaceFolder?: WorkspaceFolder) => {
+    commands.registerCommand(`code-for-ibmi.runAction`, async (item?: (TreeItem | BrowserItem | Uri), items?: (TreeItem | BrowserItem | Uri)[], action?: Action, method?: DeploymentMethod, workspaceFolder?: WorkspaceFolder) => {
```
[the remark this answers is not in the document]
sebjulliand: Yep, it was put here as a placeholder, when Project Explorer started to call it I think. So that second argument was mandatory in case the command was called with multiple item selected (otherwise it would have shifted the other arguments).
Long story short: it was never used (it was grayed out in the editor), now it is. 🤘🏻

sebjulliand:
@worksofliam I can't get the same error pressing the editor button. What's the stack trace in the console?

sebjulliand:
> Whoops, pressing the Run Action button on the editor UI causes this error.

Fixed it! Turned out calling a command from the editor UI sends the command's group id as the second parameter, and not `undefined` or an array as when run from a tree view.

worksofliam (approved):
Working now! Thanks for the fix!

sebjulliand:
> @sebjulliand I've just used your excellent change for the first time (using the prerelease version), and I have an idea for improvement:
> 
> Can we show the path of the member or streamfile for which the action is being run? Currently it's a count, but it would be nice to know the file also - especially if something hangs or is slow.
> 
> WDYT? 🤔

Thank you @chrjorgensen !
It sure is a good idea - do you want to take care of it? Else I can, no problem, I have the exact location under my eyes 😁

chrjorgensen:
@sebjulliand Unfortunately I'm still very busy with work, so I don't have the time... 😭 
but I would be grateful if you can fix this. 🙏

sebjulliand:
> @sebjulliand Unfortunately I'm still very busy with work, so I don't have the time... 😭 but I would be grateful if you can fix this. 🙏

You can count on me 🤝🏻

sebjulliand:
There you go @chrjorgensen : https://github.com/codefori/vscode-ibmi/pull/2674

PR #2633, opened 2025-04-29
Add setting to write Actions output to an IFS file

sebjulliand:
### Changes
As discussed in https://github.com/orgs/codefori/discussions/2630

This PR adds a new setting to the Action definition - namely `outputFile`:
image

This field expects an IFS path and supports variables replacement as well as the special `&i` placeholder that, if present in the path, will act as an index that gets computed when the output is written (to avoid overwriting the previous output).

`~` is also supported and will be replaced by the user actual home directory.

For example, this pattern `~/outputs/&CURLIB_&OPENMBR&i.txt` will create these files if the action is run three time:
image

If the pattern is `~/outputs/&CURLIB_&OPENMBR&i.txt`, then a single file get written over and over again:
image

A new `Variables` object has been created to isolate the variables management (see src/api/variables.ts).

### How to test this PR

Examples:
1. Define an output file path in an action
2. Run the action
4. Check the file was created with the expected name in the IFS

### Checklist
* [x] have tested my change

### DIFF
--- a/package.json
+++ b/package.json
@@ -606,6 +606,10 @@
 							"runOnProtected": {
 								"type": "boolean",
 								"description": "If enabled, this Action will be allowed to run on protected/read only targets."
+							},
+							"outputToFile": {
+								"type": "string",
+								"description": "If defined, a copy of the output will be put in the defined location; supports variable replacement."
 							}
 						}
 					},
--- a/src/api/CompileTools.ts
+++ b/src/api/CompileTools.ts
@@ -1,7 +1,8 @@
 
 import IBMi from './IBMi';
 import { Tools } from './Tools';
-import { Variable, RemoteCommand, CommandResult, StandardIO } from './types';
+import { CommandResult, RemoteCommand, StandardIO } from './types';
+import { Variables } from './variables';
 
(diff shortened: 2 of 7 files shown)


sebjulliand:
> Sébastien, thanks for adding this action. You were were very expedient!

You're welcome @ToshBimbra (hopefully it will be approved soon).

> I must admit I am confused about how this will help me save a compile listing into some IFS location. I guess I am missing the overall steps in how to apply this action for what I am needing to do.

The idea is that on the actions you'll use to compile objects, you'll define the file path were the output should be saved. Each time you'll run that action, it will write the output (i.e. compile listing in your case) in an IFS file. Since the file path supports variable replacement, you can make the file bear the name of the compiled object to find it more easily.

ToshBimbra:
Thx! I got it working. 
image
image

sebjulliand:
Great! Thanks for beta testing it 😁

worksofliam:
@sebjulliand It's looking good!

image

worksofliam (approved):
I am happy with the PR. You're welcome to get an additional review from someone else, or merge at will. Thanks!

PR #2649, opened 2025-05-05
Fixed member search path parsing

### LINKED ISSUE #2647: Bug - Search Source File results with members containing a . in the member name
### Discussed in https://github.com/orgs/codefori/discussions/2646

<sup>Originally posted by **nick-roux** May  5, 2025</sup>
When running the Search Source File action Code4i will find text in source members and member should open when clicking on the found text in the search results.

This works fine for members that do not have a . in the member name, but fails if the member name does contain a . ,reporting that the member does not exist.

Tested environment
VSCode: 1.99.3
Code4i: Release 2.15.3 and Pre-Release 2.15.4

code4i-dotbug
</div>

sebjulliand:
### Changes
Fixes #2647 .

Searching through a source file containing members with a `.` in their name would prevent them from being opened from the Search Result view. The paths displayed in the search result view were also truncated.

The member path parsing from grep results has been fixed and it works fine now.

### How to test this PR
1. Create a few members with a `.` in their name
image
2. Search for a term that can be found in several of them
3. Make sure path are complete (name + type) and that each result can be opened
image

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/Search.ts
+++ b/src/api/Search.ts
@@ -1,30 +1,43 @@
-import * as path from 'path';
 import { GetMemberInfo } from './components/getMemberInfo';
-import { IBMiMember, SearchHit, SearchResults } from './types';
-import { Tools } from './Tools';
 import IBMi from './IBMi';
+import { Tools } from './Tools';
+import { IBMiMember, SearchHit, SearchResults } from './types';
 
 export namespace Search {
-  export async function searchMembers(connection: IBMi, library: string, sourceFile: string, searchTerm: string, members: string|IBMiMember[], readOnly?: boolean,): Promise<SearchResults> {
+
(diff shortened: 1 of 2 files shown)


worksofliam (approved):
Thanks, and the tests are passing - great!

A new test case for a member name with a `.` is welcome, but optional.

sebjulliand:
> Thanks, and the tests are passing - great!
> 
> A new test case for a member name with a `.` is welcome, but optional.

OK I'll add a test case but just because it's you.

sebjulliand:
There you go @worksofliam. I'll let you merge if you're happy with the test case.

worksofliam (approved):
@sebjulliand Merge whenever you're ready!

PR #2652, opened 2025-05-06
Fix #2651 variable expansion test is failing

### LINKED ISSUE #2651: Variable expansion test is failing
# Description

There are two different ways that the ***Variable expansion test*** is failing.
1. It fails in all contexts due to variable assignment problem introduced a few days ago (98a1b8880084dbea476fad43888a861921d52ecf)
2. Assuming 1 is fixed, it still fails when run independently using the ***Extension Tests (Individual)*** option

## 1. Issue with variable assignment
Commit 98a1b8880084dbea476fad43888a861921d52ecf made some changes to the variable handling logic. It lost this `if ((/^[A-Za-z\&]/i ...` (line 119) test which makes it only apply variables with usable names (beginning with A-Z or "&") ...
https://github.com/codefori/vscode-ibmi/blob/4d66d17130cb2997332e0f5aec30c688a60dded5/src/api/CompileTools.ts#L118-L122

.. when moving the logic here (line 54):
https://github.com/codefori/vscode-ibmi/blob/f59c1511b3003fb18e34d2af1daf32f7a7693b10/src/api/variables.ts#L51-L57

This is resulting in the ***Variable expansion test*** failing because of some of the hardcoded variables that are created in the Variables class. They have names that are not assignable when it tries to pass variables over to the PASE environment. I know it is impacting this test, but I am not sure what else it impacts. I just wanted the tests to pass before I attempted any other changes.

These variables are assigned in the constructor of the Variables class when there is a connection:
https://github.com/codefori/vscode-ibmi/blob/f59c1511b3003fb18e34d2af1daf32f7a7693b10/src/api/variables.ts#L14-L22

### Screenshots
Image

## 2. Issue with running individually
Assuming the above issue is fixed, when trying to run the Variable expansion test individually, it is failing because it depends on configuration that occurs only when running the entire test set.

The `before` function (line 34) is responsible for initializing the `currentLibrary` variable (lines 57-58). That variable is needed for the `assert.strictEqual` test (line 74).
https://github.com/codefori/vscode-ibmi/blob/f59c1511b3003fb18e34d2af1daf32f7a7693b10/src/testing/action.ts#L37-L76

The `before` function of the test suite is not run when running tests in the ***individual*** mode, here (line 80):
https://github.com/codefori/vscode-ibmi/blob/f59c1511b3003fb18e34d2af1daf32f7a7693b10/src/testing/index.ts#L70-L88

Changing this to first run the `before` function and then the test case will resolve the issue with the test.

### Screenshots
 Image

# To Reproduce
1. Run the ***Extension Tests (All)***
2. Run the ***Extension Tests (Individual)*** and click on the **Variable expansion test**

## Expected behavior
Tests should be in a passing state

## Environment (please complete the following information)
 - Extension version 2.15.5-dev.0
 - IBM i OS version 7.5

# Additional context
I'll have a PR to resolve this momentarily, but I wanted to document the defect since I'm new around here and the PR might be confusing without the context. 🙂

bspotswood:
### Changes

1. Resolve issue with variable names by restoring a filter that was in previous code that was recently refactored (98a1b8880084dbea476fad43888a861921d52ecf)
2. Run the `before` function for individual test runs when the test suite has a `before`

This resolves specific problems documented in #2651 when trying to run the ***Variable expansion test***

### How to test this PR

Examples:

1. Build the code using `npm run webpack-dev`
2. Launch an **Extension Tests** instance from the **Run and Debug** section in VSCode
3. Open a workspace in the launched window, if not already open
4. Connect to an IBM i system
5. If running individually: Click on the ***Variable expansion test*** option from the **Test Cases** panel
6. If **not** running individually: The tests will execute after connection

Before changes, the ***Variable expansion test*** test will fail. After the changes, the test will pass.

### Checklist

* [x] have tested my change
* [x] have ~created~ _fixed_ one or more test cases

### DIFF
--- a/src/api/variables.ts
+++ b/src/api/variables.ts
@@ -51,7 +51,9 @@ export class Variables extends Map<string, string> {
   toPaseVariables(): Record<string, string> {
     const variables: Record<string, string> = {};
     for (const [key, value] of this.entries()) {
-      variables[key.startsWith('&') ? key.substring(1) : key] = value;
+      if ((/^[A-Za-z\&]/i).test(key)) {
+        variables[key.startsWith('&') ? key.substring(1) : key] = value;
+      }
     }
     return variables;
   }
--- a/src/api/variables.ts
+++ b/src/api/variables.ts
@@ -51,8 +51,9 @@ export class Variables extends Map<string, string> {
   toPaseVariables(): Record<string, string> {
     const variables: Record<string, string> = {};
     for (const [key, value] of this.entries()) {
-      if ((/^[A-Za-z\&]/i).test(key)) {
-        variables[key.startsWith('&') ? key.substring(1) : key] = value;
+      const cleanKey = key.startsWith('&') ? key.substring(1) : key;
+      if ((/^[a-z_]\w*$/i).test(cleanKey)) {
+        variables[cleanKey] = value;
       }
     }
     return variables;
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,3 +61,4 @@ Thanks so much to everyone [who has contributed](https://github.com/codefori/vsc
 * [@Detrytus59](https://github.com/Detrytus59)
 * [@janfh](https://github.com/janfh)
 * [@MohitKambli](https://github.com/MohitKambli)
+* [@bspotswood](https://github.com/bspotswood)
--- a/src/api/tests/suites/variables.test.ts
+++ b/src/api/tests/suites/variables.test.ts
@@ -9,23 +9,26 @@ describe(`variables tests`, { concurrent: true }, () => {
     expect(variables.size).toBe(0);
     variables.set("&FOO", "bar")
       .set("&FOOTWO", "bartwo")
-      .set("&FOOTHREE", "barthree");
-  }, CONNECTION_TIMEOUT),
+      .set("&FOOTHREE", "barthree")
+      .set("{FOOFOUR}", "barfour");
+  }),
 
     it('Variables get', () => {
-      expect(variables.size).toBe(3);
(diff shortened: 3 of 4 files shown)


sebjulliand:
Hi @bspotswood , thanks for this PR!
Could you resolve the conflict (due to this issue being partially fixed, but putting back the initial test like you did sounds like a better idea); then we'll get this reviewed.
Thanks!

bspotswood:
Merged with main and accepted my changes to resolve the conflict.

src/api/variables.ts
```diff
@@ -51,9 +51,8 @@ export class Variables extends Map<string, string> {
   toPaseVariables(): Record<string, string> {
     const variables: Record<string, string> = {};
     for (const [key, value] of this.entries()) {
-      //Env variables can't contain * or {}; keeping only &xxx variables
-      if(key.startsWith('&')){
-        variables[key.substring(1)] = value;
+      if ((/^[A-Za-z\&]/i).test(key)) {
+        variables[key.startsWith('&') ? key.substring(1) : key] = value;
       }
```
sebjulliand, line 56: To be thorough, we should test the actual [naming rule for shell variable](https://www.gnu.org/software/bash/manual/html_node/Definitions.html#index-name), stripping `&` beforhand.
```suggestion
      const cleanKey = key.startsWith('&') ? key.substring(1) : key;
      if ((/^[a-z_]\w*$/i).test(cleanKey)) {
        variables[cleanKey] = value;
      }
```

sebjulliand (approved):

sebjulliand:
Perfect, let's merge. Thank you @bspotswood !

PR #2687, opened 2025-05-22
Complete actions schema

sebjulliand:
### Changes

- Added the recently added `outputToFile` attribute to the local actions' schema and cleans it up.
- Added a snippet to easily define new actions.
image
image
- Added content assist to show variables on the `outputToFile` field since its supports variables replacement
- Moved the call to the `outputToFile` method before post download gets executed, so the output can be downloaded locally if needed

### How to test this PR
1. Open a local `.vscode/actions.json` file
2. Use content assist to check `outputToFile` is suggested as a field when defining an action
3. Use content assist when defining the `outputToFile` value to see the available variables
4. Define a value for an action's `outputToFile` and add the output file's parent folder path in `postDownload` then check the output file is downloaded. 

### Checklist

* [x] have tested my change

### DIFF
--- a/schemas/actions.json
+++ b/schemas/actions.json
@@ -1,80 +1,79 @@
 {
-  "$id": "#/actions",
   "type": "array",
   "title": "Actions",
   "description": "List of Actions that apply to this Workspace. Actions may be used to run commands on the remote system.",
-  "default": [],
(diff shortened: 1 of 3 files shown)


worksofliam (approved):
LGTM! 

image

PR #2722, opened 2025-06-10
Show warning for unsupported minimum debug version

### LINKED ISSUE #2719: Display upgrade message if user has v2 debug service
The newer Code for i (v2.16) only supports v3 debug service. If the user has an older v2.0.x debug service, we can display a message telling the user to upgrade the debug service to v3.

worksofliam:
Implement a warning message when the debug service version is below the minimum required version, prompting users to update the debug service PTF.

Closes #2719

### DIFF
--- a/src/debug/index.ts
+++ b/src/debug/index.ts
@@ -4,16 +4,16 @@ import Instance from "../Instance";
 import path from "path";
 import * as vscode from 'vscode';
 
+import { ILELibrarySettings } from "../api/CompileTools";
+import { getDebugServiceDetails, ORIGINAL_DEBUG_CONFIG_FILE, resetDebugServiceDetails } from "../api/configuration/DebugConfiguration";
+import IBMi from "../api/IBMi";
+import { getStoredPassword } from "../config/passwords";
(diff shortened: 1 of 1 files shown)


sebjulliand (changes requested):
The change by itself is OK.
But do we really want to bother users with a message about upgrading the debugger every time they connect, especially if they can't do anything about it? 😅

I'm thinking we could show the message, and offer two actions: `Remind me` (show the message next time the user connects), `Dismiss` (don't show the message ever again). The action taken should be stored in the global storage so it doesn't get synced.

What do you think?

worksofliam:
@sebjulliand Great idea. Take another look and let me know what you think.

sebjulliand (changes requested):
One small change (maybe) and we'll be good to go.

worksofliam:
@sebjulliand One more time! 😄

sebjulliand:
I made a change so that the warning only shows if the debug service is actually installed. We don't want to bother people with no debug service 😅
And I fixed the message that was showing `[Object object]` instead of the actual version.
Have a look and merge whenever you want.

PR #2724, opened 2025-06-11
Ensure unique connection storage per session

worksofliam:
Implement a mechanism to set a unique key prefix for connection storage, ensuring that each connection maintains its own distinct storage context. This change enhances the organization and retrieval of connection-specific data.


sebjulliand (approved):
Good job @worksofliam , that fixed it 😄 
Merge away!

Issue #2366, opened 2024-11-25
Just an idea, is using `validQsysName` here for validation more robust than these individual checks?


worksofliam:
Just an idea, is using `validQsysName` here for validation more robust than these individual checks?

_Originally posted by @SanjulaGanepola in https://github.com/codefori/vscode-ibmi/pull/2334#discussion_r1856968821_

PR #2731, opened 2025-06-13
Update IBMiContent.ts to use validQsysName

Nathaniel-King-Navarrete:
Use validQsysName rather than individual checks. #2366 

### Changes
Removed individual checks within `validateLibraryList` method and changed it to `IBMI.validQsysName` method.

### How to test this PR
I tested this change by using the `Add to Library List` command. I tested adding a library name of '!@#$%'.

I wasn't able to test the length > 10 case, because there is a check in `code-for-ibmi.addToLibraryList` for that case, before the call to `validateLibraryList`.

 I noticed that there was some test suites that tested `validateLibraryList`, but I wasn't sure how to run them. Is there a way for me to call these test suites? Thanks.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -453,17 +453,12 @@ export default class IBMiContent {
 
     newLibl = newLibl
       .filter(lib => {
-        if (lib.match(/^\d/)) {
-          badLibs.push(lib);
-          return false;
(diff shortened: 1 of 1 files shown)


sebjulliand (approved):
Looks good to me; good idea there!

PR #2746, opened 2025-06-19
Fixed download of members with variant characters

### LINKED ISSUE #2728: Members that start with special characters that can  not be created as output on a windows device
(submitting on behalf of a client at Liam's request)

I have a task that requires I download source members and I am using the code for i extension since it's UI is the easiest that does so easily while maintaining the source member name and type. We have some members that start with special characters that can not be created as output on a windows device and was wondering if there may be a workaround as I can't see to find it in the discussions portion of the site that I can implement timely.

"Error downloading member(s)! Error: Object not copied.  Object is /QSYS.LIB/QTEMP.LIB/QTEMPSRC.FILE/SRFMAEA\$.MBR. (38501)"
Error downloading member(s)! Error: Object not copied.  Object is /QSYS.LIB/QTEMP.LIB/QTEMPSRC.FILE/SRFMAEA\$.MBR. (38501)
Error running command code-for-ibmi.downloadMemberAsFile: Cannot read properties of undefined (reading 'name'). This is likely caused by the extension that contributes code-for-ibmi.downloadMemberAsFile.

CCSID : 37
member: SRALPH$16
download C:\Users\cnix\Downloads\subfolder name here

sebjulliand:
### Changes

Fixes #2728 

Downloading a member with the `$` or `à` variant characters would fail because:
- `$` was escaped by `Tools.qualifyPath`
- `à` would be uppercased when `CPYTOSTMF` was run (don't know exactly why...it is sent lowercased by us)

Using a dummy member name while transferring the member to a stream file fixes both issues.

### How to test this PR
1. Create a member with `$` (CSSID 37 or 297) or `à` (CSSID 297) in its name
2. Right click and download it

### Checklist

* [x] have tested my change

### DIFF
--- a/src/ui/views/objectBrowser.ts
+++ b/src/ui/views/objectBrowser.ts
@@ -943,8 +943,8 @@ Do you want to replace it?`, item.name), { modal: true }, skipAllLabel, overwrit
                 .filter(item => item.copy)
                 .map(item =>
                   [
-                    `@QSYS/CPYF FROMFILE(${item.member.library}/${item.member.file}) TOFILE(QTEMP/QTEMPSRC) FROMMBR(${item.member.name}) TOMBR(${item.member.name}) MBROPT(*ADD) CRTFILE(*YES);`,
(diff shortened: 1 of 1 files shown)


worksofliam (approved):
Changes are good but long term strategy here is to move this logic and using `runSql` out of the object explorer class and into `IBMiContent.`.

Issue #2344, opened 2024-11-12
Increase VS Code scroll back by default


worksofliam:
Because the spool files keep getting cut off for new users.

<img width="682" alt="image" src="https://github.com/user-attachments/assets/cd4c53a8-4917-4437-b7ba-109fcb8f3419">

rbeckham:
@worksofliam what are you thinking?  default value of 10,000 ?
this should probably be a minimum value....incase the user already has the scrollback setting set to something larger.
For example, If we decide on 10,000 and the user has went in and made it 20,000 on their own we wouldn't want to set it back to 10,000.  I can take a look at this issue.

worksofliam:
@rbeckham Sounds like you're on the right track!

PR #2749, opened 2025-06-21
#2344 Increase VS Code scroll back by default

Nathaniel-King-Navarrete:
If the scrollback is less than a minimum amount, then set it to that minimum amount.

### Changes
During the instantiate process, set the scollback to a minimum amount. This will globally change the terminal.integrated.scrollback value. If the user has a prefered scollback larger than the minimum, then it will not be update.

### How to test this PR
Two test cases:
1. Scollback value is defined as less than minimum
     1.  terminal.integrated.scrollback within settings.json will be updated to 10000.
2. Scrollback value is defined as equal to minimum or more than minimum.
     1. terminal.integrated.scrollback within settings.json will remain unchanged.

### Checklist

* [x] have tested my change
* [] have created one or more test cases
* [] updated relevant documentation

### DIFF
--- a/src/instantiate.ts
+++ b/src/instantiate.ts
@@ -94,6 +94,11 @@ export async function loadAllofExtension(context: vscode.ExtensionContext) {
     })
   );
 
+  const MINIMUM_SCROLLBACK = 10000;
+  if(<number>vscode.workspace.getConfiguration().get('terminal.integrated.scrollback') < MINIMUM_SCROLLBACK){
+    vscode.workspace.getConfiguration().update('terminal.integrated.scrollback', MINIMUM_SCROLLBACK, true);
+  }
+
   // Register git events based on workspace folders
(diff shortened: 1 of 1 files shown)


sebjulliand:
Hi @Nathaniel-King-Navarrete ,
why should Code for IBM i change that global setting?

Nathaniel-King-Navarrete:
Hi @sebjulliand ,

I was using this line as reference point as a possible implementation.

https://github.com/codefori/vscode-ibmi/blob/1d00df81270ce5e82afcf098e60d5fbcca1c24c7/src/ui/Terminal.ts#L109

Are you thinking that the global is unnecessary for the scrollback?

Nathaniel-King-Navarrete:
Sounds good! I will make those changes, when I get the chance.

Nathaniel-King-Navarrete:
@sebjulliand Sounds good. I will make the changes as requested.

PR #2754, opened 2025-06-24
Add additional check for SIGABRT

worksofliam:
Looks like there is a bug with our node library. We have an instance when an IBM i system is not setup correctly, then SIGABRT is thrown, but the `code` is still `0` - which in most of the extension, indicates success.

```
/home/OLVERA: /home/OLVERA/.vscode/cqsh_1
echo "Hello world"
{
    "code": 0,
    "signal": "SIGABRT",
    "stdout": "",
    "stderr": ""
}
```

I've added some logic to our statement executor to set the code to `127` when `SIGABRT` is thrown.

I have also added a page in the docs to track this issue (which we've now seen a couple of times in the wild, but we are still unsure of the cause): https://codefori.github.io/docs/tips/setup/#qsh--cqsh-returning-sigabrt

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -1115,6 +1115,7 @@ export default class IBMi {
 
     // Some simplification
     if (result.code === null) result.code = 0;
+    if (result.signal === `SIGABRT`) result.code = 127;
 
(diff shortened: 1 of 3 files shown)


sebjulliand (approved):
LGTM

PR #2762, opened 2025-06-27
Server configuration and apply to every client

worksofliam:
It's been quite clear we need a way to set a default configuration for all users that connect to a specific system. This adds a new config file that will be read on startup, `/etc/vscode/codefori.json`, and will do two things:

* apply the server settings to the user's connection config
* disable those properties from being edited in the UI

image

You will also notice that not all properties were added to the settings schema. I didn't add anything that could come down to personal preference, like _source date gutter_.

### How to test

1. Connect to a system with no system config file
2. Ensure all connection setting fields are editable
3. Create `/etc/vscode/codefori.json` and test that the JSON schema is working as expected
4. Restart the connection and ensure the fields that were configured in the server settings are reflected in the connection settings UI

### To do

* [ ] Documentation

### DIFF
--- a/schemas/codefori.json
+++ b/schemas/codefori.json
@@ -7,7 +7,8 @@
         "tempLibrary": {
           "type": "string",
           "default": "ILEDITOR",
-          "description": "Temporary library. Cannot be QTEMP."
+          "description": "Temporary library. Cannot be QTEMP.",
+          "maxLength": 10
         },
         "tempDir": {
           "type": "string",
--- a/schemas/codefori.json
+++ b/schemas/codefori.json
@@ -1,0 +1,120 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "definitions": {
+    "CodeForI": {
+      "type": "object",
+      "properties": {
+        "tempLibrary": {
+          "type": "string",
+          "default": "ILEDITOR",
+          "description": "Temporary library. Cannot be QTEMP."
+        },
+        "tempDir": {
+          "type": "string",
+          "default": "/tmp",
+          "description": "Temporary directory."
+        },
+        "autoClearTempData": {
+          "type": "boolean",
+          "default": true,
+          "description": "Automatically delete temporary objects from the temporary library on startup."
+        },
+        "sourceFileCCSID": {
+          "type": "string",
+          "default": "*FILE",
+          "description": "The CCSID of source files on your system. You should only change this setting from `*FILE` if you have a source file that is 65535 - otherwise use `*FILE`. Note that this config is used to fetch all members. If you have any source files using 65535, you have bigger problems."
+        },
+        "enableSourceDates": {
+          "type": "boolean",
+          "default": false,
+          "description": "When enabled, Code for IBM i will retain/update source dates of members when editing source members. Requires SQL to be enabled for this feature to work. If this config is changed, you must reconnect to the system."
+        },
+        "encodingFor5250": {
+          "type": "string",
+          "default": "default",
+          "enum": [
+            "default",
+            "37",
+            "256",
+            "273",
+            "277",
+            "278",
+            "280",
+            "284",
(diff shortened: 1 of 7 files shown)


src/webviews/settings/index.ts
```diff
@@ -69,6 +68,15 @@ export class SettingsUI {
           .addCheckbox(`autoSaveBeforeAction`, `Auto Save for Actions`, `When current editor has unsaved changes, automatically save it before running an action.`, config.autoSaveBeforeAction)
           .addInput(`hideCompileErrors`, `Errors to ignore`, `A comma delimited list of errors to be hidden from the result of an Action in the EVFEVENT file. Useful for codes like <code>RNF5409</code>.`, { default: config.hideCompileErrors.join(`, `) })
 
+        if (connection?.getConfigFile(`connection`).getState().server === `ok`) {
```
chrjorgensen: @worksofliam This section of code does not work - the text and button never shows...

schemas/codefori.json
```diff
@@ -0,0 +1,120 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "definitions": {
+    "CodeForI": {
+      "type": "object",
+      "properties": {
+        "tempLibrary": {
```
chrjorgensen, line 7: @worksofliam Please add a length check for library - maybe a regex for library name validation, which I think we have in the code?

chrjorgensen (changes requested):
@worksofliam Some small issues found - great 1st attempt... 👍

src/api/configuration/serverFile.ts
```diff
@@ -0,0 +1,91 @@
...
+  private serverFile: string;
+  private serverData: T|undefined;
+
+  public validateData: ((loadedConfig: any) => T)|undefined;
+
+  constructor(private connection: IBMi, configId: string, readonly fallback: T) {
+    this.basename = configId + `.json`;
+    this.workspaceFile = path.join(WORKSPACE_ROOT, this.basename);
+    this.serverFile = path.posix.join(SERVER_ROOT, this.basename);
+  }
```
SanjulaGanepola, line 29: It looks like everything to do with the workspace is not actually really used for anything (`state.workspace`, `workspaceFile`). Is this intentional?

src/api/configuration/serverFile.ts
```diff
@@ -0,0 +1,91 @@
...
+}
+
+export class ConfigFile<T> {
+  private state: LoadResult = {server: `not_loaded`, workspace: `not_loaded`};
+  private basename: string;
+  private workspaceFile: string;
+  private serverFile: string;
+  private serverData: T|undefined;
+
+  public validateData: ((loadedConfig: any) => T)|undefined;
```
SanjulaGanepola, line 23: Is this something that will be implemented later?

src/api/configuration/serverFile.ts
```diff
@@ -0,0 +1,91 @@
...
+      } else {
+        this.serverData = serverConfig;
+      }
+    }
+  }
+
+  async get(): Promise<T> {
+    let resultingConfig = this.serverData;
+
+    if (this.validateData) {
```
SanjulaGanepola, line 70: Since there is no merging of any data at this point, wouldn't the data have already been validated in `loadFromServer`?

SanjulaGanepola (changes requested):
@worksofliam Works as expected! I left a couple of suggestions/questions

src/api/configuration/serverFile.ts
worksofliam: Cleaned up.

src/api/configuration/serverFile.ts
worksofliam: Yup, good spot. Previously there was code here that would validate the workspace config data.

SanjulaGanepola (approved):

PR #2766, opened 2025-06-27
Render sequence numbers in the editor

worksofliam:
Implement functionality to display sequence numbers, replacing source dates in the editor temporarily, allowing users to switch between the two.

**note:** sequence numbers are rendered one time only. Once the user makes a change to the document, it will switch the gutter back to the source dates. This PR does not change how sequencing works when the member is saved.

## How to test

1. Enable source dates and source gutter in the connection settings
2. Connect to a system and open a source member
3. See the source date gutter appear, and hover over the gutter to switch to source sequences
4. Make a change and see the gutter switch back to source dates
5. The option to switch back to sequences will dissapear after a) editing the file, and b) saving the file

### DIFF
--- a/src/filesystems/qsys/extendedContent.ts
+++ b/src/filesystems/qsys/extendedContent.ts
@@ -42,7 +42,7 @@ export class ExtendedIBMiContent {
       }
 
       let rows = await connection.runSQL(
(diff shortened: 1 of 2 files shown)


worksofliam:
image

worksofliam:
@sebjulliand back to you!

PR #2786, opened 2025-07-10
Fixed terminal settings string

### LINKED ISSUE #2783: Incorrect help text on system terminal settings
The system terminal settings list the example connection string as "ssl:localhost 992" which is missing the colon between the host and port.

Entering it like that causes a route to host not found error and fails to start the emilator.

|Context|Version|
|-|-|
|Code for IBM i version|2.16.3|
|Visual Studio Code version|1.101.2|
|Operating System|win32_x64|

buzzia2001:
### Changes
Fixed text terminal settings
Closes #2783 

**BEFORE**
image

**AFTER**
image


src/webviews/settings/index.ts
```diff
@@ -154,7 +154,7 @@ export class SettingsUI {
...
             .addCheckbox(`setDeviceNameFor5250`, `Set Device Name for 5250`, `When enabled, the user will be able to enter a device name before the terminal starts.`, config.setDeviceNameFor5250)
-            .addInput(`connectringStringFor5250`, `Connection string for 5250`, `Default is <code>+uninhibited localhost</code> (<code>+uninhibited</code> lets you get the cursor out of protected areas with any key).<br />A common SSL string is <code>ssl:localhost 992</code>`, { default: config.connectringStringFor5250 });
+            .addInput(`connectringStringFor5250`, `Connection string for 5250`, `The syntax for tn5250 is <code>[options] HOST[:PORT]</code>, default is <code>+uninhibited localhost</code><br />If your system is configured using telnet SSL, you need to specify <code>ssl</code> prefix before the connection string (examples <code>ssl:localhost</code> or <code>ssl:localhost:992</code>)<br/>You can also choose other opitions like <ul><li><code>+ruler</code>: Draw a ruler pointing to the cursor position</li><li><code>+uninhibited</code>: lets you get the cursor out of protected areas with any key</li></ul>`, { default: config.connectringStringFor5250 });
```
sebjulliand, line 157: The exact syntax is `[PROTOCOL:]HOST[:PORT]` instead of `HOST[:PORT]`; protocol can be `tn5250` (default), `ssl` or `debug`.

sebjulliand, line 157: There is a typo on `opitions`.

sebjulliand (changes requested):
Thanks for looking into this @buzzia2001 !
However, the description is not entirely accurate. `ssl` is not a prefix, it's a protocol, as [described here (check out the HOST option)](https://linux.die.net/man/5/tn5250rc).
See my remarks.
It may be a good idea to add a link to the documentation above too, to be thorough.

buzzia2001:
> Thanks for looking into this @buzzia2001 ! However, the description is not entirely accurate. `ssl` is not a prefix, it's a protocol, as [described here (check out the HOST option)](https://linux.die.net/man/5/tn5250rc). See my remarks. It may be a good idea to add a link to the documentation above too, to be thorough.

Hi @sebjulliand ,
I've fixed the typo.
As for the other part, everything else like the prefix, syntax, options etc. were taken directly from the help of the command launched from my system. Here you can see an output I don't know if it's my version that's old or what.
image

sebjulliand:
Fair enough. As someone seems already confused [here](https://github.com/codefori/vscode-ibmi/issues/2783#issuecomment-3061300777), here is what I advise:
- The description should say the connection string template is `[options] [ssl:]host[:port]`
- `options` is a whitespace separated list of option like (e.g. `+uninhibited +ruler`)
- `port` defaults to 23 by default or 992 is `ssl:` is specified
- default connection string is `+uninhibited localhost`

With all of this in the description, it should be quite clear 😄

PR #2791, opened 2025-07-15
Add support for custom SSH client on connect

SanjulaGanepola:
This PR adds support for a custom SSH client to be provided on connect.

This change is required for the `itest` CLI tool (https://github.com/IBM/vscode-ibmi-testing/pull/137) where I am pulling in the Code for IBM i API itself as a dependency. I need to be able to provide a custom client that has the same APIs as Node SSH, but instead simply runs everything directly using `spawn` as the CLI tool can run directly on the IBM i. The custom client will look something like this: [LocalSsh.ts](https://github.com/IBM/vscode-ibmi-testing/blob/feature/itest-cli/src/cli/src/localSsh.ts)

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -241,11 +241,11 @@ export default class IBMi {
       local: `#@$`
     };
   }
-  
+
   /**
    * @returns {Promise<{success: boolean, error?: any}>} Was succesful at connecting or not.
    */
-  async connect(connectionObject: ConnectionData, callbacks: ConnectionCallbacks, reconnecting?: boolean, reloadServerSettings: boolean = false): Promise<ConnectionResult> {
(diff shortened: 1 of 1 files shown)


worksofliam:
Before we merge this, I wonder if it's time to cleanup the `connect` method before we add more parameters.

@SanjulaGanepola @sebjulliand I am thinking we should change the parameter to accept to objects:

1. The connection details as it is today
2. Connection options, which would include the remaining parameters

Thoughts?

sebjulliand:
> Before we merge this, I wonder if it's time to cleanup the `connect` method before we add more parameters.
> 
> @SanjulaGanepola @sebjulliand I am thinking we should change the parameter to accept to objects:
> 
>     1. The connection details as it is today
> 
>     2. Connection options, which would include the remaining parameters
> 
> 
> Thoughts?

@worksofliam agreed. `conectionDetails` and `options` would be fine for the connect method (and allow more flexibility in the future).

worksofliam:
@SanjulaGanepola is this something you want to do as part of this pull request?

worksofliam:
@sebjulliand I made the change and the tests are looking good - it was simple and shouldn't change anything during runtime.

PR #2798, opened 2025-07-18
Changed debug service submission to have default library configuration

### LINKED ISSUE #2785: Add "CURLIB(QGPL)" to debug service startup parameter list
Problem description:
If the connection user profile is USR1 and the current library for the profile is set to his own library USR1, starting debug service from VSCode may fail. This is because the "SBMJOB USER(QDBGSRV)..." command to start the debug service changes the job user to QDBGSRV, but it still inherits the current library from the original connection user USR1. If the QDBGSRV profile does not have access to the USR1 library, the startup of the debug service will fail with the following message:

Error: Failed to submit Debug Service job: CPF1266:  User QDBGSRV not authorized to library USR1.
CPF1338:  Errors occurred on SBMJOB command.

The fix is to add the "CURLIB(QGPL)" parameter to the debug service submit options, so that the new job will use QGPL as the new current library. QDBGSRV should have access to this library.

buzzia2001:
### Changes
Changed **Debugger Service** submission command

Now, all parameters in the SBMJOB command are using default parameters: SYSLIBL(*SYSVAL) CURLIB(*USRPRF) INLLIBL(*JOBD)

image

Closes #2785

### DIFF
--- a/src/debug/server.ts
+++ b/src/debug/server.ts
@@ -63,7 +63,7 @@ export async function startService(connection: IBMi) {
       // Change the permissions to 777
       await connection.sendCommand({ command: `chmod 777 ${debugConfig.getRemoteServiceWorkspace()}` });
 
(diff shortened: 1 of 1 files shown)


sebjulliand (approved):
Thanks @buzzia2001 !
It works fine and should help with https://github.com/codefori/vscode-ibmi/issues/2785 👏🏻

Issue #71, opened 2021-03-03
CPF3064: Library ILEDITOR not found  - when adding new source file


SJLennon:
I'm on 0.2.3. The issue occurs after I add a new source file, then try to expand it.  
![image](https://user-images.githubusercontent.com/67484051/109842933-91140600-7c18-11eb-9d63-917648336a27.png)

If I exit VSCODE and restart it, the problem is fixed and I can expand the source file.

**_Note also_** that while trying to document this issue I have managed to add QPRGLESRC several times.  This occurred because I added it, but it didn't show up.  I then added it twice more without it showing up.  Finally, when I added QDDSSRC, the three QRPGLESRCs showed up.  Timing issue maybe...

I've been on various earlier releases and I understand that things are improving rapidly. Thus my settings.json file is somewhat messy which might be contributing to this problem.  There are a lot of lines in low intensity which I believe means they are not referenced and one of these is ""code-for-ibmi.temporaryLibrary" which was an earlier setting..  I see that now that library lists are connection specific (a good change!) there I now have several  "tempLibrary": "ILEDITOR" lines.  

Here's my settings.json file:

<details>
  <summary>Click to expand!</summary>

```
{
    "telemetry.enableTelemetry": false,
    "editor.renderControlCharacters": false,
    "editor.fontFamily": "Cascadia Code, Inconsolata, Lucida Sans",
    "editor.fontLigatures": false,
    "editor.renderWhitespace": "none",
    "editor.cursorStyle": "block",
    "editor.minimap.enabled": false,
    "terminal.integrated.shell.windows": "C:\\WINDOWS\\System32\\cmd.exe",
    "window.zoomLevel": 0,
    "update.mode": "start",
    "compile-hero.disable-compile-files-on-did-save-code": false,
    "compile-hero.generate-minified-html": true,
    "[javascript]": {
        "editor.defaultFormatter": "vscode.typescript-language-features"
    },
    "[html]": {
        "editor.defaultFormatter": "vscode.html-language-features"
    },
    "code-for-ibmi.connections": [
        {
            "host": "pub400.com",
            "port": 2222,
            "username": "lennons"
        },
        {
            "host": "ipublic.online",
            "port": 2222,
            "username": "lennons"
        }
    ],
    "code-for-ibmi.homeDirectory": "/home/lennons/qrpglesrc",
    "code-for-ibmi.libraryList": "QSYS2,QSYSINC, LENNONS1",
[...]

PR #2799, opened 2025-07-18
Disable all commands for change if system-wide readonly

chrjorgensen:
### Changes
This PR will disable all commands used for changing content on the server, if the server has system-wide readonly.
This include terminals and run of actions, but not library list change command etc.

When the server is defined as readonly system-wide, the connected bar item will look like this:

image

Note the icon is now a shield, the menu show "System-wide read only" (hyperlink) and no action maintenance or terminal options are available.

EDIT: C4i will now ask for reload when the readonly value changes in the config.

EDIT: The readonly field has been moved to the Features section as the first option:

image

EDIT: The readonly connection setting is decribed in the [PR #71](https://github.com/codefori/docs/pull/71) in the Docs repo.

EDIT: Save As will now fail with filesystem error if the connection is in readonly mode.

### How to test this PR
1. Open a connection marked as readonly via [system settings](https://codefori.github.io/docs/settings/system/)
2. Commands for changing server content is not available and is greyed out in the context menus
3. Remove the readonly attribute from the system settings and reconnect.
4. The commands for changing server content should now be available as usual.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/filesystems/qsys/QSysFs.ts
+++ b/src/filesystems/qsys/QSysFs.ts
@@ -195,6 +195,10 @@ export class QSysFS implements vscode.FileSystemProvider {
     async writeFile(uri: vscode.Uri, content: Uint8Array, options: { readonly create: boolean; readonly overwrite: boolean; }) {
         const connection = instance.getConnection();
         if (connection) {
+            const readonly = connection.getConfig().readOnlyMode;
+            if (readonly) {
+              throw new FileSystemError("Connection is in readonly mode");
+            }
             const contentApi = connection.getContent();
             let { asp, library, file, name: member, extension } = this.parseMemberPath(connection, uri.path);
             asp = asp || await connection.lookupLibraryIAsp(library);


worksofliam (changes requested):
Only one issue I found:

The deploy command still works even in readonly mode (should be a simple `package.json` addition)

Once that's done, I think we can merge.

Great work!

worksofliam:
@chrjorgensen I saw you mentioned you're on holiday - let me take over this PR for you.

sebjulliand:
Fair enough @chrjorgensen ! I can agree with that.

May I request a small change in the description of this field, when defining an action: 
Adding something like "does not apply to system wide read-only" here?
image

Just so that users don't get surprised when a system is completely protected.
Thanks!

chrjorgensen:
@sebjulliand I agree and think there should be a visible distinction between locally defined and system defined read only - I will work on that.

sebjulliand (approved):
Looks good to me!
Approved an merge; thanks @chrjorgensen !

PR #2804, opened 2025-07-22
Move EditorPath type to api directory

worksofliam:
Relocate the `EditorPath` type definition from the `typings` directory to the `api` directory for better organization and accessibility.

This is to correctly singularise the `api` directory so it doesn't depend on the `vscode` namespace.


worksofliam:
This is purely simple type change with no impact to our type API or extension. Therefore, I am ok to merge this.

PR #2805, opened 2025-07-22
Remove unused imports from serverFile.ts

worksofliam:
Eliminate unnecessary imports to clean up the code in serverFile.ts.

The import was breaking our API compatability in keeping the `api` directory away from the `vscode` namespace.


worksofliam:
Like before, just fixing bad types here.

PR #2836, opened 2025-08-11
Correctly reset internal storage prefix after disconnection

sebjulliand:
### Changes

This PR fixes the internal storage key prefix not being reset correctly after disconnecting from a system.

To reproduce the issue:
- Connect to an IBM i
- Disconnect
- Try reconnecting using the `Connect to previous IBM i` action form the Connection Browser
image

- Nothing will happen

That's because the internal storage prefix was set to `settings-` instead of blank after the disconnection happens.
Setting it back to blank fixes the issue.

### How to test this PR
1. Connect to an IBM i
2. Disconnect
3. Reconnect using the `Connect to previous IBM i` action from the Connection Browser

### Checklist

* [x] have tested my change


worksofliam (approved):
Tried it and works as expected. Thanks!!

PR #2876, opened 2025-09-07
Refactored storage code + fixed global storage + fixed quickconnect

sebjulliand:
### Changes

This PR fixes the Global storage (used to store the last connection and the cached server settings). It used to use the same `BaseStorage` object reference as the connection storage. That broke the global storage behavior since it had a storage prefix key assigned once connected (because connecting sets a prefix key on the connection storage...but since the global storage used the same BaseStorage as the connection, it lost its purpose).

Fixing the global storage fixes the "connect to last connected IBM i" action and the cached server settings access.

The `quickConnect` parameter was also not correctly tested, preventing quick connect from happening when connecting.

### How to test this PR
- From a local terminal: Run `npm run test` to run the test suites
- From a debug session
  - Connect to a system and reload settings
  - Disconnect
  - Check that the system you just connected to is marked as the last connected system (yellow icon)
  - Re-connect normally: it must connect faster than when reloading the settings

### Checklist

* [x] have tested my change


.gitignore
```diff
@@ -1,6 +1,4 @@
 node_modules
-.vscode-test/
 *.vsix
 dist
-.DS_Store
-.env
\ No newline at end of file
+.*
```
chrjorgensen, line 4: Why this change? Now changes to the file `.env.sample` in `src\api` won't be committed.
Isn't it better to have a positive list of files to ignore, instead of a global ignore, and let it be up to the committer to ensure that no secrets are included?

chrjorgensen (changes requested):
@sebjulliand I ran the suggested tests, and `npm run test` ran without issues, but when I run the extension in debug, I found that
- no connection is set as previous after disconnect, i.e. no `><` in yellow in the Connection browser
- the connection does not appear any faster on subsequent connection attempts, so Quick Connect does not seem to work still...

?

chrjorgensen:
Sorry for the delay... work, you know... 😉 

When I put a breakpoint in line 306 in `IBMi.ts`, I see `reloadServerSettings` is `undefined`:

image

So `quickConnect` will always return `false`?

.gitignore
sebjulliand: Excluding all dot files is a common practice; I thought it would be good to enforce it here instead of listing all the unwanted dot files or folder.
Since `.env.sample` has already been synchronized with Git, any change made to it will actually be committed. But I'll add an exception because, as you suggested, it should be moved elsewhere.

Indeed, moving `.env`/`.env.sample` elsewhere sounds like a good idea.

chrjorgensen (commented):
@sebjulliand After adding the log point, I see `QuickConnect is false` on every connection, even after disconnect and immediate reconnect. My connection has `Quick Connect` enabled...

Issue #2931, opened 2025-10-20
No specific warning when library list becomes invalid


luke-dykstra:
If a library in the library list no longer exists, programs will fail to compile because it fails to set the library list (it's now invalid). There is no warning or error message that the library list is now invalid because a library doesn't exist, programs will fail to compile with a vague 'Unable to open file' error message.

Whenever the library list (the one shown in VSCode) is used for an operation there should be a warning if the library list is now invalid.
It would be nice if it were a specific message saying which library no longer exists.

<hr />

|Context|Version|
|-|-|
|Code for IBM i version|2.17.2|
|Visual Studio Code version|1.105.1|
|Operating System|win32_x64|

<details>
<summary>Active extensions</summary>

```
COBOL (cobol): 25.10.11
Code for IBM i Walkthroughs (vscode-ibmi-walkthroughs): 0.5.0
Configuration Editing (configuration-editing): 1.0.0
Db2 for IBM i (vscode-db2i): 1.15.1
Emmet (emmet): 1.0.0
Error Lens (errorlens): 3.26.0
Git (git): 1.0.0
Git Base (git-base): 1.0.0
GitHub (github): 0.0.1
GitHub Authentication (github-authentication): 0.0.2
IBM i Debug (ibmidebug): 3.0.0
IBM i Project Explorer (vscode-ibmi-projectexplorer): 2.12.3
JSON Language Features (json-language-features): 1.0.0
Merge Conflict (merge-conflict): 1.0.0
NPM support for VS Code (npm): 1.0.1
Node Debug Auto-attach (debug-auto-launch): 1.0.0
RPGLE (vscode-rpgle): 0.33.2
Server Ready Action (debug-server-ready): 1.0.0
Source Orbit (vscode-sourceorbit): 1.4.1
TODO Highlight (vscode-todo-highlight): 1.0.5
Terminal Suggest for VS Code (terminal-suggest): 1.0.1
TypeScript and JavaScript Language Features (typescript-language-features): 1.0.0
```
</details>

<hr />

<details>
<summary>Remote system</summary>
</details>

<details>
<summary>Shell env</summary>
</details>

<details>
<summary>Variants</summary>

```json
{
  "american": "#@$",
  "local": "#@$",
  "qsysNameRegex": {}
}
```
</details>

PR #2933, opened 2025-10-21
Add error decoration for invalid libraries in the library list

SanjulaGanepola:
### Changes
* Add red error decoration for invalid libraries in the library list. This is just to make it more visible to the user (related to #2931).
    image
* Add `library` icon to libraries in the library list
* Automatically refresh the library list if a library in the library list is deleted from the Object Browser

### How to test this PR
1. Create a new library
2. Add it to the library list
3. Delete the library
4. Observe that the library list was refreshed and the deleted library has a red error decoration

### Checklist
* [x] have tested my change

### DIFF
--- a/src/ui/views/objectBrowser.ts
+++ b/src/ui/views/objectBrowser.ts
@@ -1082,7 +1082,6 @@ Do you want to replace it?`, item.name), { modal: true }, skipAllLabel, overwrit
       });
 
       if (newLibrary) {
-
         const filters = config.objectFilters;
 
         const createResult = await connection.runCommand({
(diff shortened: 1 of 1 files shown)


chrjorgensen (changes requested):
@SanjulaGanepola Great PR, learned some new stuff about decorators. 😃 

Almost there, see other comment.

src/ui/views/objectBrowser.ts
```diff
@@ -1090,7 +1090,15 @@ Do you want to replace it?`, item.name), { modal: true }, skipAllLabel, overwrit
           noLibList: true
         });
 
-        if (createResult.code !== 0) {
+
+        const isSuccess = createResult.code === 0;
+        if (isSuccess) {
+          const config = connection.getConfig();
+          const libraryList = [config.currentLibrary, ...config.libraryList].map(library => library.toUpperCase());
+          if (libraryList.includes(newLibrary)) {
```
chrjorgensen: This does not work if the library created is not entered in uppercase...

```suggestion
          if (libraryList.includes(newLibrary.toUpperCase())) {
```

chrjorgensen (changes requested):
One issue found - see comment.

And we should not ask for adding the new library to the library list, if it's already there... right?

sebjulliand (approved):
Works as advertised and code looks good.
It can be merged, as far as I'm concerned.
@chrjorgensen, do you want to have one final look?

PR #2938, opened 2025-10-23
Log the state of dependent IBM i components

SanjulaGanepola:
### Changes

This PR just adds some additional logging so we get a more readable summary of whether IBM i components we depend on are found.

image

### How to test this PR
1. Use `Connect and Reload Server Settings`
2. Check the `Code for IBM i` log

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -467,12 +467,6 @@ export default class IBMi {
           }
         }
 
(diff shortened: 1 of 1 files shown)


SanjulaGanepola:
@sebjulliand Yes, good idea! I opted to move this outside the quick connect check so this is always logged.

Issue #1451, opened 2023-07-21
Auto save profile when switching profiles


m-tyler:
When making adjustments to any active profile it's not intuitive to right-click on the profile name and then click on the Save Current Settings to Profile.  This is not a real problem, but I sometimes add new changes for development and then forget to save the changes later when I switch to another profile.

What I would like to be considered is a flag set in the connection that auto-saves when switching profiles.  If its easiest to always save then I am okay with that option.

I don't know if saving when the user exists VS Code is possible, but that would be a plus. 

Thanks, Matt

PR #2942, opened 2025-10-27
Actions/Custom Variables/Profiles overhaul

### LINKED ISSUE #2618: Flesh out the Profiles feature with some "simple" functionality
IMO one of the most powerful features in Code for i is Profiles. But I feel the feature doesn't get enough love. It's a bit confusing to get started with profiles, and when I started out I stumbled across the feature almost by accident. But I live by it now and would like to see some basic functionality added.

1. The Profiles section should appear after creating a server connection, with "default" listed right off. This would make it easier to get started with them. Currently, a command profile has to be created before the section is enabled.
2. Changes made to libl, object and ifs browsers should be applied to the active profile automatically, rather than having to remember to update the profile manually. (I forget often and must curse myself later.)
3. Simple list functionality (like those in object and ifs browsers): sort, reorder, and rename (profile).

#2 could be a setting option. Perhaps not everybody would like to have active profiles updated automatically.

**Additional context**
I don't believe this is required for the requested functionality, but would be happy to provide if needed.

You all rock and thank you for everything you've done and continue to do.

sebjulliand:
Fixes #2618
Fixes [#1451](https://github.com/codefori/vscode-ibmi/issues/1451)
## Changes
This PR introduce a new `Context` view as a replacement to the `Connection Profiles` view.
It is intended to enhance the user experience regarding the management of Actions, Custom Variables and Profiles.
image

It also introduces a new `CustomEditor` class that works like `CustomUI` but allows to open custom editors instead of custom panels.

### Actions
image

Actions are regrouped in three local categories, always displayed:
- Member actions
- Object actions
- Streamfile actions

Optionally, there will be additional categories for each workspace declaring local actions (in a local `.vscode/actions.json` file). These categories are named after the workspace they come from.

The actions available at the `Actions` node level are:
image
- Create a new action
- Setup local actions
- Search actions
  - Enables the user to search for actions by their name and command. Each matching action is highlighted in yellow and its description shows `search match`.
  image

- Go to next search result
  - Gives focus in the list on the next item that matched the search
- Clear search results

Creating a new action is also available on each `Action Type` node.
image

Clicking on an action opens the action editor. Making a change to the action will make the editor dirty and saving the action is done like in any editor, using `Save` (ctrl+s). The action panel was simply turned into an editor.
image

Right-clicking on an action gives access to these options:
image

Standard actions and local actions are managed the same way in the UI; they are saved in the connection settings or in the local workspace file depending on their nature.

When an editor is open, an inline action is available on each action to run it on the active editor:
image

Checking of the action can run on the active editor is done when the action is executed.

### Custom Variables
image

Each variable declared in the current connection settings are listed here and can be created/changed/deleted from this view.
image
image

Clicking on variable lets you change its value.

### Profiles
image

**Command profiles and Connection profiles have been merged - a process will run when the extension starts to create profiles out of existing Command profiles and show a dialog to warn the user the Command profiles have been migrated**.

Profiles can be created from the `Profiles` node using one of these inline actions:
image

- Create an empty profile
  - Asks for the name of the new profile then opens the profile editor.
- Create a new profile from the current settings
  - Asks for the name of the new profile then shows a notification to switch to it.
  image

Clicking on a profile opens the Profile editor that lets the user change some of the profile settings - object filters, IFS shortcuts and Custom variables are simply listed:
image
image

Right-clicking on a profile gives access to these options:
image

Activating a profile is done using the related inline action on the profile:
image

Activating a profile will replace the current settings with the profile's. Before the settings are replaced, they are saved back into the profile they belong to, or into the default profile if no profile was active (default profile is a unique profile with no name). This allows to automatically save the profile state before it gets erased.

The active profile is shown in green in the list, and it can't be deleted:
image

Its name is shown in the context description as well:
image

Switching back to the default profile is done using the action that appears on the Profiles node inline actions:
image

When a profile has a Library List Command (formerly a `Command Profile`), that command is ran and the library list is computed from it when the profile is activated. That operation can be re-run after the profile is made active by clicking the inline action on the profile item (the action only shows up if that profile has a library list command):
image

## How to test this PR

1. Create, edit, delete, rename, run actions
  a. Try generating local actions and make sure they are managed correctly in the local file
3. Create, edit, delete, rename variables
4. Create, edit, delete, rename, profiles
  a. Make sure custom variables are saved on the profile
  b. Make sure settings are correctly saved into the profile so they don't get lost when re-activating it
  c. Make sure the UI is consistent with the profile's content (views must be refreshed, etc)
5. Try actions described in the PR 😅

## Checklist

* [x] have tested my change

### DIFF
--- a/package.json
+++ b/package.json
@@ -1878,29 +1878,29 @@
 				{
 					"id": "environmentView",
 					"name": "Environment",
-					"when": "code-for-ibmi:connected && !(code-for-ibmi:profilesViewDisabled || code-for-ibmi:environmentViewDisabled)",
+					"when": "code-for-ibmi:connected && !code-for-ibmi:environmentViewDisabled",
 					"visibility": "collapsed"
 				},
 				{
 					"id": "libraryListView",
 					"name": "User Library List",
-					"when": "code-for-ibmi:connected && code-for-ibmi:libraryListDisabled !== true"
+					"when": "code-for-ibmi:connected && !code-for-ibmi:libraryListDisabled"
 				},
 				{
 					"id": "libraryListViewDisabled",
 					"name": "User Library List",
-					"when": "code-for-ibmi:connected && code-for-ibmi:libraryListDisabled == true",
+					"when": "code-for-ibmi:connected && code-for-ibmi:libraryListDisabled",
 					"visibility": "collapsed"
 				},
 				{
 					"id": "objectBrowser",
 					"name": "Object Browser",
-					"when": "code-for-ibmi:connected && code-for-ibmi:objectBrowserDisabled !== true"
+					"when": "code-for-ibmi:connected && !code-for-ibmi:objectBrowserDisabled"
 				},
 				{
 					"id": "ifsBrowser",
 					"name": "IFS Browser",
-					"when": "code-for-ibmi:connected && code-for-ibmi:ifsBrowserDisabled !== true"
+					"when": "code-for-ibmi:connected && !code-for-ibmi:ifsBrowserDisabled"
 				},
 				{
 					"id": "testingView",


src/ui/diagnostics.ts
```diff
@@ -78,7 +77,7 @@ export async function refreshDiagnosticsFromServer(instance: Instance, evfeventI
 
 export async function refreshDiagnosticsFromLocal(instance: Instance, evfeventInfo: EvfEventInfo) {
   if (evfeventInfo.workspace) {
-    const evfeventFiles = await getEvfeventFiles(evfeventInfo.workspace);
+    const evfeventFiles = await vscode.workspace.findFiles(new vscode.RelativePattern(evfeventInfo.workspace, `**/.evfevent/*`), null);
```
worksofliam: Why this change

chrjorgensen (commented):
@sebjulliand As you know, I'm currently traveling, so I haven't got the time for a thorough review but just played around a little.

Nice work, which will simplify the concept of profiles. 👍 

Two comments:
1. After creating a new profile and copying a command, the context view is not refreshed and I didn't see the new command until a manual refresh.
2. Could we have the active profile attached to the connection name in the status bar, like "System (profile)"?

package.json
```diff
@@ -1708,6 +1614,160 @@
...
+			},
+			{
+				"command": "code-for-ibmi.context.action.search.clear",
+				"enablement": "code-for-ibmi:connected && code-for-ibmi:hasActionSearched",
+				"title": "Clear search",
+				"category": "IBM i",
+				"icon": "$(search-stop)"
+			},
+			{
+				"command": "code-for-ibmi.context.action.create",
```
SanjulaGanepola: When using `Create action` on a workspace in the `Context` view, I get this error:
image

SanjulaGanepola:
Shall we also link these issues:
- https://github.com/codefori/vscode-ibmi/issues/1451
- https://github.com/codefori/vscode-ibmi/issues/2618

Looks like you have most of those requests covered here in this PR!

package.json
sebjulliand: And fixed 👌🏻

sebjulliand:
Back at you folks, I updated the icons and fixed the reported issues.
I also added additional checks so actions and profiles can't be renamed/deleted while being edited.

SanjulaGanepola:
Other ideas:
1. When we do a `Launch Workspace Actions Setup...`, can we auto refresh the view so it loads the new actions?
2. When we get the `Local actions are already defined for this workspace. Do you want to overwrite them?` message, can we get an option to `Append`?

src/ui/Tools.ts
```diff
@@ -199,7 +199,23 @@ export namespace VscodeTools {
     return tooltip;
   }
 
+  export function profileToToolTip(profile: ConnectionProfile) {
+    const tooltip = new MarkdownString(generateTooltipHtmlTable('', {
+      "Home directory": profile.homeDirectory,
```
SanjulaGanepola:
```suggestion
      "Home Directory": profile.homeDirectory,
```

SanjulaGanepola (approved):
Great work!!

PR #2966, opened 2025-11-11
Fixed CRTDUPOBJ and italian translation

buzzia2001:
### Changes
Fixed CRTDUPOBJ command when an object is duplicated

### How to test this PR
When now you duplicate a *FILE object, now all data will be duplicated. For instance, if you duplicate a SRCPF now members will have also source statements.
Also fixed a typo in the italian translation


sebjulliand (changes requested):
Good one @buzzia2001 , thank you!
The logic is OK, but see my comment about the code.

sebjulliand:
Thanks @buzzia2001 !

Issue #2974, opened 2025-11-18
Strikethough text in RPG compile output listing


heymchri:
RPG compile output listing puts all text after '<DEL' in a comment line in strikethrough.

Test program:
```
     FQAUOOPT   IF   E           K DISK
       //    <DEL
     C                   EVAL      *INLR = *Ox

```
(note the incorrect *Ox to cause the compilation to fail, to better show the issue).

Compile listing:

<img width="1099" height="458" alt="Image" src="https://github.com/user-attachments/assets/be4ebabb-a742-4059-8eef-2ef0cc675a04" />

<hr />

|Context|Version|
|-|-|
|Code for IBM i version|2.17.2|
|Visual Studio Code version|1.105.1|
|Operating System|darwin_arm64|

<details>
<summary>Active extensions</summary>

```
.NET Install Tool (vscode-dotnet-runtime): 2.3.7
Amazon Q (amazon-q-vscode): 1.103.0
Amazon Q for command line (codewhisperer-for-command-line-companion): 1.19.6
Bob Cozzi's RPG IV to RPG Free Conversion (rpgiv2free): 1.0.12
CL (vscode-clle): 1.1.7
COBOL (cobol): 25.10.27
Code for IBM i Walkthroughs (vscode-ibmi-walkthroughs): 0.5.0
Configuration Editing (configuration-editing): 1.0.0
Db2 for IBM i (vscode-db2i): 1.15.1
Debugger for Java (vscode-java-debug): 0.58.3
Dev Containers (remote-containers): 0.431.0
Emmet (emmet): 1.0.0
Error Lens (errorlens): 3.26.0
Excel Viewer (gc-excelviewer): 4.2.64
Extension Pack for Java (vscode-java-pack): 0.30.4
Git (git): 1.0.0
Git Base (git-base): 1.0.0
GitHub (github): 0.0.1
GitHub Authentication (github-authentication): 0.0.2
IBM i Debug (ibmidebug): 3.0.0
IBM i Project Explorer (vscode-ibmi-projectexplorer): 2.12.3
IntelliCode (vscodeintellicode): 1.3.2
JSON Language Features (json-language-features): 1.0.0
Language Support for Java(TM) by Red Hat (java): 1.47.0
Maven for Java (vscode-maven): 0.44.0
Merge Conflict (merge-conflict): 1.0.0
NPM support for VS Code (npm): 1.0.1
Node Debug Auto-attach (debug-auto-launch): 1.0.0
[...]

PR #2975, opened 2025-11-18
Escape HTML entities in action output

chrjorgensen:
This PR will fix issue #2974 by escaping HTML entities in output from actions / compiles.

### How to test this PR

1. Compile the RPG code from issue #2974.
Notice the strike-through in the output.
3. Switch to this PR.
4. Compile the RPG code from issue #2974 again -
the strike-through is no longer there.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/ui/Tools.ts
+++ b/src/ui/Tools.ts
@@ -104,11 +104,11 @@ export namespace VscodeTools {
 
   export function escapeHtml(html: string) {
     return html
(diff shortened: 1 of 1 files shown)


sebjulliand (changes requested):
Good one @chrjorgensen .
I would just request a small change, for performance's sake.

Thanks!

PR #2983, opened 2025-11-21
Correctly return and handle connection command status

sebjulliand:
### Changes

The `code-for-ibmi.connectTo` command did not always return the actual result of the connection process. It would return `true` even if the connection failed (i.e. trying to connect to an unreachable system would return `true`).

This PR fixes this issue and makes sure that the connection result is sent back to whichever process calls it.

### How to test this PR
Put a logpoint or a breakpoint in ConnectionBrowser.ts on line 68 and check the value of the `connected` field. It must reflect the actual connection process result.

Connecting to an unreachable system must return `false`.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/filesystems/qsys/FSUtils.ts
+++ b/src/filesystems/qsys/FSUtils.ts
@@ -1,8 +1,8 @@
 import path from "path";
 import vscode, { l10n } from "vscode";
-import { VscodeTools } from "../../ui/Tools";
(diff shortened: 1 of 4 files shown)


chrjorgensen (approved):
@sebjulliand LGTM and works as expected... now! 🎉 

Approved - merge and I will make a release v2.18.0.

PR #2993, opened 2025-12-04
Added return type for run build and compile tool

rapra41:


rapra41:
If a developer wants to programmatically use the output of the build command(vscode-ibmi-projectexplorer.projectExplorer.runBuild), there is no way to capture the result.

The command behaves as a “fire-and-forget” action, which is fine for normal terminal usage, but does not support advanced use cases where the calling code expects structured data.
Need some changes in this PR(https://github.com/IBM/vscode-ibmi-projectexplorer/pull/703) so also made changes here too.

SanjulaGanepola:
Closing as changes migrated to https://github.com/codefori/vscode-ibmi/pull/3187

PR #3023, opened 2025-12-23
Fix actions highlighting when there is no active editor

### LINKED ISSUE #3018: Action highlights remain even when no tabs are open.
I love that it's highlighting the available actions for the active editor, and it switches when I swap - but it does not clear them when I close the active editor and then have zero tabs open.

Image

sebjulliand:
### Changes

Fixes #3018

Closing the last active editor will not let actions highlighted anymore.

### How to test this PR

1. Open a member or IFS file on which at least on action can be run on
2. The action must be highlighted
3. Close the editor
4. No action must be highlighted anymore

### Checklist

* [x] have tested my change

### DIFF
--- a/src/ui/views/environment/actions.ts
+++ b/src/ui/views/environment/actions.ts
@@ -92,7 +92,7 @@ export class ActionsNode extends EnvironmentItem {
       };
     }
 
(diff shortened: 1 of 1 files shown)


worksofliam (approved):
The colours now clear correctly when a tab is closed and no tabs remain open. Fab!

PR #3041, opened 2026-01-20
Fixed IFS browser sort icon

sebjulliand:
### Changes

The sort icon in the IFS browser would always show the Descending order icon (🔽), regardless of the actual sorting order.
This PR fixes this to show the correct sort icon.

### How to test this PR

1. Sort an IFS folder by name
2. The icon must be 🔽
3. Select `Sort by -> Name` gain
4. The icon must be 🔼
5. The same behaviour must be observed when sorting by date

### Checklist

* [x] have tested my change


SanjulaGanepola (approved):
Good catch

PR #3055, opened 2026-02-02
Check and change password

### LINKED ISSUE #3049: Password expiry warning
When my password expires, VS Code gives me a l "Cannot connect to server. All attempts failed" or similar message. No indication that my password expired. Which I understand. But I'm wondering if with the new connection engine (M something) if you'll be able to prompt the VS CODE User for a new password and update the User Profile's password. 
I can't see off hand how to do that short of knowing a backup user profile and password for this situation. 
Perhaps Merlin or the new Bob editor have a method that we can incorporate into VS CODE.

sebjulliand:
## Changes

Resolves #3049

This PR adds support for checking password expiry date and for changing passwords.

### New component: Password Manager
The password manager is a new IBM i component that can:
- Return how many days are left before the current user's password is going to expire and the expiration date
- Change the current user's password using a temporary Java program. The Java program is compiled on the IBM i, in a temporary folder that gets deleted after it's done. The program contains the old and new password so they are not printed in the output when it's called. It uses the JTOpen jar shipped with the OS to perform the change.

### New Change password status bar action
image

image

A new `Change password` action located in the connection's status bar menu allows users to change their password the IBM i they's connected to.
The action opens a webview asking for the current password and the new password (twice, for confirmation).

This webview will also display errors occurring before submitting the form and after, if the password change program fails.

### Password expiry check
After a user connects, a password expiry check is performed.
The check is performed once for any new connection, and it's performed again one week later (in case the expiry policy changes over time).

After the check, if the password expires in less than two weeks, a modal notification shows up and offer to change the password.
image

Clicking on `Change password` will open the webview above.

## How to test this PR
1. Connect to an IBM i
2. Hover over the connection status bar and select "Change password"
3. Try to change your password

Password expiry - if you have a *SECADM profile to play with:
1. Create a dummy profile with PWDEXPITV(5)
2. Create a new connection and connect with this profile
3. The password expiry dialog must show up
4. Close the dialog
5. Re-connect
6. The dialog must not show up...until tomorrow

## Checklist

* [x] have tested my change

### DIFF
--- a/src/commands/password.ts
+++ b/src/commands/password.ts
@@ -168,7 +168,7 @@ async function updateNextPasswordCheck(instance: Instance) {
     const today = new Date();
     const daysLeft = (await connection?.getComponent<PasswordManager>(PasswordManager.ID)?.getPasswordExpiration(connection))?.daysLeft || 24;
     let whenNextDay = 14; //Next check in two weeks by default
-    if (daysLeft < 7) {
+    if (daysLeft <= 7) {
       //Less than a week left: check every day
       whenNextDay = 1;
     }


worksofliam (changes requested):
Can we use `QSYCHGPW` via SQL instead?

src/api/components/password.ts
```diff
@@ -14,12 +15,12 @@ export class PasswordManager implements IBMiComponent {
...
-    //Virtual component, always installed
+    //Volatile component - procedure is created and removed when invoked
     return "Installed";
   }
 
   update(_connection: IBMi, _installDirectory: string): ComponentState {
-    //Virtual component, always installed
+    //Volatile component - procedure is created and removed when invoked
     return "Installed";
   }
```
worksofliam: Isn't the purpose of components to install on startup and then flag it as installed instead of installing it on demand?

src/commands/password.ts
```diff
@@ -82,6 +116,60 @@ export function registerPasswordCommands(context: ExtensionContext, instance: In
...
+            else if (data.currentPassword === data.newPassword) {
+              error = l10n.t("New password must be different from the current one.")
+            }
+            else if (data.newPassword !== data.newPasswordConfirm) {
+              error = l10n.t("New password field and confirmation field don't match.")
+            }
+            else {
+              try {
+                await window.withProgress({ title: l10n.t("Changing password..."), location: ProgressLocation.Notification }, async () => await connection.getComponent<PasswordManager>(PasswordManager.ID)?.changePassword(connection, currentPassword, newPassword));
+                await setStoredPassword(context, connection.currentConnectionName, newPassword);
```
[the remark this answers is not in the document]
sebjulliand: Definitely, I overlooked that. Well spoted!

SanjulaGanepola (changes requested):

sebjulliand:
@SanjulaGanepola right, that's better as you suggested 😅 
Have a look and keep me posted. Thanks!

src/commands/password.ts
```diff
@@ -82,6 +102,86 @@ export function registerPasswordCommands(context: ExtensionContext, instance: In
...
   ]
+}
+
+async function updateNextPasswordCheck(instance: Instance) {
+  const connection = instance.getConnection();
+  if (connection) {
+    const today = new Date();
+    const daysLeft = (await connection?.getComponent<PasswordManager>(PasswordManager.ID)?.getPasswordExpiration(connection))?.daysLeft || 24;
+    let whenNextDay = 14; //Next check in two weeks by default
+    if (daysLeft < 7) {
```
SanjulaGanepola:
```suggestion
    if (daysLeft <= 7) {
```

With the current implementation, if the expiration days left is 7, we would end up in the `else if` block meaning `whenNextDay` would become `0`. This would result in constantly checking every time the user connects and disconnect that day which would also result in many warning pop ups for them to address it.

sebjulliand: Ugh...well spoted 😅

PR #3056, opened 2026-02-03
Removed `netstat_job_info` calls

sebjulliand:
### Changes

This PR removes the queries to `netstat_job_info` as it requires `*IOSYSCFG` special authority since 7.6.
It was used to retrieve the debug server and service job information. 

The queries are replaced by queries to `job_info`, using the service and server job names (as they are not supposed to change).

Finally, a direc call to `updateConnectedBar` has been removed as it turns out it's already called twice when the user connects (indirect calls).

### How to test this PR

1. Connect
2. Check the debugger status
3. Ensure debug service can't be started from Code for i anymore
4. Check that the service status is correctly updated once started/stopped from iNav and refreshed in Code for i

### Checklist

* [x] have tested my change

### DIFF
--- a/src/debug/server.ts
+++ b/src/debug/server.ts
@@ -86,14 +86,13 @@ export async function startService(connection: IBMi) {
           if (job) {
             let tries = 0;
             let debugServiceJob: string | undefined;
-            const debugPort = Number(config.debugPort);
-            const debugSepPort = Number(config.debugSepPort);
+            const debugPort = Number(debugConfig.getRemoteServiceSecuredPort());
             const checkJob = async (done: (started: boolean) => void) => {
               if (tries++ < 40) {
                 if (debugServiceJob) {
                   if ((await getDebugEngineJobs()).service) {
                     //Debug service job running Java is still alive
-                    if ((await Promise.all([checkPort(connection, debugPort), checkPort(connection, debugSepPort)])).every(Boolean)) {
+                    if (await checkPort(connection, debugPort)) {
                       window.showInformationMessage(l10n.t(`Debug service started.`));
                       refreshDebugSensitiveItems();
                       done(true);


src/debug/server.ts
```diff
@@ -145,30 +145,22 @@ export async function stopService(connection: IBMi) {
...
-      return {
-        name: String(rows[0].JOB_NAME),
-        ports: rows.map(row => Number(row.LOCAL_PORT)).sort()
-      } as DebugJob;
-    }
+export async function getDebugEngineJobs(): Promise<DebugJobs> {
+  const rows = await instance.getConnection()?.runSQL([
+    "select 'SERVER' as TYPE, JOB_NAME from table(qsys2.job_info(job_status_filter => '*ACTIVE', job_type_filter => '*BATCH', job_name_filter => 'QB5ROUTER'))",
+    "Union",
+    "select 'SERVICE' as TYPE, JOB_NAME from table(qsys2.job_info(job_status_filter => '*ACTIVE', job_type_filter => '*BATCH', job_name_filter => 'QDBGSRV'))"
```
SanjulaGanepola: When I tried this on one of our 7.4 IBM i, I got this for some reason when I wrapped in a try catch because it was failing.
```
[SQ20483] Named argument JOB_NAME_FILTER for routine JOB_INFO not valid for reason code 1., 4274K, -20483
```
When I checked the docs online, I do see `JOB_NAME_FILTER` still available on 7.4, but for some reason I don't have that parameter. Any ideas why?

Image

sebjulliand:
@mkwan01 @SanjulaGanepola since netstat_info can't be used anymore to check the debug service status, we need another way to check the actual status of the service (hence why it seems to start faster than before: we return ok when the submitted job is started which is very fast, but doesn't mean the Java process actually started and keeps on running).

So what would be the best way to ensure the service is running? Any idea is welcome.

mkwan01:
@sebjulliand Can Code for IBM i client retrieve a host job by filter? If it has this capability, you can search for the job "*/QDBGSRV/QP0ZSP??", where the job user is QDBGSRV, and the job name starts with "QP0ZSP", followed by two other arbitrary characters. If this job is found, then debug service is started. The underlying Java process normally takes 10-15 seconds to start up. If you do not find it at the beginning, you can repeat the search for up to 30 seconds.

mkwan01:
@sebjulliand I am OK if you want to implement a new status check solution for debug service in a different PR. If that is the case, we can approve this PR first, and give you more time to think about how to address the status check issue later.

sebjulliand:
Thank you very much @mkwan01 , this was insightful 😄 
I pushed a change to now look for the `QP0ZSP*` kind of job for the Debug Service.

Let me know how it goes; thanks! If it fails, I'll implement the debug log ckech instead.
We'll try to sort this out in this PR, if possible 😉

sebjulliand:
Thanks for the feedback @mkwan01 !
I made some changes and now the debug service start routine will:
- Check if the submitted job started
- If found, look for the `QP0ZSP*` job
- Once found, check if the debug and SEP ports are opened
- If the QP0ZSP job dies in the meantime, show an error and offer to open the debug service log file

 I could't test it through and through since I don't have access to an LPAR with an updated debug service. But I could check that the routine failed as expected.

I'll let you validate it once more on your LPAR. Thanks!

sebjulliand:
> What is the status of this PR (aka should I give this a try again)?

@SanjulaGanepola yep, if you can give it a try that would be great. I don't have an LPAR with a working service at my disposal right now 🥴

sebjulliand:
@SanjulaGanepola thanks a lot 🙏🏻

mkwan01 (approved):
Tested the latest change (dated 03/03) on a 7.6 machine. The debug service start/stop functions are working as expected and the view is also updated correctly.

PR #3064, opened 2026-02-06
Added support for SSH keys passphrase

### LINKED ISSUE #3062: Cannot set passphrase for encrypted SSH keys
### Discussed in https://github.com/orgs/codefori/discussions/3061

<sup>Originally posted by **huejoc** February  6, 2026</sup>
We are trying to connect to an IBM i from VS Code via CODE for i. 
We want to use keys with passphrases, but are unable to do so.
We receive the following message:

Not connected to XXXXXXXXXX: Cannot parse privateKey: Encrypted private OpenSSH key detected, but no passphrase given

Our question: Is it not possible to log in via SSH and passphrase?
</div>

sebjulliand:
### Changes

Fixes #3062 and #3061 

This PR adds a new Passphrase field in the Login settings to let users define a passphrase for their SSH key.
image

Passphrase modification is handled the same way as password: it remains unchanged unless a new one is provided.
image

Changing the private key path will clear the passphrase (unless one is also provided when the key path is changed).

### How to test this PR
1. Create an SSH key with a passphrase
2. Add the public key to your `~/.ssh/authorized_keys` file on an IBM i IFS
3. Define a new connection to this IBM i using that key and set the passphrase
4. Connect: the connection must succeed with the ssh key + passphrase

### Checklist

* [x] have tested my change

### DIFF
--- a/src/webviews/login/index.ts
+++ b/src/webviews/login/index.ts
@@ -53,7 +53,7 @@ export class Login {
         { id: `connect`, label: `Connect`, requiresValidation: true },
         { id: `saveExit`, label: `Save & Exit` }
       )
-      .loadPage<NewLoginSettings>(`IBM i Login`);
+      .loadPage<NewLoginSettings>(`IBM i Login`, checkLoginForm);
 
     if (page && page.data) {
       const data = page.data;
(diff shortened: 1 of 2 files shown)


SanjulaGanepola (approved):
Looks good and works well! Thanks

In case this was missed above:

> Shall we also link our oldest issue yet! -> https://github.com/codefori/vscode-ibmi/issues/472

PR #3069, opened 2026-02-11
Update connection.defaultUserLibraries by fetching the current library list from the server

dariocs:
### Changes
Update connection.defaultUserLibraries by fetching the current library list from the server
This PR fixes an issue where connection.defaultUserLibraries was not being updated after changing the user's JOBD profile, causing compilation failures.

### How to test this PR

Test Scenario

Steps to Test
1. Initial Setup - Connect  
Connect to your IBM i system using Code for IBM i extension
Use your personal user profile

2. Verify Current Job Description(JOBD)
Open a 5250 terminal session to the IBM i system
Run the command to display your user profile:
DSPUSRPRF USRPRF(MYUSER)
Note the current JOBD parameter value

3. Change the User's JOBD
Change your user profile to use a different JOBD with a different library list:
CHGUSRPRF USRPRF(MYUSER) JOBD(MYLIB/NEWJOBD)
This changes the default library list that will be used for new jobs

4. Test Compilation
In VS Code, compile any program using the Actions feature
The compilation should complete successfully.

❌ Without This Fix (Before PR)
The connection.defaultUserLibraries retained the old library list from the original JOBD
When attempting to remove libraries with the command:
liblist -d ${IBMi.escapeForShell(Tools.sanitizeObjNamesForPase(connection.defaultUserLibraries).join(` `))}
The command tried to remove libraries that no longer existed in the current job's library list
Compilation failed with library-related errors

Expected Behavior
✅ With This Fix (After PR)
The updateDefaultUserLibraries() function fetches the current library list before compilation
The command to remove libraries  in the current job's library list 
The compilation succeeds without errors

Technical Details
The issue occurred because:

defaultUserLibraries was set only during initial connection
Our users changed their JOBD because they work in different company (which changes the default library list), the cached value became stale
Compilation attempts to reset the library list by removing the default libraries before adding custom ones
Trying to remove non-existent libraries caused the compilation to fail
This PR ensures the library list is refreshed from the server before each compilation, keeping it synchronized with the current job's actual library list.

Additional Verification
You can also verify the fix by:

Checking the IBM i Output channel for the library list being fetched
Confirming that the compilation log shows the correct library list is being used
Testing with different JOBDs that have significantly different library lists

### Checklist

* [X] have tested my change
* [X] have created one or more test cases

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "code-for-ibmi",
-	"version": "3.0.10",
+	"version": "3.0.11-dev.0",
 	"lockfileVersion": 3,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "code-for-ibmi",
-			"version": "3.0.10",
+			"version": "3.0.11-dev.0",
 			"hasInstallScript": true,
 			"license": "MIT",
 			"dependencies": {
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
 	"icon": "icon.png",
 	"displayName": "Code for IBM i",
 	"description": "Maintain your RPGLE, CL, COBOL, C/CPP on IBM i right from Visual Studio Code.",
-	"version": "3.0.10",
+	"version": "3.0.11-dev.0",
 	"keywords": [
 		"ibmi",
 		"rpgle",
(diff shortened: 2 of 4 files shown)


sebjulliand:
Hi @dariocs, thanks for looking into this and submitting a fix.

Even though that will fix the issue for sure, reloading the current library list _every time_ a command runs (not just actions byt the way, but any command run by the extension) sounds a bit overkill.

How about reloading the current library list only when the user connects ?

PR #3071, opened 2026-02-12
Dispose SSH connection when disconnecting & fixed disconnection events handling

### LINKED ISSUE #2578: Opening connections then dropping vpn causes multiple pop ups
**Describe the bug**
When a connection drops, it prompts for reconnection, but after picking any options, it prompts all of the previous connections to reconnect as well. 

**To Reproduce**
1. Connect to a VPN (this might work with just wifi)
2. Connection to a system
3. Disconnect from system
4. Repeat steps 2 and 3 multiple times, ending with step 2
5. Drop the vpn (or wifi might work) 
6. Wait for the pop up that asks you to reconnect
7. Click any of the options
8. Now multiple pop ups appear for all the previous systems you connected to (including if you connected to the same system in step 4) 

**Expected behavior**
Only the last connection should pop up, so after picking an option in the pop up, there should be no pop ups after

**Environment (please complete the following information):**
 - Extension version: 2.15.3
 - IBM i OS version: all 

_Tried to upload a screen recording but it would not let me_

sebjulliand:
### Changes

Fixes #2578

This PR fixes:
- The SSH connection not being disposed after the user manually disconnects
- The handling of disconnection events:
  - When `end` happens, then a normal disconnection happened (i.e. the user manually disconnected)
  - When `error` happens it means the connection was dropped on the client or server side

The connection not being disposed was causing #2578.
A broken connection will now only show the reconnect dialog once, for the connection that actually dropped.

### How to test this PR

1. Reproduce the steps from https://github.com/codefori/vscode-ibmi/issues/2578#issue-2920216819
2. Shut down VPN, or Wifi, or pull you Ethernet cable
3. Wait for a timeout for happen
4. The reconnect dialog shoud show up just once, for the connection that just dropped

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/CompileTools.ts
+++ b/src/api/CompileTools.ts
@@ -25,6 +25,11 @@ export namespace CompileTools {
     updateProgress?: (message: string) => void
   }
 
+  export function reset(){
+    ileQueue.clear();
+    jobLogOrdinal = 0;
+  }
+
   /**
    * Execute a command
    */
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -1170,6 +1170,8 @@ export default class IBMi {
   }
 
   private async dispose() {
+    CompileTools.reset();
+
     //Clear connected resources
     if (this.sqlJob) {
       delete this.sqlJob;
--- a/src/api/queue.ts
+++ b/src/api/queue.ts
@@ -1,19 +1,14 @@
 export class SimpleQueue {
-	private static _instance: SimpleQueue | null = null;
-
-	static get instance() {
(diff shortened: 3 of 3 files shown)


sebjulliand:
> Side note: Can we pull in @types/ssh2 as a dev dependency?

I'm definitely adding it; it's been bugging me so far not having these types 😛

SanjulaGanepola (approved):
Tested and works great now!

sebjulliand:
Thank you @SanjulaGanepola ! Let's merge!

PR #3093, opened 2026-02-27
Keep editors opened when connection drops

### LINKED ISSUE #3092: Timeout loses all open tabs
As the title suggests, no one wants to be hunting down 10+ open RPG/CLE/CMD files (tabs) just because the system timed out when they went to lunch. If you have the files set to load read-only by default, it doesn't matter. Reload everything I had loaded up like every other editor. It should not matter that I didn't edit the files. 

**Expected behavior**
All my previously open tabs remain open, no matter what happens to the connection. 
It should also maintain the scroll position.

**Environment**
Extension version: 2.18.0
IBM i OS version: 7.5

**Work Around**
VSCode even knows what files were open, you can go to File->Open Recent->Reopen Closed Editor and it will reopen at least a short list of tabs. I don't know what kind of limit there is, I've not pushed it beyond 10. It's a pain, being a connection can timeout 4-5 times in a typical work day, but it seems to work most of the time. It's not great that you have to do 1 file at a time. 

**Additional Context**
You can read the bug report/feature request in #1192, which is marked as closed completed, but the original request "All my previously open tabs remain open" remains un-implemented. This is still a very valid feature request.

Give me a list of previously opened files in the left menu so I can reopen them as a fallback if it can't be done. Certainly not the preferred fix, but better than nothing. Maybe let me "pin" frequently edited files? Possibly a whole different feature request there as a fallback.

sebjulliand:
### Changes

Resolves #3092 

Opened members/streamfiles/objects used to be closed inconditionnaly.
This PR changes that behavior and let the editors opened until the user decides to reconnect or not.
If the user choses to reconnect, the editors will await on the connection process. Should the connection fail and the user decides to give up, then they'll be closed.

### How to test this PR

1. Open several members/streamfiles
2. Close the VPN or Wifi or pull the ethernet cable
3. Wait for the timeout to happen
4. Click on reconnect
5. While the reconnection happens, the editors must stay opened and not try to reload their content
6. Once the reconnection is done, the editors must behave as usual

### Checklist

* [x] have tested my change

### DIFF
--- a/src/filesystems/qsys/FSUtils.ts
+++ b/src/filesystems/qsys/FSUtils.ts
@@ -42,5 +42,12 @@ export function handleEditorsLeftOpened(context: vscode.ExtensionContext) {
  * @returns `true` if the user has chose to reconnect, `false` otherwise.
  */
 export async function waitOnReconnect() {
-  return restoreEditors ? await restoreEditors : false;
+  if(restoreEditors){
+    return await restoreEditors
+  }
+  else if(instance.connected){
+    return (await instance.connected).success;
+  }
+
+  return false;
 }
--- a/src/Instance.ts
+++ b/src/Instance.ts
@@ -28,6 +28,7 @@ export interface ConnectionOptions {
 
 export default class Instance {
   private connection: IBMi | undefined;
+  connected?: Promise<ConnectionResult>;
 
   private output = {
     channel: vscode.window.createOutputChannel(`Code for IBM i`),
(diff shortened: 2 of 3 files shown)


src/filesystems/qsys/FSUtils.ts
```diff
@@ -42,5 +42,12 @@ export function handleEditorsLeftOpened(context: vscode.ExtensionContext) {
  * @returns `true` if the user has chose to reconnect, `false` otherwise.
  */
 export async function waitOnReconnect() {
-  return restoreEditors ? await restoreEditors : false;
+  if(restoreEditors){
+    return await restoreEditors
+  }
+  else if(instance.connected){
+    return (await instance.connected).success;
```
SanjulaGanepola: I never really used promise variables like this much before. Cool stuff!

PR #3110, opened 2026-03-12
Added missing `postDownload` and `deployFirst` fields in action editor

sebjulliand:
### Changes

Following up on #3109 

This PR adds a Post Execution Downloads and a Deploy before running field to the Action editor for Local Actions only.
image

Other types of action will not display these fields in the editor.

These fields could only be added and edited manually from the settings/actions.json files.They can now also be changed from the editor.

The `postDownload` field also now supports variables replacement when the action is executed.

The PR also makes sure that some optional action fields are cleaned up before being saved to avoid writing blank strings/empty array in the settings.

### How to test this PR

1. Edit an action
2. Set one or more post execution download paths
3. Save
4. The paths must be saved as an array in the settings
5. Remove the paths
6. Save
7. The postDownload array must be removed from the settings

### Checklist

* [x] have tested my change

### DIFF
--- a/src/editors/actionEditor.ts
+++ b/src/editors/actionEditor.ts
@@ -149,7 +149,7 @@ async function save(targetAction: Action, actionData: ActionData, workspace?: vs
   Object.assign(targetAction, actionData,
     //Fields that needs to be tranformed before saving
     {
-      command: targetAction.command.replace(new RegExp(`\\\r`, `g`), ``), // We don't want \r (Windows line endings)
+      command: actionData.command.replace(new RegExp(`\\\r`, `g`), ``), // We don't want \r (Windows line endings)
       extensions: extensions.length ? extensions : undefined,
       outputToFile: actionData.outputToFile || undefined,
       postDownload: postDownload.length ? postDownload : undefined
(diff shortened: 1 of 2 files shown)


SanjulaGanepola (changes requested):
@sebjulliand The actual change in the PR is working as expected. But noticed some odd behavior when testing.

When editing local actions, for some reason the edits to the following are not being saved to the `actions.json` file:
- `Command(s) to run`
- `Environment`
- `Refresh`
- `Run on protected/read only`

Some of these don't seem to be editable on main either actually so likely unrelated to this PR. But `Command(s) to run` for example does work on main but not on this branch so semi-related 🤔

rishabkaushik:
@sebjulliand When will a new release be available with this change merged?

sebjulliand:
The next release will be the major 3.0.0 version, which is almost done.
There is another PR regarding security that's been asked by IBM, so it will have to be reviewed and merged first.

As for this one, I'll ask for another review.

SanjulaGanepola (approved):
Works as expected.

> But I couldn't get the other fields to fail to be updated.

I wasn't able to reproduce the issue any more so all is good.

sebjulliand:
Thanks @SanjulaGanepola

PR #3113, opened 2026-03-15
updated actions for rpg and rpgle member file extensions

supravi96:
### Changes
As of now code for ibm i returns rpgle actions for when user invokes get actions tool to fetch available actions for rpg member file. These 4 commands are only valid for .rpgle. This pr will update the code for ibmi to return 4 valid rpgle commands and 1 valid .rpg command . test cases are added for .rpg command as well.

### Checklist

* [x] have tested my change

### DIFF
--- a/package.json
+++ b/package.json
@@ -648,8 +648,7 @@
 						{
 							"type": "member",
 							"extensions": [
-								"RPGLE",
-								"RPG"
+								"RPGLE"
 							],
 							"name": "Create Bound RPG Program (CRTBNDRPG)",
 							"command": "CRTBNDRPG PGM(&OPENLIB/&OPENMBR) SRCFILE(&OPENLIB/&OPENSPF) OPTION(*EVENTF) DBGVIEW(*SOURCE) TGTRLS(*CURRENT)"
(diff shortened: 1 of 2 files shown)


sebjulliand (changes requested):
Sorry, this won't do.
- CRTPGM and CRTSRVPGM are meant to be `GLOBAL` - these are not commands related to RPGLE
- Why add support out-of-the-box for RPG36 and RPG38?

The changes on `CRTBNDRPG` and `CRTRPGMOD` are fine.

sebjulliand (approved):

hamsak2:
@sebjulliand CRTPGM and CRTSRVPGM can remain global, but they are not appropriate for .pf, .lf, .dspf and some more files. So, is it okay to display those commands to users when they select the Run Action option for .pf, .lf, and .dspf files?

below screenshot is when i take runaction for .pf file:
image

sebjulliand:
@hamsak2 sorry for the late reply.

> So, is it okay to display those commands to users when they select the Run Action option for .pf, .lf, and .dspf files

No, in my opinion these two commands should not be global and they should be restricted to objects and to the following types:
- RPGLE
- SQLRPGLE
- CBLE
- SQLCBLE
- CLLE
- CLE
- CPPLE

The idea would be to run these actions on modules to create a program or service program out of it.
If we want to keep them on the IFS files/members, then at least restrict the types as listed above.

PR #3123, opened 2026-03-20
Add extensions for action to action tooltip

chrjorgensen:
### Changes
Having multiple actions with the same name for different extensions makes it hard to find the right one in the list of action in the ENVIRONMENT view.

This PR adds the extensions to the action tooltip:
image

### Checklist
* [x] have tested my change


julesyan (approved):
I personally like it on the hover, its not something i am frequently checking

PR #3142, opened 2026-04-04
feat: DedicatedJob API for companion extensions

joelcamacho97:
### Changes

Adds a public `DedicatedJob` API that allows companion extensions to create persistent, isolated IBM i jobs through Mapepire.

#### Context

This addresses a recurring need discussed in:
- [discussions/2468](https://github.com/codefori/vscode-ibmi/discussions/2468) — Member Lock
- [discussions/3132](https://github.com/orgs/codefori/discussions/3132) — A way to lock member after opening

In both discussions, @sebjulliand identified the core limitation: _"ALCOBJ would not even be an option since commands executed from SSH are run in non-persistent jobs"_ ([#2468](https://github.com/orgs/codefori/discussions/2468#discussioncomment-11875855)) and _"Since the jobs running commands and queries are not persisted, members can't be locked. But when Mapepire will be integrated (starting with v3), this may become possible."_ ([#3132](https://github.com/orgs/codefori/discussions/3132)).

Mapepire is now integrated. This PR leverages it to provide exactly that capability — persistent jobs that companion extensions can use through the public API.

#### Why a generic DedicatedJob API instead of a built-in lock mechanism

Rather than implementing a specific lock policy inside Code for IBM i, this provides the underlying execution model that makes locking (and many other workflows) possible. A dedicated job persists for the lifetime of the connection, so:

- **Member locking** — `ALCOBJ`/`DLCOBJ` work because the lock lives in the job that holds it, and saves go through that same job
- **SEU-style validation programs** — open/close programs can run in the same job context
- **Job-scoped state** — library lists, temporary tables in QTEMP, open cursors, and other job-level resources are preserved across operations
- **Stateful workflows** — sequences of dependent CL commands that must share execution context

This means anyone who needs a custom locking or validation workflow can build a companion extension on top of this API, without requiring changes to Code for IBM i itself. The extension doesn't impose any policy — it only provides the persistent job capability that IBM i workflows depend on.

#### What changed

**New file:**
- `src/api/DedicatedJob.ts` — `DedicatedJob` interface (`runSQL`, `runCommand`, `close`, `jobId`) and `createDedicatedJob()` factory

**Modified files:**
- `src/api/IBMi.ts` — `createDedicatedJob()`, `setDedicatedSaveJob()`, `getDedicatedSaveJob()`, cleanup in `dispose()`
- `src/api/IBMiContent.ts` — `uploadMemberContent()` accepts optional `DedicatedJob` to route saves through the dedicated job
- `src/filesystems/qsys/extendedContent.ts` — `uploadMemberContentWithDates()` accepts optional `DedicatedJob`
- `src/filesystems/qsys/QSysFs.ts` — `writeFile()` passes the dedicated save job to content methods
- `src/typings.ts` — exports `DedicatedJob` type

**Design decisions:**
- No UI changes, no new settings, no schema changes — purely a programmatic API for extensions
- When no `DedicatedJob` is set, behavior is identical to current code (zero impact on existing users)
- All dedicated jobs are automatically cleaned up on disconnect
- The `runCommand()` method routes CL through the Mapepire SQL job (same pattern as `CompileTools.runCommand` for ILE), keeping everything in the same persistent job
- `setDedicatedSaveJob()` allows a companion extension to route member saves through its job, which is essential for lock-based workflows where the save must happen in the same job that holds the lock

#### API usage by companion extensions

```ts
const connection = instance.getConnection();

// Create an isolated persistent job
const job = await connection.createDedicatedJob();

// Run CL in the persistent job — the lock stays alive because the job persists
await job.runCommand('ALCOBJ OBJ((LIB/FILE *FILE *EXCLRD MBR1))');

// Route member saves through this job so saves respect the lock
connection.setDedicatedSaveJob(job);

// ... user edits and saves members normally in VS Code ...

// When done, release and clean up
await job.runCommand('DLCOBJ OBJ((LIB/FILE *FILE *EXCLRD MBR1))');
connection.setDedicatedSaveJob(undefined);
await job.close();
```

### How to test this PR

1. Build the extension (`npm run webpack-dev`)
2. Verify existing member save functionality works unchanged (no `DedicatedJob` set = same behavior as before)
3. For full testing with a companion extension, create a dedicated job via the API and verify:
   - `job.runSQL()` executes SQL in its own isolated job
   - `job.runCommand()` executes CL commands in the same job
   - `connection.setDedicatedSaveJob(job)` routes member saves through the dedicated job
   - `job.close()` properly ends the IBM i job
   - Disconnecting cleans up all dedicated jobs automatically

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/DedicatedJob.ts
+++ b/src/api/DedicatedJob.ts
@@ -1,0 +1,131 @@
+import { BindingValue } from "@ibm/mapepire-js";
+import type IBMi from "./IBMi";
+import { Mapepire } from "./components/mapepire";
+import { sshSqlJob } from "./components/mapepire/sqlJob";
+import { Tools } from "./Tools";
+import { CommandResult } from "./types";
+
+/**
+ * Public interface for companion extensions to interact with a dedicated SQL job.
+ * Created via `connection.createDedicatedJob()`, this provides an isolated SQL job
+ * that preserves job-scoped IBM i state (locks, library list, temporary tables, etc.)
+ * across multiple calls.
+ *
+ * The job is automatically cleaned up on disconnect, but should be explicitly closed
+ * when no longer needed to free IBM i resources.
+ *
+ * Companion extensions can also register a DedicatedJob as the "save job" via
+ * `connection.setDedicatedSaveJob(job)`, so that member saves go through the same
+ * job that holds locks or context.
+ */
+export interface DedicatedJob {
+  /** Execute one or more SQL statements on this dedicated job. */
+  runSQL(statements: string | string[], options?: { bindings?: BindingValue[] }): Promise<Tools.DB2Row[]>;
+  /** Run a CL command on this dedicated job and return a CommandResult. */
+  runCommand(command: string): Promise<CommandResult>;
+  /** Close the dedicated job and release its IBM i resources. */
+  close(): Promise<void>;
+  /** The IBM i job identifier (e.g. "123456/USER/JOBNAME"), if available. */
+  readonly jobId?: string;
+}
+
+/**
+ * Creates a new dedicated SQL job for use by companion extensions.
+ * Wraps a Mapepire sshSqlJob with a simple public interface.
+ */
+export async function createDedicatedJob(connection: IBMi): Promise<DedicatedJob> {
+  const mapepire = connection.getComponent<Mapepire>(Mapepire.ID);
+  if (!mapepire) {
+    throw new Error(`Mapepire component not available. Cannot create dedicated job.`);
+  }
+
+  const sqlJob = await mapepire.newJob(connection);
+
+  return {
+    get jobId() {
+      return sqlJob.id;
+    },
+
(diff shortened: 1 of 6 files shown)


sebjulliand:
Thank you @joelcamacho97 , but that PR is rejected for several reasons:
- The member lock logic will be implemented in this extension in the near future - but 3.0.0 needs to be released first and foremost. Member lock should not be delegated to other extensions in the first place.
- Mapepire is already exposed through the Component API, there is no need for a dedicated API.
- The logic around `DedicatedJob` has been copy/pasted from the existing logic which would double the maintainance efforts. 
- Some of the logic to clear the `DedicatedJob`s would come into conflict with the current `dispose()` logic.
- Saving/Loading processes should not be delegated to a different job or extension as long as the editors are opened from Code for i.
- Most if not all the PR content may have been generated by AI and not thoroughly reviewed.

PR #489, opened 2022-01-29
Support for EBCDIC streamfiles

worksofliam:
### Changes

New functionality to auto-convert from and to EBCDIC for streamfiles

* New connection setting to enable this feature

Screenshot 2022-01-29 at 12 45 07 PM
Screenshot 2022-01-29 at 12 45 53 PM

### Checklist

* [x] have tested my change
* [ ] updated relevant documentation
* [x] Remove any/all `console.log`s I added
* [x] eslint is not complaining
* [ ] have added myself to the contributors' list in the README
* [x] **for feature PRs**: PR only includes one feature enhancement.

PR #3147, opened 2026-04-08
fix: clear stale gutter decorations on toggle and nudge ruler redraw Fixes issue #489

bobcozzi:
Three issues caused the source date gutter to remain visible after being toggled off, and the RPGLE fixed-format ruler to misalign when the gutter was re-enabled on a different member:

1. Decorations never cleared on toggle-off _diffRefreshGutter only painted decorations inside the 'if (config.sourceDateGutter)' block. When the gutter was toggled off the block was skipped entirely, leaving stale 'before:' content on the editor. Added an else branch that explicitly calls setDecorations(annotationDecoration, []) and setDecorations(lineDecor, []) so the gutter is visually cleared immediately.

2. lineDecor leaked into sequence-numbers view The shouldShowSequences branch never cleared lineDecor (the line- highlight background) when switching from the date view. Added an explicit setDecorations(lineDecor, []) there as well.

3. RPGLE fixed-format ruler misalignment after toggle The RPGLE extension (halcyontechltd.vscode-rpgle) redraws its fixed-format column ruler only inside onDidChangeTextEditorSelection. Toggling our gutter does not fire that event, so the ruler stayed painted from its previous context and bled into or out of the gutter area. After calling _diffRefreshGutter in toggleSourceDateGutter we now reassign editor.selection = editor.selection, which VS Code treats as a selection change and fires the event, causing the RPGLE extension to redraw the ruler in the updated decoration context.

### Changes
Fixes a rendering issue where the source date gutter remained visible after being toggled off, and where toggling the gutter caused the RPGLE fixed-format column ruler to misalign.

### Checklist

* [ ] have tested my change

### DIFF
--- a/src/filesystems/qsys/sourceDateHandler.ts
+++ b/src/filesystems/qsys/sourceDateHandler.ts
@@ -235,6 +235,7 @@ export class SourceDateHandler {
           const activeEditor = vscode.window.activeTextEditor;
           if (activeEditor && activeEditor.document.uri.fsPath === document.uri.fsPath) {
             activeEditor.setDecorations(annotationDecoration, lineGutters);
+            activeEditor.setDecorations(lineDecor, []);
           }
 
         } else if (sourceDates) {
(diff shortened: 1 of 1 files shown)


bobcozzi:
Added a new commit that fixes a new issue with the change-date gutter. This time, when you leave VSCODE active, go over to another editor (like SEU or RDi) and change the same member that is left open in VSCODE, when you return to VSCODE the auto-reload kicks in and was NOT repainting the Change-Date gutter. So the newly inserted lines (if any) were inserted with no date gutter, thus misalignment occurs.  
That commit [661e611](https://github.com/codefori/vscode-ibmi/pull/3147/commits/661e611e059ac3f49431891bcf507af4499a7d4b) fixes that issue.

sebjulliand (approved):

PR #3151, opened 2026-04-09
Fix logger reading for debugger

julesyan:
### Changes
- Fix debugger start up to read the log file used
- Checks if debug service is started before trying to start it up
  - Will refresh the list if it does find it started. 
- If the startup job has ended, check the logs from start up instead of QPRINT

### How to test this PR
- for successful start up just start and stop the Debug Service
- for error messages, start up something on port 8001, then start up the Debug Service. It should fail with 
  ```
  Error: Debug Service failed to start:
  EQARD1006E The Remote Debug Service daemon failed to start because port 8001 is invalid or in use.
  ```

### Checklist

* [x] have tested my change

### DIFF
--- a/src/debug/server.ts
+++ b/src/debug/server.ts
@@ -22,6 +22,14 @@ export async function isDebugSupported(connection: IBMi) {
 }
 
 export async function startService(connection: IBMi) {
+  // Check if debug engine jobs are already running
+  const existingJobs = await getDebugEngineJobs();
+  if (existingJobs.service) {
+    window.showInformationMessage(l10n.t(`Debug service is already started.`));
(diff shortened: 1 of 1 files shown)


sebjulliand:
@julesyan I can (almost) guess where that comes from but can you explain why that change was required? Thanks!

julesyan:
Ive gotten feedback that the debugger startup does not provide ample feedback from the debug service if the job is not properly started. This PR changes how we fetch the errors from the debug service startup so any errors are bubbled up.

sebjulliand (approved):
Works fine. I can start the service as usual. Thanks!

PR #3152, opened 2026-04-09
Always clear mapepire spooled files if not generated by an action

sebjulliand:
### Changes

Initially reported by Ricky Thompson on Zulip.

Some commands run internally by Code for i may generate a spooled file that will not be deleted after the command has run.

For example, if source date support is enabled, since saving a source member is performed using `RUNSQLTM`, it will leave a spooled file:
image

This PR forces the spooled file generated from Mapepire to always be deleted, unless the command the spooled file must be kept (i.e. "Keep actions splooed file" is enabled in the settings)

#### Before the PR
1. Make sure source date support is enabled
2. Open a member
3. Save it
4. Open your spooled files list (either from WRKSPLF or using an extension)
5. There must be a `VSCODETEMP` spooled file

#### After the PR
Saving a member with source date support enabled will not leave a `VSCODETEMP` spooled file.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/CompileTools.ts
+++ b/src/api/CompileTools.ts
@@ -25,7 +25,7 @@ export namespace CompileTools {
     updateProgress?: (message: string) => void
   }
 
-  export function reset(){
+  export function reset() {
     ileQueue.clear();
     jobLogOrdinal = 0;
   }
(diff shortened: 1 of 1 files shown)


SanjulaGanepola (approved):
Tested and works!

sebjulliand:
Thanks @SanjulaGanepola !

bobcozzi:
Doesn't RUNSQLSTM have an OUTPUT paramaeter with several options, including *ERROR? I assume we generated the listing to check for errors, right? Maybe only check the error report instead. Those lengthy multi-10k+ source members produces some pretty lengthy output to deal with.

PR #3156, opened 2026-04-10
Fix write operations in QSYS file system where deleting entire contents

### LINKED ISSUE #3155: Unable to delete the contents of a member
When deleting the complete contents of a member, we get the following error:

Image

SanjulaGanepola:
### Changes
In `QsysFs#writeFile`, we always assumed that the content was empty only when doing a `Save as` [here](https://github.com/codefori/vscode-ibmi/blob/master/src/filesystems/qsys/QSysFs.ts#L207). However, this can also happen when the user is deleting the entire contents of a member. This PR now checks for `CPF5812` (example - `Member TEST already exists in file QCLSRC in library SANJULA`) when we do an `ADDPFM` and proceed as normal in this scenario.

Fixes #3155

### How to test this PR
1. Create a member and save some content into it.
2. Now wipe the entire contents and try to save.
3. Observe that you do not hit any error.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/filesystems/qsys/QSysFs.ts
+++ b/src/filesystems/qsys/QSysFs.ts
@@ -198,7 +198,7 @@ export class QSysFS implements vscode.FileSystemProvider {
         if (connection) {
             const readonly = connection.getConfig().readOnlyMode;
             if (readonly) {
-              throw new FileSystemError("Connection is in readonly mode");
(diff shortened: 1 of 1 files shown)


sebjulliand (approved):
Fixed as expected; good one!

PR #3160, opened 2026-04-15
Remove chsh check around updating a users default shell to bash

Nathaniel-King-Navarrete:
### Changes
Since pr #3131 allows us to fallback on SQL to set bash as default shell. We no longer need to check for chsh to be installed before throwing the `default_not_bash` connection error code

### Checklist

* [X] have tested my change


sebjulliand (approved):
Nice and simple, thanks!

PR #3163, opened 2026-04-16
Add job information to tooltip

julesyan:
### Changes
Add job information to the tooltip 

### How to test this PR
1. Connect to system
2. Hover over connection in bottom bar

### Checklist

* [x] have tested my change

### DIFF
--- a/src/instantiate.ts
+++ b/src/instantiate.ts
@@ -115,6 +115,9 @@ async function updateConnectedBar() {
     const actionsMenuItem = systemReadOnly ? `` : `[$(file-binary) Actions](command:code-for-ibmi.environment.actions.focus)`;
     const debugRunning = await isDebugEngineRunning();
     const connectedBarItemTooltips: String[] = systemReadOnly ? [`[System-wide read only](https://codefori.github.io/docs/settings/system/)`] : [];
+    const sqlJobId = connection.getSqlJobId();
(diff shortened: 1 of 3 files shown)


src/instantiate.ts
```diff
@@ -123,7 +126,8 @@ async function updateConnectedBar() {
       debugPTFInstalled(connection) ?
         `[$(${debugRunning ? "bug" : "debug"}) Debugger ${((await getDebugServiceDetails(connection)).version)} (${debugRunning ? "on" : "off"})](command:ibmiDebugBrowser.focus)`
         :
-        `[$(debug) No debug PTF](https://codefori.github.io/docs/developing/debug/#required-ptfs)`
+        `[$(debug) No debug PTF](https://codefori.github.io/docs/developing/debug/#required-ptfs)`,
+      sqlJobInfo
```
chrjorgensen, line 130: It would be nice if the job was clickable and copied to the clipboard - is that possible and easy to code?

sebjulliand: I would be even nicer to have both the option to open the joblog and copy the job id ! We could have these two on the same line.

chrjorgensen (approved):
@julesyan Looks good to me and works as expected.

One small comment above about getting the job info into the clipboard - but also approved. :smiley: 
So merge when appropriate...

PR #3171, opened 2026-04-24
Additional logs

julesyan:
### Changes
Added a timestamp to the logs and the option to not clear the output when it hits 150 lines. This should give us more information when debugging issues

### How to test this PR
1. Open Code for IBM i output
2. Connect to system

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -519,7 +519,7 @@ export default class IBMi {
         callbacks.message(`warning`, `IBM i ${this.systemVersion} is not supported. Code for IBM i only supports 7.3 and above. Some features may not work correctly.`);
       }
 
-      callbacks.progress({ message: `Checking Code for IBM i components.` });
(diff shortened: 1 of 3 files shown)


src/api/IBMi.ts
```diff
@@ -148,7 +148,18 @@ export default class IBMi {
   maximumArgsLength = 0;
 
   public appendOutput: (text: string) => void = (text) => {
-    process.stdout.write(text);
+    const now = new Date();
+    const timestamp = now.toLocaleString('en-US', {
```
sebjulliand: Maybe we should use a more international format rater than the US's. Moth/Day/Year format proves to be confusing.

Year/month/day would be better IMHO.

sebjulliand (changes requested):
Works well; I have one concern reported above.

sebjulliand (approved):

sebjulliand:
Thanks @julesyan !

PR #3174, opened 2026-04-25
Security fixes/enhancement

sebjulliand:
### Changes

As required by IBM, this PR aims at fixing potential security vulnerabilities.
- Components should now report a signature and read their remote signature so it can be compared. A mismatch in the signature means the component has been tampered with
- The PasswordManager component has been removed and replaced by a direct command call to the API
- The NewLibl procedure has been removed and replaced by direct SQL calls
- The GetMemberInfo component has been replaced by a simple query to `qsys2.PARTITION_STATISTICS`
- The checks for temporary library and temporary folder have been moved before they are used in the connection process.
- Some SQL calls have been changed to use table functions instead of views to speed up things.
- QTEMP is used whenever possible for temporary objects.

The only components that remains with this PR is Mapepire. The rest has been removed.

A `getSQLRoutineSignature` and `getSHA256FileHash` method have been added to `IBMiContent` to help other extension checks their components.

✔️ Unit test are OK (on remote test LPAR + LPARs with CCSID !== 37)
✔️ Unit tests from VS Code are OK

### How to test this PR
- Run the unit tests
- Connect to an IBM i
- List objects, members and open members
- Run a search and make sure results are consistents
- Use a profile with a Library List command; make sure the command actually changes the library list
- Use the `Go To File` command (ctrl+alt+o), use the * search to load libraries, files, members

### Checklist

* [x] have tested my change


src/api/IBMiContent.ts
```diff
@@ -1128,4 +1127,55 @@ export default class IBMiContent {
...
+   * @returns 
+   */
+  async getSQLRoutineSignature(library: string, name: string, type: "PROCEDURE" | "FUNCTION") {
+    return (await this.ibmi.runSQL(
+      /* sql */`select hash(ROUTINEDEF) SIGNATURE from qsys2.sysroutines where routine_type = '${type}' and rtnschema = '${library}' and RTNNAME = '${name}' fetch first row only`
+    )).at(0)?.SIGNATURE as string;
+  }
+
+  async getSHA256FileHash(remoteFile: string) {
+    if (this.ibmi.remoteFeatures.md5sum) {
```
julesyan: If md5sum is not installed, it will return empty, but it does not seem we allow an empty signature, is that correct? 
I am basing this off of the comparison done in manager.ts

If so then is md5sum now a requirement?

src/api/components/runtime.ts
```diff
@@ -24,42 +24,74 @@ export class IBMiComponentRuntime {
...
         await this.update(installDirectory);
       }
     }
     catch (error) {
       console.log(`Error occurred while checking component ${this.toString()}`);
       console.log(error);
 
-      this.state = `Error`;
+      this.state = { status: `Error`, remoteSignature: "" };
       this.setState(this.state);
```
julesyan, line 61: in my testing this check takes less than a second, so then can we not run this check each time the program is used?

src/api/components/getNewLibl.ts
```diff
@@ -1,61 +1,46 @@
...
-  private readonly procedureName = 'GETNEWLIBL';
-  private readonly currentVersion = 1;
-  private installedVersion = 0;
-
-  reset() {
-    this.installedVersion = 0;
-  }
+  static readonly ID = "GetNewLibl";
+  private static readonly VERSION = 2;
+  private static readonly SIGNATURE = "";
```
julesyan: This should also have a signature

sebjulliand: My bad, this one is not used anymore. I forgot to delete it 😅

src/api/components/runtime.ts
sebjulliand: Yes, we'll do that in the getComponent method I think. I just forgot that part 😅

src/api/IBMiContent.ts
sebjulliand: Since md5sum is part of the base open source package, it shouldn't happen.

And it's already used by the transfer operations.

julesyan: C4i should still work without the open source package no? I did not know that was a requirement

sebjulliand: Check the requirements 😆 
https://codefori.github.io/docs/install/#requirements

On the other, I see it says it requires 5733-SC1 but not the OSS package....which is wrong, it should mention that because of bash and all.

julesyan: Resolving this since we will be adding OSS package as part of the requirements

julesyan (approved):
Tested and working for me

PR #3175, opened 2026-04-27
Test with ssh key

buzzia2001:
### Changes
This PR allows testing using ssh private key instead of user's password

### How to test this PR

1. add the **VITE_PRIVATE_KEY_PATH** directive that specifies your private key path inside your **.env** file (see **.env.sample**)
2. run `npm run test`

### Checklist
* [X] have tested my change


sebjulliand (approved):
Great 👍🏻 
Thanks @buzzia2001 😊

PR #3176, opened 2026-04-27
Improve getMemberList performance using object_statistics and PARTITION_STATISTICS

bobcozzi:
## Summary

Replaces the `getMemberList` SQL implementation with a more efficient approach using `qsys2.object_statistics` and `qsys2.PARTITION_STATISTICS` instead of joining `QSYS2.SYSTABLES` and `QSYS2.SYSPARTITIONSTAT`.

## Changes

- `src/api/IBMiContent.ts`: Replace the CTE-based member list query with a `TABLE(object_statistics(...))` / `LATERAL PARTITION_STATISTICS(...)` query. Library and source file selection are passed directly into the `object_statistics` function call, eliminating a full cross-schema join.

## Performance

Testing shows significantly faster results on large source files (long member lists), while shorter lists perform comparably to the previous approach.

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -718,9 +718,7 @@ export default class IBMiContent {
         ${singleMemberExtension ? `${singleMember ? `AND` : `WHERE`} RTRIM(CAST(b.SOURCE_TYPE AS VARCHAR(10))) like '${singleMemberExtension}'` : ``}
         ORDER BY ${sort.order === 'name' ? 'NAME' : 'CHANGED'} ${!sort.ascending ? 'DESC' : 'ASC'}`;
 
-    const start = Date.now();
     const results = await this.ibmi.runSQL(statement);
-    console.log(`getMemberList ${library}/${sourceFile}: ${results.length} members in ${Date.now() - start}ms`);
     if (results.length) {
       const asp = this.ibmi.getIAspName(Number(results[0]?.ASP));
       return results.map(result => ({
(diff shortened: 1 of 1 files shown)


sebjulliand (approved):

sebjulliand:
Thanks @bobcozzi ; I'll incorporate it as it and fix the CCSID issue (table functions don't require the `sysNameInAmerican`) in #3174 as it should replace one the components there .

bobcozzi:
> Thanks @bobcozzi ; I'll incorporate it as it and fix the CCSID issue (table functions don't require the `sysNameInAmerican`) in #3174 as it should replace one the components there .

Thanks. I was doubtful that sysNameInAmerican needed was necessary, but left it in to match what was in there originally.

sebjulliand:
Nobody knows when it's actually necessary...damned CCSID 😅

bobcozzi:
Note that I'm creating a new PR for the getLibraryList function. Its use of the SYSTOOLS.SPLIT function has always bothered me since that SPLIT function is so inefficient and not on V7R2 (which is really the only thing stopping VS CODE from working on V7R2 (until we get to V3.0.0 of course). But the new method should be more efficient. https://github.com/codefori/vscode-ibmi/pull/3179

sebjulliand:
I don't know why we even bother with SPLIT here. We should just use a CTE, generate a list of values using `values` and use `OBJECT_STATISTICS` as it is now. That may be your plan maybe?

bobcozzi:
check it out I posted it. https://github.com/codefori/vscode-ibmi/pull/3179
But I use TS to create the list of values and then use VALUES(...) in the statement. 
I also want to review some of the CTE usage as I've seen where it creates a list of all objects, in the WITH... and then filters them out in one or two places--that's likely because IBM doesn't have a way to access object lists using the Object Attributes (only Object Type), but for database files, like Source Files, using SYSTABLES and one of the keyed files over Attribute, may help with file lists. But I'll look at that late this week.

sebjulliand:
> check it out I posted it. #3179 

Yep, great, that's what I had in mind. I'll merge it. Thanks Bob!

PR #3179, opened 2026-05-03
perf: optimize getLibraryList by replacing SYSTOOLS.SPLIT with VALUES clause

bobcozzi:
## Summary
This PR optimizes the `getLibraryList()` function by replacing `SYSTOOLS.SPLIT()` with a `TABLE(VALUES ...)` clause for better performance on older Power systems.

## Changes
- Replace `SYSTOOLS.SPLIT()` with `TABLE(VALUES ...)` in the SQL query
- Build VALUES clause dynamically from the libraries array

## Performance Benefit
`SYSTOOLS.SPLIT` is a SQL/PL function with overhead, while the VALUES clause uses native SQL. This provides better performance, especially on older/slower IBM i systems.

## Testing
Tested on older IBM i systems and verified:
- Library list retrieval works correctly
- Generic library names (e.g., `SQL*`) are supported
- Performance improvement observed on older Power servers

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -354,6 +354,9 @@ export default class IBMiContent {
   async getLibraryList(libraries: string[]): Promise<IBMiObject[]> {
     let objects: IBMiObject[];
     if (this.ibmi.enableSQL) {
+      // Build VALUES clause from the libraries array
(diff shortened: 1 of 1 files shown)


sebjulliand (approved):
Works good, thanks!

PR #3182, opened 2026-05-06
feat(addRowsParam): add rows parameter to runSQL method

Mohammed-Yaseen-Ali-2081:
## Changes

Added an optional `rows` parameter to the `runSQL` method to allow users to limit the number of rows returned from SQL queries, replacing the hardcoded value of 99999.

### Key Changes:
- Updated `runSQL` method signature to accept `rows?: number` in the options parameter
- Modified the query execution to use `options.rows ?? 99999`, maintaining backward compatibility
- Added comprehensive test case covering multiple scenarios

### Benefits:
- Improves performance when working with large result sets by allowing users to limit rows
- Maintains backward compatibility - existing code continues to work without changes
- Provides more control over memory usage and query performance

### API Usage:
```typescript
// Default behavior (fetches up to 99999 rows)
await connection.runSQL("SELECT * FROM myTable");

// Fetch only the first 100 rows
await connection.runSQL("SELECT * FROM myTable", { rows: 100 });

// With bindings and custom row limit
await connection.runSQL("SELECT * FROM myTable WHERE id = ?", { 
  bindings: [123], 
  rows: 50 
});
```

## How to test this PR

1. Run the test suite: `npm test`
2. Run the specific test case: `npm test -- -t "with rows parameter"`
3. Verify all test scenarios pass:
   - Limiting rows to 3 returns exactly 3 rows
   - Limiting rows to 1 returns exactly 1 row
   - Default behavior without `rows` parameter still works
   - Requesting more rows than available returns all available rows
   - Data integrity is maintained with the new parameter

### Manual Testing (optional):
1. Connect to an IBM i system
2. Execute SQL queries using the `runSQL` method with different `rows` values
3. Verify the returned row count matches the specified limit
4. Test with queries that have fewer rows than the limit to ensure graceful handling

## Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -1358,7 +1358,7 @@ export default class IBMi {
    * @param statements
    * @returns a Result set
    */
-  async runSQL(statements: string | string[], options: { bindings?: BindingValue[] } = {}): Promise<Tools.DB2Row[]> {
+  async runSQL(statements: string | string[], options: { bindings?: BindingValue[], rows?: number } = {}): Promise<Tools.DB2Row[]> {
     if (this.sqlJob) {
       let list = Array.isArray(statements) ? statements : statements.split(`;`).filter(x => x.trim().length > 0);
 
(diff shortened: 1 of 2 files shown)


sebjulliand (approved):
Works as expected.
THanks for the additional tests!

PR #3183, opened 2026-05-07
Changed tmp folder to be user's tmp folder. Also updated vsce to use latest version

julesyan:
### Changes
- changed /tmp to /home/USER/.vscode/tmp
- updated this.config to use the new getConfig()
- added tool for relative paths to be fully qualified
- updated vsce to 3.9.1 for security

### How to test this PR
- Download/Open member files

### Checklist

* [x] have tested my change


sebjulliand (changes requested):
It looks good!
I'd require two small things:
- Can you add a few test cases to test and cover use cases of `Tools.ensureFullPath` ?
- Since /tmp will automatically be changed to `.vscode/tmp` unconditionally, it would be nice to show a notification to warn the user about the change that happened in their config and/or document the fact that `/tmp` will not be permitted

Thanks! 🙏🏻

src/api/configuration/config/ConnectionManager.ts
```diff
@@ -1,8 +1,16 @@
...
+// Conditionally import vscode to support test environments
+let vscodeWindow: typeof import("vscode").window | undefined;
+try {
+  vscodeWindow = require("vscode").window;
+} catch {
+  // Running in test environment without vscode
+  vscodeWindow = undefined;
+}
+
 function initialize(parameters: Partial<ConnectionConfig>): ConnectionConfig {
```
sebjulliand: Ouch, that looks convulted just to display a warning.
How about moving the temporary folder check/migration into the `IBMi::connect method`?

It would have two benefits:
- Avoid that try/catch import; there is the `callbacks.message` that can be used there to display warnings in VS Code and tests alike
- The config is always saved after the `connect` method is called (https://github.com/codefori/vscode-ibmi/blob/a467304425a71361d77dfc2afc82c70352579cb7/src/api/IBMi.ts#L1012); so any change is recorded, not mater what.

Putting the check/change here would be a good fit since that method is dedicated to checking the temporary folder: https://github.com/codefori/vscode-ibmi/blob/a467304425a71361d77dfc2afc82c70352579cb7/src/api/IBMi.ts#L1073

sebjulliand (changes requested):
Tests are good, thanks!
I'd request one small change to make the whole thing cleaner.
Thanks @julesyan ! 🙏🏻

src/api/IBMi.ts
```diff
@@ -318,6 +318,12 @@ export default class IBMi {
       //Load existing config
       this.config = await IBMi.connectionManager.load(this.currentConnectionName);
 
+      // Migration: update old default tempDir (/tmp) to new default (.vscode/tmp)
+      if (this.config.tempDir === `/tmp`) {
+        this.config.tempDir = `.vscode/tmp`;
+        callbacks.message(`info`, `Temporary directory updated to /home/usr/.vscode/tmp`);
+      }
```
sebjulliand: Move this block at line 435, after the actual home directory is revolved and change the notification message to:
`Temporary directory updated to ${this.getConfig().homeDirectory}/.vscode/tmp`

The `/home/usr` may be confusing to some users.

sebjulliand (changes requested):
Almost there!
One small change and we'll be good to go. Thanks.

sebjulliand (approved):
Perfect. Approved and merged!

PR #3187, opened 2026-05-13
Add `ActionTools` to exported API and support simultaneous actions

SanjulaGanepola:
### Changes
- [x] Added `path` property to task definitions: The impact of this is that we can run multiple compiles at the same time. This change also means that a new task terminal is only reused for actions targeting the same target path.
- [x] Add `ActionTools` to exported API to support other extensions retrieving usable actions
- [x] Replace `boolean` return type for running actions with `ActionResult` to support more detailed results being accessible to other extensions starting actions

### How to test this PR
1. Run an action on a some file `A`
2. Run an action on a different file `B`
3. Observe that there is 2 task terminals for `A` and `B`
4. Re-compile `A` and observe that its task terminal was re-used

### Checklist
* [x] have tested my change


worksofliam:
@SanjulaGanepola can you also run the runtime tests and share the results here?

SanjulaGanepola:
@worksofliam I ran the tests (against this branch and `master`) and looks like there are already failing test cases due to the temp directory change. I opened an issue to track resolving those here: https://github.com/codefori/vscode-ibmi/issues/3197

worksofliam (approved):
urgent we get those broken tests fixed

PR #3209, opened 2026-05-25
Removed warning that is no longer relevant to debug

julesyan:
Talked to Morris, the warning for IPv4 is no longer needed. One thing to note is that the IP address needs to be added to the alternative name for the certificate.

### Checklist

* [x] have tested my change


sebjulliand (approved):
Great, thanks!

PR #3210, opened 2026-05-27
Added a fix to check whether the file exists or not. If it does not exist, the system will attempt to create it.

mounigopisetti1162:
### Changes

When writing content into a QSYS member, if the target member does not exist, the system currently prompts whether the member should be created and sometimes fails as member doesn't exist. With this fix, the system will first check if the member exists. If the member is not found, it will be automatically created, and the content writing process will continue without additional prompts.

### How to test this PR
* Write content into the specified QSYS member.
* Ensure that the member does **not already exist** before running the test.
* Use a **non-existent member name** for validation.
* Verify that the system checks for member existence.
* If the member is not present, confirm that it is **automatically created**.
* Validate that the content is successfully written after member creation.

### Checklist

* [x] have tested my change

fix for https://github.ibm.com/bob-for-ibm-i/vscode-ibmi-bob/issues/593 and https://github.ibm.com/bob-for-ibm-i/vscode-ibmi-bob/issues/608


sebjulliand:
> Write content into the specified QSYS member.

@mounigopisetti1162 how do you get into that scenario?

sebjulliand (changes requested):
Don't use `getMemberList` to check if a member exist, it is way too expensive for an operation involved in the saving process.

Use `IBMiContent::checkObject` or `IBMiContent::testStreamFile` with a /QSYS.LIB path. It will be faster and more appropriate.

SanjulaGanepola:
> it is way too expensive for an operation involved in the saving process.

I agree with this as well!

In fact, I think having an existence check in general even if we use `IBMiContent::checkObject` or `IBMiContent::testStreamFile` is still heavy as this will happen on every single save. I think we should only check once the first time.

@sebjulliand What do you think?

SanjulaGanepola:
Unless the `IBMiContent::checkObject` or `IBMiContent::testStreamFile` doesn't slow things down in which case maybe it is fine to add on every call?

sebjulliand:
> Unless the `IBMiContent::checkObject` or `IBMiContent::testStreamFile` doesn't slow things down in which case maybe it is fine to add on every call?

It would have to be benchmarked, to be honest. But my assumption would be that `testStreamFile` is the fastest since it runs straight in Bash through the SSH channel. `CHKOBJ` runs through Mapepire and this may induce a slight overhead.

supravi96:
Tried testStreamFile but it did not work for QSYS member paths, so went with CHKOBJ — that resolves the issue.

sebjulliand:
What QSYS path did you try?
To get this approved, we'd better get `testStreamFile` to work for it as `CHKOBJ` will be a bit expansive, especially if it's an edge case.

So with which scenario do you get in a state where you save a member that doesn't exist?

supravi96:
@sebjulliand the scenario I am trying is _write content to a non existing file_ , basically I 
provided file path which needs to be created first, also I was thinking since we are sort of caching using `savedAsMembers` to track member creation , we can do something on similar lines to track which members we checked for existence to avoid checkobj overhead.

sebjulliand:
@supravi96 OK, but how did you provide this path?
Was it from the Code for i UI or from another extension?

supravi96:
@sebjulliand we are able to replicate this scenario when I give prompt in Bob

sebjulliand:
@supravi96 OK so that's coming from Bob. You should get Bob to check for the member and create it if needed, instead. 

Alternatively, does the `create` option here is set to true when Bob triggers this method?
image

If it is, then maybe that would be better than running a test every time a file is going to be written.

supravi96:
@sebjulliand currently no values are being passed as part of options , using create will be a good option , but requires changes to pass correct values when called from Bob.

sebjulliand:
> @sebjulliand currently no values are being passed as part of options , using create will be a good option , but requires changes to pass correct values when called from Bob.

Ok, good. Sounds like the best option.
Have Bob changed to set `create` to `true` here.

Then change the code here to take that into account and create the member if `create` is true.

Let's mark this PR as draft in the meantime.

mounigopisetti1162:
One more thing to add: the TestStreamFile functionality is working as expected. So we currently have two options: either use the existing create option, or perform the existence check using TestStreamFile before creating the file/member. Which approach do you think would be better?

sebjulliand:
@mounigopisetti1162 let's go with the "create flag" option. It will be cleaner in the end.

It will also complete the implementation of that method, since it is supposed to honor the create and overwrite flags.

Thanks 👍🏻

mounigopisetti1162:
okay @sebjulliand will go with that approach

mounigopisetti1162:
@sebjulliand

Please find my observations regarding the addition of the **create flag**.

There are two scenarios when we ask Bob to write to a file:

1. **Bob calls the search tools to check whether the file exists.**

   * If the file exists, it directly calls the **writeFile** tool.
   * If the file does not exist, it first creates the file and then calls **writeFile**. In this scenario, we do not need a create flag because the file has already been created before the write operation.

2. **Bob does not call any search tools.**

   * In this case, we have no way of knowing whether the file exists or not.
   * Since the file's existence is unknown, we cannot reliably pass the create flag.

I am still thinking about whether there is any way to pass the **createFlag** without performing any validation checks.

@SanjulaGanepola, could you share your thoughts on this or suggest any alternative approaches that might still be possible?

sebjulliand:
@mounigopisetti1162 
> Bob does not call any search tools.
>
>   In this case, we have no way of knowing whether the file exists or not.
 >  Since the file's existence is unknown, we cannot reliably pass the create flag.

My guess in this case is it's Bob who must check wether or the member exist. It could use the `checkObject` API. I assume it has access to the Code for IBM i API since it's using the `member://` filesystem.

SanjulaGanepola:
Yes that is correct. Since there isn't really a way to actually pass in a create flag using the standard VS Code filesystem APIs, we will have to do this on our side in Bob. Closing this PR

PR #3216, opened 2026-05-28
Fix: Replaced sha256sum with openssl  for getting ifs file's hash

### LINKED ISSUE #3196: Getting a Mapepire does not match it's local signature on remote error message
I'm on version 3.0.0 (pre-release).  I did clear the .vscode folder but not sure if that's the folder it's wanting me to clear.

Image

buzzia2001:
### Changes
This patch replaces the use of sha256sum (a package that needs to be installed) with OpenSSL (which is already installed on the systems) for calculating the hash of ifs files

### Checklist

* [X] have tested my change

closes #3196

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -1188,24 +1188,14 @@ export default class IBMiContent {
 
   async getSHA256FileHash(remoteFile: string) {
 
-    //We use OPENSSL that is already build in inside os
-    /*if (this.ibmi.remoteFeatures.sha256sum) {
-      const sha256Sum = await this.ibmi.sendCommand({ command: `${this.ibmi.remoteFeatures.sha256sum} ${remoteFile}` });
-      if (sha256Sum.code === 0) {
-        return sha256Sum.stdout.split(' ')[0];
-      }
-      else {
-        throw new Error(`Call to sha256sum failed: ${sha256Sum.stderr || sha256Sum.stdout}`);
-      }
-    }*/
-   
+    //We use OPENSSL that is already build in inside os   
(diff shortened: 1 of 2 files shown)


src/api/IBMiContent.ts
```diff
@@ -1187,14 +1187,26 @@ export default class IBMiContent {
...
       }
       else {
         throw new Error(`Call to sha256sum failed: ${sha256Sum.stderr || sha256Sum.stdout}`);
       }
+    }*/
+   
+    if (this.ibmi.remoteFeatures.openssl) {
+      const sha256Sum = await this.ibmi.sendCommand({ command: `${this.ibmi.remoteFeatures.openssl} dgst -sha256 ${remoteFile} ` });
+      if (sha256Sum.code === 0) {
+        return sha256Sum.stdout.split(' ')[1];
```
sebjulliand: It will be fine until there's a space in the path being tested (unlikely but possible). I'd suggest to use `pop` since the hash is always the last element of the `split` here.
```suggestion
        return sha256Sum.stdout.split(' ').pop();
```

sebjulliand (changes requested):
It's a good change @buzzia2001 , thanks!
I'd ask for a small change though, to be on the safe side.

src/api/IBMiContent.ts
buzzia2001: Thanks for the suggestion; while I was at it, I went ahead and removed the old logic as well.

sebjulliand (approved):

sebjulliand:
Works great, thanks @buzzia2001 !

PR #3236, opened 2026-06-01
Added support to connect to Mapepire Server

### LINKED ISSUE #3239: VSCode connection to IBM i server
## Describe the bug

After upgrading to Code for IBM i 3.x, all users connecting to our IBM i development system become stuck at **"Starting Mapepire"** and the connection never completes.

We confirmed that the issue affects multiple users and workstations, so it does not appear to be user-profile specific.

During troubleshooting, we found that Code for IBM i starts Mapepire with:

```bash
QIBM_JAVA_STDIO_CONVERT=N
QIBM_PASE_DESCRIPTOR_STDIO=B
QIBM_USE_DESCRIPTOR_STDIO=Y
QIBM_MULTI_THREADED=Y

/QOpenSys/QIBM/ProdData/JavaVM/jdk11/64bit/bin/java \
-Dos400.stdio.convert=N \
-jar /home/<user>/.vscode/mapepire-server-2.3.5.jar --single
```

When Mapepire is started this way, VS Code remains indefinitely at **"Starting Mapepire"**.

We manually tested Mapepire outside VS Code and discovered that when:

```bash
QIBM_JAVA_STDIO_CONVERT=Y
-Dos400.stdio.convert=Y
```

are used instead, Mapepire starts normally and produces readable output.

When `QIBM_JAVA_STDIO_CONVERT=N` is used, the output appears garbled/corrupted and the startup process does not complete successfully.

This suggests a possible issue related to stdout conversion or single-mode communication between Code for IBM i and Mapepire.

## To Reproduce

1. Install Code for IBM i 3.0.7.
2. Connect to an IBM i V7R4M0 system.
3. Connection reaches "Starting Mapepire".
4. Connection never completes.
5. Output log shows:

```text
Starting Mapepire: QIBM_JAVA_STDIO_CONVERT=N QIBM_PASE_DESCRIPTOR_STDIO=B QIBM_USE_DESCRIPTOR_STDIO=Y QIBM_MULTI_THREADED=Y exec /QOpenSys/QIBM/ProdData/JavaVM/jdk11/64bit/bin/java -Dos400.stdio.convert=N -jar /home/CHAND/.vscode/mapepire-server-2.3.5.jar --single
```

6. No Mapepire connection is established.

## Expected behavior

Mapepire should start successfully and the IBM i connection should complete normally without hanging at "Starting Mapepire".

## Screenshots

Image

Image

Image

## Environment

* Extension version: 3.0.7
* IBM i OS version: V7R4M0
* Java: IBM Semeru Runtime Certified Edition 11.0.27
* Mapepire Server: 2.3.5

## Additional context

Additional observations:

* Manual startup of Mapepire in daemon mode works correctly.
* Port 8076 is successfully opened when started manually.
* The issue affects multiple users on the same IBM i partition.
* We verified that Java 11 is installed and functioning correctly.
* Changing the Mapepire Java Runtime setting from Java 8 to Java 11 did not resolve the issue.
* Disabling Quick Connect and Secure SQL Connection did not resolve the issue.
* Downgrading the extension also did not resolve the issue.
* The problem appears related to the use of `QIBM_JAVA_STDIO_CONVERT=N` and `-Dos400.stdio.convert=N`.

sebjulliand:
### Changes

This PR adds support for connecting to a Mapepire Server running on the remote IBM i instead of running it in single mode through SSH.

Three new settings have been added to the connection settings:
image

- `Connect to remote Mapepire server`: switch to connecting to Mapepire over HTTPS
- `Allow all certificates`: allows to connect using a self-signed certificate
- `Remote Mapepire Server port`: set the port to connect to. Defaults to `8076`.

Enabling `Connect to remote Mapepire server` will connect Code for i to Mapepire using an HTTPS connection and skip uploading Mapepire's jar file.

### How to test this PR

1. Start Mapepire as a server on IBM i
2. Change the connection settings to enable `Connect to Mapepire over HTTPS` (and `Allow all certificates`)
3. Connect
4. The connection must succeed (and be faster than usual)

### Checklist

* [x] have tested my change


buzzia2001:
Hi @sebjulliand ,
I’ll take a look in the next few days... but I’d like to discuss with you the feasibility of taking it a step further... Let me explain: why not consider using an existing Mapepire server if one is already running on the system, rather than setting up new ones? This would significantly reduce the initial setup time if Mapepire is already active. Consider, for example, that we’re currently using it on systems with development

sebjulliand:
@buzzia2001 I guess we're in agreement then; this is exactly the point of this PR: to connect to a Mapepire Server already running on the partition 😅 
It won't start a new process, it will connect to an existing one.

src/extension.ts
```diff
@@ -124,7 +125,13 @@ export async function activate(context: ExtensionContext): Promise<CodeForIBMi>
       commands.executeCommand("code-for-ibmi.environment.refresh");
     });
 
-  const mapepire = new Mapepire(path.join(context.extensionPath, `dist`));
+  const mapepire = new Mapepire(path.join(context.extensionPath, `dist`), async (connection) => {
+    return await getStoredPassword(context, connection.currentConnectionName) ||
+      await window.showInputBox({
+        password: true,
+        prompt: l10n.t(`Password for user profile {} on {} is required to connect to Mapepire Server.`, connection.currentUser, connection.currentConnectionName)
```
buzzia2001:
```suggestion
        prompt: l10n.t(`Password for user profile {0} on {1} is required to connect to Mapepire Server.`, connection.currentUser, connection.currentConnectionName)
```

buzzia2001 (changes requested):
It's great, even though saying “HTTPS” might be misleading, at least it was for me. I don't know...
Let's just say that if you already have the service running, you avoid having to upload files to individual home directories, and in terms of performance, you skip the JVM startup time, so that's a big deal.

src/extension.ts
buzzia2001: @sebjulliand I don't know how to resume a code review, but I'll replace some text descriptions in order to be more clear:

image

-> **Connect to Mapepire deamon**: When enabled, Code for IBM i connects to an already running Mapepire server (daemon) on the IBM i over HTTPS, instead of launching Mapepire through an SSH session.

image
-> When enabled, allows either self-signed certificates or certificates from a CA when connecting to Mapepire daemon over HTTPS.

image

-> **Mapepire deamon port** : The TCP port number Code for IBM will use when connecting to Mapepire deamon over HTTPS.

sebjulliand:
@buzzia2001 How about now?
image

buzzia2001 (approved):
@sebjulliand nice job, faster and faster

Image

sebjulliand:
Much obliged @buzzia2001 🙏🏻

Issue #3201, opened 2026-05-20
Searching source file does not work if filter contains lowercase letters and *


agatabrzeczek:
Hi, I'd like to report an issue with searching a source file. If your filter contains lowercase letters and regex in the member name, the feature won't work.

**To Reproduce**
1. Create a filter, make sure that the member name field includes lowercase letters and "\*" (e.g. "a\*" will do)
2. Expand the filter and search for any string in any object

**Expected behavior**
Results should appear. If there are no results, a popup informing about lack of results should appear.

**Screenshots**
<img width="273" height="56" alt="Image" src="https://github.com/user-attachments/assets/ef82d0a0-9e11-4c9d-b9de-f706159ecbbe" />

**Environment (please complete the following information):**
 - Extension version 2.18.0
 - IBM i OS version 7.5

**Additional context**
Here is the output:

```/QSYS.LIB/AGATA1.LIB/QRPGLESRC.FILE: LC_ALL=EN_US.UTF-8 /home/AGATA/.vscode/cqsh_1
/usr/bin/grep -inHR -F "hello" a*.MBR
{
    "code": 2,
    "signal": null,
    "stdout": "",
    "stderr": "grep: 001-2113 Error found getting information for object a*.MBR. No such path or directory."
}
```

The error does not occur if you use A* instead of a*.

Thanks for looking into it :)
Agata

SJLennon:
@agatabrzeczek 

If I understand this correctly, you have a filter that looks like this:
<img width="398" height="536" alt="Image" src="https://github.com/user-attachments/assets/b784881f-bf6e-442d-a7f8-305e4e56d2a6" />

As far as I know in the QSYS file system, all names are uppercase, so probably the code should be changed so that when creating a source file filter the "Member names filter" should be upper cased.

For now, a workaround is to enter uppercase characters in the "Member names filter" section of the filter.

agatabrzeczek:
Yes, that's exactly the issue I'm seeing.

And yes, as I mentioned, using A* would be a valid workaround.

PR #3238, opened 2026-06-01
Member name in filter mapped to uppercase

agatabrzeczek:
### Changes
This is a change caused by issue #3201. 

As @SJLennon suggested, the member name in the filter is now mapped to uppercase while creating the filter, to avoid the error described in the issue.

For instance, "a\*.MBR" is now mapped to "A\*.MBR"

### How to test this PR
Example:
1. Create a member filter. Make sure the member name contains lowercase letters and "\*" (e.g. "a\*")
2. Attempt to use the "Search Source File" functionality

### Checklist

* [x] have tested my change


src/webviews/filters/index.ts
```diff
@@ -83,6 +83,7 @@ export async function editFilter(filter?: ObjectFilters, copy = false) {
               .join(",");
             break;
           case `member`:
+            data[key] = String(data[key]).toUpperCase();
```
sebjulliand: Use the connection `uppercaseName` method here, since this is a name. `uppercaseName` will handle the special cases where a CCSID has lowercase variant characters that must not be uppercased (like CCSID 297 where one of the variant is `à`. Using this method ensures the name will remain valid in any case.
```suggestion
            data[key] = connection.uppercaseName(String(data[key]));
```

sebjulliand (changes requested):
Thanks @agatabrzeczek !
I will just ask for a small change, to avoid a CCSID related trap 😉

src/webviews/filters/index.ts
agatabrzeczek: Thanks for the suggestion @sebjulliand ! Tested OK from my side, change applied

sebjulliand (changes requested):
It looks good!
Now, since you're tackling this issue, could you make a similar change here, on this line?
 https://github.com/codefori/vscode-ibmi/blob/38f8b67423d2ed8aeeac23e12f1d9d456b7ea5f3/src/api/SearchTools.ts#L31

This way, your change will prevent filters from being lowercase and also make sure the search will work regardless of the member case (if that API is called from another extension, for example).

Thans in advance! 🙏🏻

agatabrzeczek:
As suggested, I tested and added the uppercasing in SearchTools.ts as well

Thanks!

sebjulliand (approved):
Perfect, approved!
Thanks @agatabrzeczek

PR #3253, opened 2026-06-03
Fix CCSID 65535 source files not readable with source dates enabled

bobcozzi:
## Problem
When source dates are enabled (`enableSourceDates: true`), opening source members with CCSID 65535 columns fails.

## Root Cause
The `downloadMemberContentWithDates` method uses SQL SELECT to read source members. The Mapepire SQL engine cannot read columns with CCSID 65535 (binary/no conversion) without an explicit CAST operation.

## Solution
This PR adds automatic detection and handling of CCSID 65535 columns:

1. **Query SYSCOLUMNS** - Gets the column-level CCSID and LENGTH for the SRCDTA column
2. **Conditional CAST** - When CCSID 65535 is detected, wraps SRCDTA in `CAST(srcdta AS VARCHAR(length) CCSID jobCcsid)`
3. **Improved LENGTH detection** - Uses the column definition from SYSCOLUMNS for SRCDTA column length

## Changes
- Modified `downloadMemberContentWithDates()` in `src/filesystems/qsys/extendedContent.ts`
- Added SYSCOLUMNS query to detect column CCSID and LENGTH
- Added conditional SQL CAST for CCSID 65535 columns
- Added single log statement when CCSID 65535 is detected

## Testing
Tested with:
- Source file with CCSID 65535 columns - now displays readable text ✅
- Normal source files with standard CCSID - continue to work normally ✅
- Upload/save operations - work without changes (INSERT doesn't need CAST) ✅

## Notes
- Only affects `downloadMemberContentWithDates` (when source dates enabled)
- Upload operations (`uploadMemberContentWithDates`) work without modification because SQL INSERT doesn't have the same CCSID read limitation
- Single file change, fully contained fix


bobcozzi:
Closing this PR in favor of #3254, which is rebased on the latest upstream/master. The upstream code was refactored to use `OVRDBF` instead of `CREATE ALIAS`, so the fix needed to be adapted to the new code structure.

PR #3254, opened 2026-06-03
Fix CCSID 65535 source files not readable with source dates enabled

### LINKED ISSUE #3228: Source members opened directly are not converting from EBCIDIC to ASCII.
I just upgraded to the latest version and opening a source member directly from the System i does not convert from EBCDIC to ASCII.  It shows up as a long series of Hex codes.  

Open a source member directly from an System i and you'll see what I mean.  It was working yesterday and with the upgrade, it no longer works.

bobcozzi:
## Problem
When source dates are enabled (`enableSourceDates: true`), opening a source member with CCSID 65535 does not work, even though it did on prior releases.

## Root Cause
- `downloadMemberContentWithDates` uses SQL SELECT to read source member lines
- The Mapepire SQL engine cannot read columns with CCSID 65535 (binary/no conversion) without an explicit CAST to a specific CCSID
- File-level CCSID ≠ column-level CCSID; must query `QSYS2.SYSCOLUMNS` for the actual column CCSID

## Solution
Query `QSYS2.SYSCOLUMNS` to check if the `SRCDTA` column has CCSID 65535. If so, wrap the column in a `CAST` to the job CCSID:

```sql
CAST(srcdta AS VARCHAR(length) CCSID jobCcsid)
```

## Changes
- **src/filesystems/qsys/extendedContent.ts**:
  - Added SYSCOLUMNS query to detect column-level CCSID and LENGTH
  - Apply conditional CAST when CCSID 65535 detected
  - Cache the column length for reuse in upload operations
  - Added console log for diagnostic purposes when CCSID 65535 is detected

## Testing
- Tested with CCSID 65535 source file: now displays readable text
- Tested with normal CCSID source files: behavior unchanged
- Upload operations continue to work (no CAST needed for INSERT)

## Notes
- This only affects `downloadMemberContentWithDates` (when source dates enabled)
- Standard download path (`downloadMemberContent`) is unaffected
- Upload operations work without CAST (SQL INSERT has different behavior than SELECT)
- Replaces PR #3253 (rebased on latest upstream/master)

Closes #3228

### DIFF
--- a/src/filesystems/qsys/extendedContent.ts
+++ b/src/filesystems/qsys/extendedContent.ts
@@ -30,9 +30,33 @@ export class ExtendedIBMiContent {
       const { library, file, name } = connection.parserMemberPath(uri.path);
       const overFile = await this.overDBFile(connection, library, file, name);
       try {
-        await this.readRecordLength(connection, alias, overFile);
+        // Query the column-specific CCSID and LENGTH for SRCDTA from SYSCOLUMNS
+        // Mapepire SQL cannot read columns with CCSID 65535, so we must cast to the job CCSID
+        const [columnInfo] = await connection.runSQL(
+          `SELECT CCSID, LENGTH FROM QSYS2.SYSCOLUMNS WHERE TABLE_SCHEMA = '${library}' AND TABLE_NAME = '${file}' AND COLUMN_NAME = 'SRCDTA'`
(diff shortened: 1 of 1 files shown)


buzzia2001 (approved):
Nice job @bobcozzi 
BobCozzi 1 - CCSID 0

PR #3273, opened 2026-06-10
Fix EVFEVENT errors mapping to server file instead of open local document

### LINKED ISSUE #3274: Clicking error in Problems panel opens server member instead of local file
## Description

When compiling a member-type action that uses `*EVENTF`, clicking an error in the Problems panel opens the source member from the server instead of the locally open file in the editor.

**This bug was found when compiling using a custom compile command. 
In our case we have a precompile step where we copy the source to a native QRPGLESRC file before compiling.**

## Steps to reproduce

1. Open a local rpgle source in the editor
2. Run a compile action that uses `*EVENTF`
3. Click an error in the Problems panel

**Expected:** The locally open file is focused at the error location.  
**Actual:** The source member is opened from the server.

## Root cause

`findExistingDocumentByName` — which searches all open editor tabs by filename — was inside the `if (evfeventInfo.workspace)` guard in `diagnostics.ts`. For member-type actions `workspace` is never set (only `file` actions populate it), so the lookup was always skipped and errors fell through to the server member URI.

## Fix

PR #3273 moves `findExistingDocumentByName` outside the workspace guard so it runs for member-type actions too.

neerup:
Fixes #3274

## Summary

When compiling a member-type action, clicking an error in the Problems
panel opened the source member on the server instead of the locally open
file.

Root cause: `findExistingDocumentByName` — which searches all open editor
tabs by filename — was inside the `if (evfeventInfo.workspace)` guard.
For member-type actions `workspace` is never set (only `file` actions
populate it), so the lookup was always skipped and errors fell through to
the server member URI.

Fix: move `findExistingDocumentByName` outside the workspace guard.
It needs no workspace context, so it now runs for member-type actions too,
resolving to the locally open file when one matches before falling back to
the server URI.

## Test plan

- [ ] Open a source member locally in the editor
- [ ] Run a compile action that uses `*EVENTF`
- [ ] Verify that clicking an error in the Problems panel opens the local file, not the server member
- [ ] Verify that workspace deploy-path mapping still works for `file`-type actions (errors in deployed IFS files still map to local workspace files)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -75,3 +75,4 @@ Thanks so much to everyone [who has contributed](https://github.com/codefori/vsc
 * [@MerelSoftware](https://github.com/MerelSoftware)
 * [@agatabrzeczek](https://github.com/agatabrzeczek)
 * [@phpdave](https://github.com/phpdave)
+* [@neerup](https://github.com/neerup)
--- a/src/ui/diagnostics.ts
+++ b/src/ui/diagnostics.ts
@@ -169,17 +169,19 @@ export function handleEvfeventLines(connection: IBMi, lines: string[], evfeventI
             continue;
           }
 
-          // If we get there, that means that even though we compiled from local, we likely had to use a temp member.
-          // We should try to find the file in the workspace. Since we can use findFile (it's async), then we look for open
-          // tabs like we do below.
-          if (evfeventInfo.extension) {
-            const baseName = file.split(`/`).pop();
-            const openFile = VscodeTools.findExistingDocumentByName(`${baseName}.${evfeventInfo.extension}`);
-            if (openFile) {
-              ileDiagnostics.set(openFile, diagnostics);
-              continue;
-            }
-          }
+        }
+      }
+
+      // For member paths: try to find an open local document by name before falling back to server URI.
+      // findExistingDocumentByName searches all open tabs — no workspace needed — so this works for
+      // member-type actions too (where evfeventInfo.workspace is not set).
+      if (!file.startsWith(`/`) && evfeventInfo.extension) {
+        const baseName = file.split(`/`).pop();
+        const lookupName = `${baseName}.${evfeventInfo.extension}`;
+        const openFile = VscodeTools.findExistingDocumentByName(lookupName);
+        if (openFile) {
+          ileDiagnostics.set(openFile, diagnostics);
+          continue;
         }
       }
 


sebjulliand:
@neerup I'm missing something here. Why should the `member` and `streamfile` actions should look for local files, since they are run on remote members /streamfile?

Workspace actions, run on local files (which are automatically uploaded before running the action), must look into the workspace, no problem. But for members and streamfiles, I don't see why they should 🤔

neerup:
> @neerup I'm missing something here. Why should the `member` and `streamfile` actions should look for local files, since they are run on remote members /streamfile?
> 
> Workspace actions, run on local files (which are automatically uploaded before running the action), must look into the workspace, no problem. But for members and streamfiles, I don't see why they should 🤔

Sorry that is Claude talking :) 
We do local development and we have our own custom compile command. When we compile and the source contains errors, clicking on the error in the werrorview opens the sourcefile on the ifs instead of the local source. In our case we do alot of precompiling before the actual ibm compile, our precompile is a little dated so we copy the ifs source file to QRPGLESRC where all the processing is done and the compile is also done on that source... even if we use the &SRCFILE to have the source copied to tmp source file, it still does not open the local file when clicking on the error in the errorview...
I'm aware of the ENVFEVENT file and the record:
FILEID     0 999 000000 023 QTEMP/QSQLPRE(EG0110RA) 20260610113611 0   
FILEID     0 001 000000 025 A4TTHN1/QRPGSRC(EG0110RA) 20260610113607 0 
FILEID     0 002 000112 026 EYAS4SRC/QRPGSRC(CS0600RA) 20240815154742 0
The fix works and just opens the local file as expected... 
Hope this clarifies...

sebjulliand:
Oh ok, so it is related to local actions - you run them from the workspace, not the IFS/Object browser, right? 😅

neerup:
> Oh ok, so it is related to local actions - you run them from the workspace, not the IFS/Object browser, right? 😅

Exactly...

sebjulliand (changes requested):
Nice fix, @neerup , thanks.
Works as expected.

We should also enhance that part so it can find and open these files even when closed...but that will take more time and efforts so let's roll with that.

Would you mind adding yourself to the contributors list in CONTRIBUTING.md before I approve and merge this? Thanks!

neerup:
Done... Added myself to CONTRIBUTING.md

sebjulliand (approved):
Perfect, thanks again!

PR #3246, opened 2026-06-02
Conditional WHERE clause if member type is *

buzzia2001:
### Changes
With this PR, it will be possible to view source file members without extensions in the object browser

### How to test this PR

1. Create a member without extension
2. Update object browser
3. Now you can see created member

image

image

**Member type must be * in the filter**
image

### Checklist

* [x] have tested my change
* [ ] have created one or more test cases
* [ ] updated relevant documentation
* [ ] Remove any/all `console.log`s I added
* [ ] have added myself to the contributors' list in [CONTRIBUTING.md](https://github.com/codefori/vscode-ibmi/blob/master/CONTRIBUTING.md)

Closes #3240

PR #3289, opened 2026-06-13
Fix SQL query to ensure non-empty member names are processed correctly

buzzia2001:
### Changes
With #3246 the where clause of the query used for members has been changed in order to allow member without extension. With this PR now the query returns only rows with member name.

### How to test this PR
1. Create an empty source file
2. Expand it, if it's empty now you won't see any row (before that you are able to see a . member)

### Checklist
* [x] have tested my change


sebjulliand (approved):
Work as expected. Nice fix 😄

PR #3307, opened 2026-06-18
Add CLPrompter extension integration for library list commands

bobcozzi:
When a connection profile's `setLibraryListCommand` starts with `?`, check if the [CLPrompter extension](https://marketplace.visualstudio.com/items?itemName=CozziResearch.clprompter) (`CozziResearch.clprompter`) is installed. If it is, use CLPrompter for advanced IBM i command prompting; otherwise fall back to the existing simple input box.


bobcozzi:
Did we get anywhere is this one?

PR #3310, opened 2026-06-19
Fix EVENTF retrieval and problems display when compiling from an iASP

### LINKED ISSUE #3241: TypeError crash in `runAction` when compile command uses non-standard parameters
Hi. I had the AI generate a summary of the problem for me. If it doesn't make sense at all, I apologize. In any case, the suggested workaround helped, but maybe the problem can also be fixed in the code.

# Bug Report: TypeError crash in `runAction` when compile command uses non-standard parameters

**Repository:** codefori/vscode-ibmi  
**Component:** `src/ui/actions.ts` — `runAction()`  
**Severity:** Error / Extension crash  
**Date:** 2026-06-02

---

## Summary

When a custom IBM i compile command (wrapper program) is executed via "Code for IBM i" Actions,
the extension crashes with a JavaScript `TypeError` at the end of the compile run — even when
the compile itself ran (with or without RPG errors). The extension output panel shows the full
compiler listing, then terminates with:

```
TypeError: Cannot read properties of undefined (reading 'library')
*  The terminal process terminated with exit code: 1.  *
```

---

## Environment

| Item | Value |
|---|---|
| Extension | Code for IBM i (codefori.vscode-ibmi) |
| Extension version | current release / pre-release (pre-release includes PR #2926, merged 2025-12-12) |
| VS Code | latest stable |
| IBM i OS | IBM i 7.5 (5770-WDS V7R5M0) |
| Source type | SQLRPGLE source member |

---

## Steps to Reproduce

1. Create an Action in "Code for IBM i" that calls a **custom compile wrapper program** instead of a standard ILE compile command. Example:

   ```
   CT0786CMD SRCLIB(&OPENLIB) SRCF(&OPENSPF) SRCMBR(&OPENMBR) TOLIB(&OPENLIB) WSID(*FIRST) BATCH(0) OPTION(*EVENTF)
   ```

2. Open an SQLRPGLE source member (e.g. `BRVZTDM/QRPGLESRC(AT0067)`).

3. Run the action (e.g. via right-click → "Run Action").

4. The compile command executes (compiler listing appears in the output terminal).

5. **Crash occurs** — the output ends with:

   ```
   TypeError: Cannot read properties of undefined (reading 'library')
   *  The terminal process terminated with exit code: 1.  *
   ```

---

## Expected Behavior

The extension should either:
- Display compiler diagnostics in the Problems panel (if `*EVENTF` output is available), **or**
- Silently skip the EVFEVENT fetch if the object location could not be determined from the command.

In no case should the extension crash with an unhandled `TypeError`.

---

## Actual Behavior

The extension crashes with:

```
TypeError: Cannot read properties of undefined (reading 'library')
```

No diagnostics are shown in the Problems panel. The terminal process exits with code 1.

---

### The two lookup functions

In `src/ui/actions.ts`, after a compile command runs, the extension tries to locate the EVFEVENT file using two functions:

```ts
const possibleObjects =
  getObjectsFromJoblog(commandResult.stderr) ||   // (1) looks for "EVFEVENT: LIB,OBJ" in joblog
  getObjectFromCommand(commandResult.command);    // (2) regex-matches standard CRT command params
```

**`getObjectFromCommand`** uses this regex:

```ts
const PARM_REGEX = /(^|\s+)(PNLGRP|OBJ|PGM|MODULE|FILE|MENU)\((?<object>.+?)\)/;
```

This only matches standard ILE compile parameters (`OBJ`, `PGM`, `MODULE`, `FILE`, `MENU`, `PNLGRP`).

**`getObjectsFromJoblog`** looks for lines starting with `EVFEVENT:` in stderr — only present if the wrapper explicitly writes such a message.

### Why the crash occurs with custom wrapper commands

The command `CT0786CMD SRCLIB(...) SRCF(...) SRCMBR(...) TOLIB(...) ...` uses **none of the parameter names** known to `PARM_REGEX`.
The wrapper does **not** write an `EVFEVENT:` joblog message either.

Result:
- `getObjectFromCommand()` → `undefined`
- `getObjectsFromJoblog()` → `undefined`
- `possibleObjects` → `undefined`
- `evfeventInfos` array stays **empty** (`[]`)

Later in the same function, the extension attempts to log which object it is fetching:

```ts
writeEmitter.fire(`Fetching errors for ` +
  (evfeventInfos.length > 1
    ? `multiple objects`
    : `${evfeventInfos[0].library}/${evfeventInfos[0].object}.`)  // ← CRASH
);
await refreshDiagnosticsFromServer(instance, evfeventInfos);      // ← called with []
```

`evfeventInfos[0]` is `undefined` → accessing `.library` throws `TypeError`.

### Why standard commands do NOT crash

A standard command like:

```
CRTSQLRPGI OBJ(&OPENLIB/&OPENMBR) SRCFILE(&OPENLIB/&OPENSPF) OPTION(*EVENTF) ...
```

contains `OBJ(LIB/MEMBER)` which matches `PARM_REGEX` → `possibleObjects` is populated →
`evfeventInfos` is filled → no crash.

---

## Downstream Impact

`refreshDiagnosticsFromServer` in `src/ui/diagnostics.ts` iterates over `evfeventInfos` to call:

```ts
evfeventInfo.forEach(async e => {
  const tableData = await content.getTable(e.library, `EVFEVENT`, e.object);
  // ...
});
```

When `evfeventInfos` is empty, no diagnostics are fetched. But the crash happens **before**
`refreshDiagnosticsFromServer` is even called — at the `writeEmitter.fire(...)` line.

---

## Proposed Fix

In `src/ui/actions.ts`, guard the log message and the `refreshDiagnosticsFromServer` call against
an empty `evfeventInfos` array. Use the pre-populated `evfeventInfo` (singular) variable as a
fallback — it already holds the source member's library and object name:

```ts
// Before (crashes when evfeventInfos is empty):
writeEmitter.fire(`Fetching errors for ` +
  (evfeventInfos.length > 1
    ? `multiple objects`
    : `${evfeventInfos[0].library}/${evfeventInfos[0].object}.`)
);
await refreshDiagnosticsFromServer(instance, evfeventInfos);

// After (safe):
const infoForLog = evfeventInfos.length > 0 ? evfeventInfos[0] : evfeventInfo;
writeEmitter.fire(`Fetching errors for ` +
  (evfeventInfos.length > 1
    ? `multiple objects`
    : `${infoForLog.library}/${infoForLog.object}.`)
);
await refreshDiagnosticsFromServer(
  instance,
  evfeventInfos.length > 0 ? evfeventInfos : [evfeventInfo]
);
```

`evfeventInfo` (singular) is the `EvfEventInfo` object built from the source member context
(library, object, extension, workspace folder) before the command runs. It always has valid
`library` and `object` values — exactly what `refreshDiagnosticsFromServer` needs to call
`content.getTable(e.library, 'EVFEVENT', e.object)`.

---

## Workaround (without code change)

Install the pre-release version of the extension (contains PR #2926) **and** add the following
line to the Action definition **before** the custom compile command:

```
RUNSQL 'CALL SYSTOOLS.LPRINTF(''EVFEVENT: &OPENLIB,&OPENMBR'')' COMMIT(*NONE)
```

This writes an `EVFEVENT: LIB,MBR` message to the joblog. The `getObjectsFromJoblog()` function
(introduced in PR #2926) will find it, populate `evfeventInfos`, and the crash will not occur.
Diagnostics will also be displayed correctly.

> **Note:** This workaround requires the IBM i partition to support `SYSTOOLS.LPRINTF`.
> On some partitions this stored procedure may behave differently (see discussion in PR #2926).

---

## Related Issues / PRs

| Reference | Title | Relation |
|---|---|---|
| [PR #2926](https://github.com/codefori/vscode-ibmi/pull/2926) | Listen for EVFEVENT signals from server | Adds `getObjectsFromJoblog()`; merged 2025-12-12; does **not** fix the TypeError crash itself |
| [PR #3096](https://github.com/codefori/vscode-ibmi/pull/3096) | Fixed EVFEVENT target retrieval from command line | Fixes `SRCFILE` being matched before `OBJ` in the regex; merged 2026-03-10 |
| [Issue #3208](https://github.com/codefori/vscode-ibmi/issues/3208) | (related) | Wrong EVFEVENT member name when using a wrapper `CALL PGM(...)` — same root domain, different failure mode (wrong name fetched vs. TypeError crash) |

### LINKED ISSUE #3304: EVFEVENT fetching fails in 3.0.10 with custom compile wrapper, works in 3.0.7
**Describe the bug**

EVFEVENT diagnostics fetching fails with Code for IBM i 3.0.10 when using a custom compile wrapper program.

The same compile setup works correctly with Code for IBM i 3.0.7.

This seems related to #3208 and #3241, but it appears to be a different failure mode:

* #3208 seems related to wrong EVFEVENT target/member detection when using compile wrapper programs.
* #3241 seems related to a `runAction` TypeError when compile commands use non-standard parameters.
* In this case, Code for IBM i appears to identify the correct target library/member, but the actual EVFEVENT fetching still fails in 3.0.10.

The issue does not appear to be an authority problem. EVFEVENT is created successfully, the member exists, and authorities were checked.

**To Reproduce**

1. Connect to an IBM i system using Code for IBM i 3.0.10.
2. Run an Action that uses a custom compile wrapper program.
3. The wrapper command is similar to:

```text
WOBJCRTRDI OBJ(TESTPGM) OBJATR(RPGLE) CRTOPT('i d') SRCFILE(QRPGLESRC) SRCLIB(TESTLIB) OBJLIB(TESTLIB) JOBD(TESTJOBD) OPTION(*EVENTF) SRCMBR(TESTPGM)
```

4. The wrapper internally runs a compile command similar to:

```text
CRTRPGMOD MODULE(QTEMP/TESTPGM)
          SRCFILE(TESTLIB/QRPGLESRC)
          DBGVIEW(*ALL)
          OPTION(*EVENTF *SRCSTMT *NOXREF *NOEXPDDS *NOUNREF *NOSHOWCPY *NOEXT *NODEBUGIO)
```

5. Because the module is compiled into QTEMP first, the compiler initially creates the EVFEVENT file/member in QTEMP:

```text
CPC7301: File EVFEVENT created in library QTEMP.
CPC7305: Member TESTPGM added to file EVFEVENT in QTEMP.
```

6. Afterwards, the wrapper copies/adds the EVFEVENT member into the target library:

```text
CPF2889: Member TESTPGM added to file EVFEVENT in TESTLIB.
CPC2955: 7 records copied from member TESTPGM.
```

7. Code for IBM i then tries to fetch diagnostics from the correct target library/member:

```text
Fetching errors for TESTLIB/TESTPGM.
```

8. EVFEVENT fetching fails in Code for IBM i 3.0.10.
9. Downgrade to Code for IBM i 3.0.7.
10. Run the same compile wrapper action again.
11. Diagnostics are fetched correctly and shown in the Problems view.

**Expected behavior**

Code for IBM i should fetch diagnostics from:

```text
TESTLIB/EVFEVENT(TESTPGM)
```

and show the compile diagnostics in the Problems view.

**Screenshots**

I cannot provide screenshots or full output logs because they contain internal source code, library names, user names, paths, server names, and company-specific information.

However, the relevant sanitized behavior is included above.

**Environment**

* Extension version failing: 3.0.10
* Extension version working: 3.0.7
* IBM i OS version: V7R5M0
* Source type: RPGLE / SQLRPGLE
* Compile option: OPTION(*EVENTF)
* Compile method: custom compile wrapper program

**Additional context**

The same Code for IBM i version 3.0.10 works on another IBM i system with the same IBM i release.

So this might not be a simple version-only regression. It may be related to the combination of:

* Code for IBM i 3.0.10
* custom compile wrapper program
* EVFEVENT initially created in QTEMP
* EVFEVENT member copied to the target library
* server-specific environment differences

Possible areas to investigate:

* EVFEVENT retrieval changes after 3.0.7
* OVRDBF-based EVFEVENT fetching
* copied EVFEVENT members
* compile wrapper behavior
* QTEMP to target library EVFEVENT handling
* CCSID / job CCSID differences
* current library / library list handling
* server-specific job environment differences

Important point:

Code for IBM i appears to identify the correct target library/member because the output contains:

```text
Fetching errors for TESTLIB/TESTPGM.
```

So the failure seems to happen during the actual EVFEVENT fetch, not during target library/member detection.

sebjulliand:
### Changes

Fixes #3304
Fixes #3241

The logic used to identify the iasp of source members linked to compilation problems was wrong. On top of that, a Promise was not awaited and concatenated with a string which lead to the wrong path being created - as seen here:
image

This PR fixes this and also discards the problems returned by the parser for the `.` file.

### How to test this PR

1. Use a profile with an iASP
2. Compile a program with errors, using a source member located on this iASP 
3. The problems view should list all the issues for that member
4. Clicking on a problem should focus the right line if the member is opened
5. It should open the correct member and focus on the line if the member was closed  

### Checklist

* [x] have tested my change

### DIFF
--- a/src/ui/diagnostics.ts
+++ b/src/ui/diagnostics.ts
@@ -72,7 +72,7 @@ export function refreshDiagnosticsFromServer(connection: IBMi, evfeventInfo: Evf
         await handleEvfeventLines(connection, lines, e);
       }
     }
-    finally{
+    finally {
       connection.getContent().deleteOVRDBFile(overFile);
     }
   });
@@ -109,32 +109,33 @@ export async function handleEvfeventLines(connection: IBMi, lines: string[], evf
 
   const diagnostics: vscode.Diagnostic[] = [];
   if (errorsByFiles.size) {
-    for (const [file, errors] of errorsByFiles.entries()) {
-      diagnostics.length = 0;
-      for (const error of errors) {
-        error.column = Math.max(error.column - 1, 0);
-        error.lineNum = Math.max(error.lineNum - 1, 0);
-        error.toLineNum = Math.max(error.toLineNum - 1, 0);
+    for (const [file, errors] of errorsByFiles) {
+      if (file !== '.') {
+        for (const error of errors) {
+          error.column = Math.max(error.column - 1, 0);
+          error.lineNum = Math.max(error.lineNum - 1, 0);
+          error.toLineNum = Math.max(error.toLineNum - 1, 0);
 
-        if (error.column === 0 && error.toColumn === 0) {
-          error.column = 0;
-          error.toColumn = 100;
-        }
+          if (error.column === 0 && error.toColumn === 0) {
+            error.column = 0;
+            error.toColumn = 100;
+          }
 
-        const diagnostic = new vscode.Diagnostic(
-          new vscode.Range(error.lineNum, error.column, error.toLineNum, error.toColumn),
-          `${error.text} (${error.sev})`,
-          diagnosticSeverity(error)
-        );
+          const diagnostic = new vscode.Diagnostic(
+            new vscode.Range(error.lineNum, error.column, error.toLineNum, error.toColumn),
+            `${error.text} (${error.sev})`,
+            diagnosticSeverity(error)
+          );
 
-        diagnostic.code = error.code;
+          diagnostic.code = error.code;
 
-        if (config) {
-          if (!config.hideCompileErrors.includes(error.code)) {
+          if (config) {
+            if (!config.hideCompileErrors.includes(error.code)) {
+              diagnostics.push(diagnostic);
+            }
+          } else {
             diagnostics.push(diagnostic);
           }
-        } else {
-          diagnostics.push(diagnostic);
         }
       }
 
@@ -186,7 +187,7 @@ export async function handleEvfeventLines(connection: IBMi, lines: string[], evf
       }
       else {
         const asp = await connection.getLibraryIAsp(file.split('/')[0]);
-        const memberUri = VscodeTools.findExistingDocumentUri(vscode.Uri.from({ scheme: `member`, path: `/${asp ? `${asp}/` : '' }${file}${evfeventInfo.extension ? `.` + evfeventInfo.extension : ``}` }));
+        const memberUri = VscodeTools.findExistingDocumentUri(vscode.Uri.from({ scheme: `member`, path: `/${asp ? `${asp}/` : ''}${file}${evfeventInfo.extension ? `.` + evfeventInfo.extension : ``}` }));
         ileDiagnostics.set(memberUri, diagnostics);
       }
     }
--- a/src/ui/diagnostics.ts
+++ b/src/ui/diagnostics.ts
@@ -69,7 +69,7 @@ export function refreshDiagnosticsFromServer(connection: IBMi, evfeventInfo: Evf
         .map(row => String(row.EVFEVENT));
 
       if (lines.length) {
-        handleEvfeventLines(connection, lines, e);
+        await handleEvfeventLines(connection, lines, e);
       }
     }
     finally{
@@ -94,7 +94,7 @@ export async function refreshDiagnosticsFromLocal(connection: IBMi, evfeventInfo
         const eol = content.includes(`\r\n`) ? `\r\n` : `\n`;
         const lines = content.split(eol);
 
-        handleEvfeventLines(connection, lines, evfeventInfo);
+        await handleEvfeventLines(connection, lines, evfeventInfo);
       }
 
     } else {
@@ -103,10 +103,8 @@ export async function refreshDiagnosticsFromLocal(connection: IBMi, evfeventInfo
   }
 }
 
-export function handleEvfeventLines(connection: IBMi, lines: string[], evfeventInfo: EvfEventInfo) {
+export async function handleEvfeventLines(connection: IBMi, lines: string[], evfeventInfo: EvfEventInfo) {
   const config = connection.getConfig();
-  const asp = evfeventInfo.asp ? `${connection.getLibraryIAsp(evfeventInfo.library)}/` : ``;
-
   const errorsByFiles = parseErrors(lines);
 
   const diagnostics: vscode.Diagnostic[] = [];
@@ -151,9 +149,9 @@ export function handleEvfeventLines(connection: IBMi, lines: string[], evfeventI
           let relativeCompilePath = (deployPathIndex !== -1 ? file.substring(0, deployPathIndex) + file.substring(deployPathIndex + workspaceDeployPath.length) : undefined);
 
           if (relativeCompilePath) {
-            if (asp) {
+            if (evfeventInfo.asp) {
               // Believe it or not, sometimes if the deploy directory is symlinked into an ASP, this can be a problem              
-              const aspRoot = `/${asp}`;
+              const aspRoot = `/${evfeventInfo.asp}`;
               if (relativeCompilePath.startsWith(aspRoot)) {
                 relativeCompilePath = relativeCompilePath.substring(aspRoot.length);
                 break;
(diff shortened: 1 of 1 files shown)


buzzia2001 (approved):
Thank you @sebjulliand

PR #3312, opened 2026-06-19
Support for SSH agent

### LINKED ISSUE #3299: Support native SSH Agent connection instead of manual key path and password configuration
Currently, the extension requires users to manually specify the absolute path to their private SSH key and hardcode their password/passphrase within the extension settings.
This approach has several drawbacks:
1. Redundancy: Users have to replicate their SSH configuration across every single application or extension they use, rather than having a single, centralized source of truth.
2. Security: Storing passphrases or key configurations directly within application settings is less secure than relying on system-level cryptographic stores.
3. Friction: If a user rotates their keys or changes their passphrase, they have to manually update vscode-ibmi settings.
 
We would love to see an option (or a default behavior) that allows vscode-ibmi to connect using the native system SSH Agent (like ssh-agent on macOS/Linux or the OpenSSH Authentication Agent on Windows).
Instead of asking for a specific file path and passphrase, the extension would simply look for keys already loaded and authenticated in the system's keychain/agent. If the local system is already configured via ~/.ssh/config to automatically load keys (using AddKeysToAgent yes or UseKeychain yes), the extension would seamlessly inherit this authentication state.
This makes the onboarding process incredibly smooth: if your terminal can already ssh user@ibmi without a password prompt, vscode-ibmi should be able to do the exact same out of the box.

Current behavior: Manually maintaining paths and secrets inside the extension settings, which is tedious and introduces configuration drift between the terminal and VS Code.

**Additional context**
Most modern VS Code extensions utilizing SSH (like Microsoft's official Remote - SSH) rely entirely on the system's local SSH agent and configuration file, rather than managing keys internally. Implementing this would align vscode-ibmi with modern SSH best practices.

Thank you for the great work on this extension, just an idea of improvement

buzzia2001:
### Changes
With this PR support for SSH-AGENT as been included

### How to test this PR
1. Create a ssh private key with ssh-keygen
2. Import pub key in authorized_keys 

**Windows:**
3. Run Powershell as admin
4. Make sure the ssh-agent service is running (to start it, run `Start-Service ssh-agent`)
5. Import your private key: `ssh-add C:\Users\XXXXXX\.ssh\id_ed25519`
6. Check that is ok with: `ssh-add -L`

**Mac/Linux:**
7. Open your terminal
8. Check ssh-agent is active: `eval "$(ssh-agent -s)"`
7. Import your private key: `ssh-add /home/XXXXXX/.ssh/id_ed25519`
9. Check that is ok with: `ssh-add -L`

10. Update your connection to use SSH-AGENT:
image
12. Connect to the system

### Checklist
* [x] have tested my change
* [x] have update one or more test cases

Closes #3299


src/api/IBMi.ts
```diff
@@ -278,12 +278,39 @@ export default class IBMi {
         }
       }
 
-      await this.client.connect({
+      const connectConfig: any = {
```
sebjulliand: You can use the actual type `node_ssh.Config` here:
```suggestion
      const connectConfig: node_ssh.Config = {
```

src/webviews/login/index.ts
```diff
@@ -31,12 +31,13 @@ export class Login {
       .addInput(`port`, l10n.t(`Port (SSH)`), ``, { default: `22`, min: 1, max: 65535, inputType: "number" })
       .addInput(`username`, l10n.t(`Username`), undefined, { minlength: 1, maxlength: 10 })
       .addHorizontalRule()
-      .addParagraph(l10n.t(`Only provide either the password or a private key - not both.`))
+      .addParagraph(l10n.t(`Choose only one authentication method from among password, private key, or SSH agent`))
```
sebjulliand: Small typo
```suggestion
      .addParagraph(l10n.t(`Choose only one authentication method among password, private key, or SSH agent`))
```

src/webviews/settings/index.ts
```diff
@@ -438,11 +438,12 @@ export class SettingsUI {
               .addInput(`port`, vscode.l10n.t(`Port (SSH)`), undefined, { default: String(stored.port), min: 1, max: 65535, inputType: "number" })
               .addInput(`username`, vscode.l10n.t(`Username`), undefined, { default: stored.username, minlength: 1 })
               .addHorizontalRule()
-              .addParagraph(vscode.l10n.t(`Only provide either the password or a private key - not both.`))
+              .addParagraph(vscode.l10n.t(`Choose only one authentication method from among password, private key, or SSH agent`))
```
sebjulliand: Small typo
```suggestion
              .addParagraph(vscode.l10n.t(`Choose only one authentication method among password, private key, or SSH agent`))
```

sebjulliand (changes requested):
@buzzia2001 great work; that's going to be a useful change, no doubt.
I'll just need you to fix a type and a small typo (in the l10n files as well).

But it works great!

After it's merged, the documentation will need to be updated.

buzzia2001:
@sebjulliand thank you for the correction... 

To you!

sebjulliand (approved):
Perfect! Approved !

PR #3315, opened 2026-06-21
Add support for double-extension action matching

### LINKED ISSUE #2741: Enable actions that are specific to double extensions like .pgm.clle or .pgm.rpgle
We use the naming convention where a .rpgle or .clle extension is for modules, whereas .pgm.rpgle or .pgm.clle is for bound programs. 
Right now, there is no way to have actions specific to modules vs. bound programs.

It would be nice to have the ability to define actions specific to the .pgm.rpgle/.pgm.clle source types (CRTBNDCL/CRTBNDRPG) and the .rpgle/.clle types (CRTCLMOD, CRTRPGMOD).

yuvalnn:
Fixes #2741
### Changes
- Added support for actions targeting double extensions such as PGM.RPGLE and PGM.CLLE.

### How to test this PR
1. Create actions for both RPGLE and PGM.RPGLE.
2. Verify that a file such as foo.pgm.rpgle matches the expected action.
3. Verify that existing single-extension actions still work.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/ui/actions.ts
+++ b/src/ui/actions.ts
@@ -670,9 +670,8 @@ export async function getAllAvailableActions(targets: ActionTarget[], scheme: st
   });
 
   // Then we get all the available Actions for the current context
-  const availableActions: AvailableAction[] = allActions.filter(action => action.type === scheme)
-    .filter(action => !action.extensions || action.extensions.every(e => !e) || targets.every(t => action.extensions!.includes(t.extension) || action.extensions!.includes(t.fragment)) || action.extensions.includes(`GLOBAL`))
-    .filter(action => action.runOnProtected || !targets.some(t => t.protected))
+  const availableActions: AvailableAction[] = allActions
+    .filter(action => isActionAvailable(action, scheme, targets))
     .sort((a, b) => (actionUsed.get(b.name) || 0) - (actionUsed.get(a.name) || 0))
     .map(action => ({
       label: action.name,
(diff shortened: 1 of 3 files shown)


src/ui/actions.ts
```diff
@@ -682,6 +681,41 @@ export async function getAllAvailableActions(targets: ActionTarget[], scheme: st
   return availableActions;
 }
 
+export function isActionAvailable(action: Action, scheme: string, targets: ActionTarget[]): boolean {
+  if (action.type !== scheme) return false;
+
+  // action isn't cleared to run on protected targets and some of them are 
+  if (!action.runOnProtected && targets.some(t => t.protected)) return false;
+
+  return targets.every((t) => targetMatchesExtensions(t, action.extensions));
```
sebjulliand, line 690: You can fuse all the checks in one statement; that will avoid intermediate `return`s.
```suggestion
  return action.type === scheme &&
          // action isn't cleared to run on protected targets and some of them are 
          (action.runOnProtected || !targets.some(t => t.protected)) &&
          targets.every((t) => targetMatchesExtensions(t, action.extensions));
```

src/ui/actions.ts
```diff
@@ -682,6 +681,41 @@ export async function getAllAvailableActions(targets: ActionTarget[], scheme: st
...
+}
+
+export function targetMatchesExtensions(target: ActionTarget, extensions?: string[]): boolean {
+  // action has no extension requirements, or is global
+  if (!extensions || extensions.every(e => !e) || extensions.includes("GLOBAL")) return true;
```
sebjulliand, line 695: Checking if `extensions` is truthy or empty will be less confusing than doing it with `extensions.every(e => !e)`
```suggestion
  if (!extensions?.length || extensions.includes("GLOBAL")) return true;
```

src/ui/actions.ts
```diff
@@ -682,6 +681,41 @@ export async function getAllAvailableActions(targets: ActionTarget[], scheme: st
...
+
+  const targetExtParts = [target.extension.toUpperCase(), target.fragment.toUpperCase()];
```
sebjulliand, line 697: Filter on distinct and filter out empty string here to avoid unwanted iterations in the for loop after. 
```suggestion
const targetExtParts = [target.extension.toUpperCase(), target.fragment.toUpperCase()]
    .filter(Boolean)
    .filter(Tools.distinct);
```

sebjulliand (changes requested):
Thanks for looking into this @yuvalnn !
I left a few suggestions for code clarity.

However, the logic you implemented needs to be changed a bit, regarding the initial need.

With this PR, I'm still able to run an `RPGLE` action on a `.pgm.rpgle` file. So, i'll be presented with both "create a module" and "create a program" action on a file that I want to compile only into a program.

The logic should be something like:

-> if the action extension ends with `pgm.xxx`, then target file path must end with `.pgm.xxx` to be a valid target.

-> if the action extension is only `xxx`, then files ending with `pgm.xxx` must not be valid targets.

-> All of the above is only valid for local files or IFS files. For the other target, strictly checking the extension (or fragment, for objects) is enough


One way to check the multi-part extensions would be to use split(".") on both the action extension and the target file name and use these arrays to check the target validity.

Something along these lines:
```ts
  if (["file", "streamfile"].includes(target.uri.scheme)) {
    //Check all extension parts for workspace files or IFS files
    const parts = path.parse(target.uri.path).base.split(".").map(t => t.toUpperCase());
    parts.splice(0, 1);
    for (const e of extensions.map(ext => ext.split('.'))) {
      //every parts must match
      if (e.every((val, index) => val === parts[index])) {
        return true;
      }
    }
  }
  else {
    //Check extension only for 
    return [target.extension.toUpperCase(), target.fragment.toUpperCase()]
      .filter(Tools.distinct)
      .some(e => extensions.includes(e));
  }
  return false;
```
Hope this helps 😄

PR #3317, opened 2026-06-22
Fix: add DETAILED_INFO=>'NO' to QSYS2.LIBRARY_INFO table function calls

bobcozzi:
## Problem

`QSYS2.LIBRARY_INFO` without `DETAILED_INFO => 'NO'` retrieves additional metadata columns that are not needed in either call site. This adds unnecessary overhead on every library lookup.

## Fix

Added `DETAILED_INFO => 'NO'` to both locations where `QSYS2.LIBRARY_INFO` is called:

- **`src/api/IBMi.ts`** — `getLibraryIAsp()`: only `IASP_NAME` is needed
- **`src/api/IBMiContent.ts`** — library list query: only the columns already referenced (`NAME`, `TYPE`, etc.) are needed; detailed info is not used

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -149,7 +149,7 @@ export default class IBMi {
   /**
    * getConfigFile can return pre-defined configuration files,
    * but can lazy load new configuration files as well.
-   * 
+   *
    * This does not load the configuration file from the server,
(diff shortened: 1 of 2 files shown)


sebjulliand (approved):
Good one, thanks @bobcozzi !

PR #3319, opened 2026-06-22
Improvements to IFS Permission Control

### LINKED ISSUE #3316: Home directory permissions prompt should not appear when current permissions are more strict than suggested
Currently, when home directory permissions are more strict than suggested 750 (e.g. 700), a prompt like below is displayed.

> Directory permissions
> Security recommendation: Update directory permissions?
> • Home directory (/home/XXX) has permissions 700, recommended: 750
> Would you like to update these permissions for better security?

It should appear only when current permissions are less strict than 750

buzzia2001:
### Changes
This PR introduces a dynamic mechanism for controlling IFS permissions.

### How to test this PR

If you connect to a system with a home directory set to 700 permissions, you will no longer receive warnings about incorrect permissions.

### Checklist
* [x] have tested my change

Closes #3316

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -465,9 +465,9 @@ export default class IBMi {
 
         // Check if permissions need updating
         const needsPermissionUpdate =
-          (homePerms !== 'error' && homePerms !== '750') ||
+          (homePerms !== 'error' && !Tools.checkPermissionsValue(homePerms, 750)) ||
           !vscodeExists ||
-          (vscodeExists && vscodePerms !== 'error' && vscodePerms !== '700');
+          (vscodeExists && vscodePerms !== 'error' && !Tools.checkPermissionsValue(vscodePerms, 700));
 
         if (needsPermissionUpdate && !options.reconnecting) {
           await callbacks.uiErrorHandler(this, `home_directory_permissions`, {
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -465,9 +465,9 @@ export default class IBMi {
 
         // Check if permissions need updating
         const needsPermissionUpdate =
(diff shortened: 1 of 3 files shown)


src/api/Tools.ts
```diff
@@ -278,4 +278,21 @@ export namespace Tools {
...
+   * @param permission current permissions as a three-digit octal string
+   * @param target optional target permission value to compare against
+   * @returns `true` if each digit of `permission` is less than or equal to the
+   *          corresponding digit of `target`, otherwise `false`
+   */
+  export function checkPermissionsValue(permission: string, target: number = 750): boolean | undefined {
+    const curPermission = Array.from(permission).map(Number);
+    const curTarget = Array.from(target.toString()).map(Number);
+
+    return curPermission.every((val, i) => val <= curTarget[i]);
```
rgala: For group permissions -wx (3) or -w- (2) this will return true in some situations, like target 750 or 740 etc. I think for home directories write should be allowed only for owner

src/api/IBMi.ts
rgala, line 1: This should also be added in  ui/connection.ts

buzzia2001:
Hi @rgala 750 is the minimum acceptable limit defined by IBM... although I agree that we could restrict it even further to 700, but for now we'll stick with their recommendation.

src/api/Tools.ts
sebjulliand: What's the purpose of `checkPermissionsValue` since it's not used anywhere? 🤔

sebjulliand (changes requested):

src/api/Tools.ts
buzzia2001: you are right, I've already removed

buzzia2001:
@sebjulliand to you

sebjulliand (approved):
Good to go!

PR #3322, opened 2026-06-23
feat: expose getSystemVersion() in CodeForIBMi public API

bobcozzi:
## Summary

Exposes `getSystemVersion()` as a first-class member of the `CodeForIBMi` public API interface, allowing consumer extensions (e.g. `vscode-db2i`) to retrieve the IBM i OS version without accessing internal implementation details.

## Changes

- **`src/typings.ts`** — Added `getSystemVersion: () => number | undefined` to the `CodeForIBMi` interface
- **`src/extension.ts`** — Wired up the implementation in the `activate()` return object via `instance.getConnection()?.getSystemVersion()`

## Usage (from a consumer extension)

```typescript
const api = vscode.extensions.getExtension<CodeForIBMi>('halcyontechltd.code-for-ibmi')?.exports;
const version = api?.getSystemVersion(); // e.g. 7.5
```

Returns `undefined` if no connection is active, consistent with other connection-dependent API members.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -153,7 +153,8 @@ export async function activate(context: ExtensionContext): Promise<CodeForIBMi> 
     actionTools: ActionTools,
     componentRegistry: extensionComponentRegistry,
     connectionManager: IBMi.connectionManager,
-    searchTools: SearchTools
+    searchTools: SearchTools,
+    getSystemVersion: () => instance.getConnection()?.getSystemVersion(),
   };
 }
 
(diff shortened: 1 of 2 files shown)


SanjulaGanepola:
@bobcozzi The `getSystemVersion` API is already accessible to other extensions:
```ts
const ibmi = getInstance();
const connection = ibmi!.getConnection();
const systemVersion = connection.getSystemVersion();
```

bobcozzi:
Okay, so then they should use it instead of that length SYSIBM function that takes centuries.

PR #3328, opened 2026-06-24
Use node's SHA256 to hash routines body and add 1 minute cooldown between components check

sebjulliand:
, ### Changes

The Db2 for i `HASH_SHA256` built-in function is not available on 7.3.
Therefor, the ValidateStatement SQL function from Db2 for i isn't checked on 7.3 but can still be used, unsecurely. 

This PR enhance the `getSQLRoutineSignature` method to fix that:
- The method will now return the routine's external program for an external routine
- It will return the routine's body hashed locally using node's crypto sha256 algorithm

It also adds a one minute cooldown between two component's state check to reduce the workload for some component that were checked every seconde or (like Db2 syntax checker).

NB: once this is released, Db2 extension will have to be updated and released right after with the new SHA256 hash as the result from the local hash differs from remote's.

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -1,3 +1,4 @@
+import crypto from 'crypto';
 import { parse } from 'csv-parse/sync';
 import fs from 'fs';
 import * as node_ssh from "node-ssh";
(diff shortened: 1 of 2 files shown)


buzzia2001 (approved):
It works fine, thank you @sebjulliand

PR #3337, opened 2026-06-30
Mapepire config section

sebjulliand:
### Changes

This PR adds a specfic `Mapepire` section to th conection settings, regrouping all the settings related to Mapepire and the SQL connection it creates for the core extension.

image

### How to test this PR

1. Open connection settings
2. There must be a Mapepire section

### Checklist

* [x] have tested my change


buzzia2001:
Bonne soirée @sebjulliand , 
I really like the idea of having a dedicated panel for Mapepire since it lightens the load on the other tabs... 
But I was thinking that maybe it would also be a good idea to logically group the various settings while we’re at it:
- general settings (SSL/spool/naming)
- Mapepire SSH client (java version)
- Mapepire HTTP client (connect to remote/ports/certificates)

At least that seems more organized to me...

Ideally, we could even add a switch that specifies the type of mapepire being used (ssh/http) and display only the information related to it...

What do you think about?

sebjulliand:
Agreed, I'll regroup the settings as you suggested.

The switch bothers me a bit since the UI framework doesn't support dynamic show:hide of fields depending on other fields. We could disable the fields once connected since we know which connection type is being used at that time. But not when disconnected though.

buzzia2001:
> Agreed, I'll regroup the settings as you suggested.
> 
> The switch bothers me a bit since the UI framework doesn't support dynamic show:hide of fields depending on other fields. We could disable the fields once connected since we know which connection type is being used at that time. But not when disconnected though.

Merci beaucoup, mon ami

sebjulliand:
Prego @buzzia2001 😄 

How about now?
image

buzzia2001:
> Prego @buzzia2001 😄 
> 
> 
> 
> How about now?
> 
> image
> 
> 


It's perfect!
Thank you again!

PR #3342, opened 2026-07-01
Enhance action sorting options for actions

### LINKED ISSUE #3338: Sort by Natural order for actions from actions.json
Based on the fullfilled request https://github.com/codefori/vscode-ibmi/issues/3309, 
i wanted to ask for an additional sorting order in which the elements are ordered by their order inside actions.json.

buzzia2001:
### Changes
This PR introduces support for displaying actions in four ways:
1. by usage
2. by name
3. based on the original configuration file
4. based on the actions.jso file

### How to test this PR
Change **Code-for-ibmi: Sort Actions By** in Preferences, now you'll see actions depending on what you decide to sort:

image

### Checklist
* [x] have tested my change

Closes #3338

### DIFF
--- a/src/ui/actions.ts
+++ b/src/ui/actions.ts
@@ -645,7 +645,8 @@ export async function runAction(instance: Instance, uris: vscode.Uri | vscode.Ur
 export type AvailableAction = { label: string; action: Action; }
 
 export async function getAllAvailableActions(targets: ActionTarget[], scheme: string) {
-  const allActions = [...await ActionTools.getActions()];
+  const configActions = await ActionTools.getActions();
+  const actionsFileActions: Action[] = [];
 
   // Then, if we're being called from a local file
   // we fetch the Actions defined from the workspace.
@@ -657,10 +658,11 @@ export async function getAllAvailableActions(targets: ActionTarget[], scheme: st
     const allTargetsInOne = targets.every(t => t.workspaceFolder?.index === workspaceId);
 
     if (allTargetsInOne) {
-      const localActions = await ActionTools.getActions(firstWorkspace);
-      allActions.push(...localActions);
+      actionsFileActions.push(...await ActionTools.getActions(firstWorkspace));
     }
   }
+
+  const allActions = [...configActions, ...actionsFileActions];
 
   // We make sure all extensions are uppercase
   allActions.forEach(action => {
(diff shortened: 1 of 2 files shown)


sebjulliand (changes requested):
Thanks for looking into this @buzzia2001 .

However, that change breaks the UX.
- Actions should not be hidden based on which sorting order is selected - especially if the user is working on local projects and members alike.
- Choosing `config` renders the local actions unusable from a local project
- Choosing `actionsFile` shows a warning every time an action is run on a member or IFS file 

For example, I usually work in a hybrid environment, with a local project opened in VS Code but I also work on members at the same time. I don't see myself going back and forth in the settings to change the sorting order 😉 

Besides, it's about sorting, this should not hide anything.

So, I'd suggest you just keep the `config` option that will sort actions in the order they are defined in the configuration, regardless of their origin.

Thanks!

buzzia2001:
Hi @sebjulliand ,
to you.

Thank you for your suggestion

src/ui/actions.ts
```diff
@@ -673,22 +673,23 @@ export async function getAllAvailableActions(targets: ActionTarget[], scheme: st
   const sortBy = IBMi.connectionManager.get<string>(`sortActionsBy`) || `usage`;
```
sebjulliand: Let's make the possible values explicit here. It will be clearer to read.
```suggestion
  const sortBy = IBMi.connectionManager.get<'name'|'usage'|'config'>(`sortActionsBy`) || `usage`;
```

src/ui/actions.ts
```diff
@@ -673,22 +673,23 @@ export async function getAllAvailableActions(targets: ActionTarget[], scheme: st
...
-    .map(action => ({
-      label: action.name,
-      action
-    }));
+    .filter(action => action.runOnProtected || !targets.some(t => t.protected));
+
+  if (sortBy === `name`) {
+    // Sort alphabetically by name
+    contextActions.sort((a, b) => a.name.localeCompare(b.name));
+  } else if (sortBy !== `config`) {
```
sebjulliand: Make the comparison a strict equal on the expected config value. It will makes it easier to understand for whoever will read that code.
```suggestion
  } else if (sortBy === `usage`) {
```

sebjulliand (changes requested):
Greta job @buzzia2001 , thanks for the modification.
I just made a couple of suggestions to make the code more explicit.
We'll merge after that.

buzzia2001:
I followed your advice, and I've pushed

sebjulliand (approved):
Many thanks @buzzia2001 ☺️

PR #3346, opened 2026-07-04
new job info menubar with Mapepire Job Infos

NicolasSchindler:
### Changes
Added a new Menubar with Information about the Mapepire Job used for the Core Extension.
image
Moved the Job ID from the existing "ConnectedBar" to the new "JobInfoBar". OnClick is still the same (copyJobID).
There is a new Tooltip Menu first shows the current JDBC Config used for the Job (SQL Naming, secure mode, all changed settings that are not default, so if we change them in the future the user can see it there). So the user can see what configuration for Job Naming is used for the current Job, currently you can't see that Information.
And the "Job Info" which opens a new Window with WRKJOB Information.

### How to test this PR
Connect to a System and then there is a new Menubar on the bottom contains the Job name. 
Hovering over shows the new Menu with the Information about the Job.

### Checklist
* [X] have tested my change


SanjulaGanepola:
@NicolasSchindler @buzzia2001 Now that https://github.com/codefori/vscode-ibmi/pull/3348 was merged, shall we look to get this PR updated based on those changes.

In regards to creating a new status bar item, I wonder if it would be better to enhance the current one with the ability to view the job log. The only reason being that the status bar always gets filled with several items and it becomes hard to fit stuff depending on how zoomed in you are. For IBM i, having everything under a single status bar item hover seems far easier to find and more discoverable. What do you think?

image

For example, other status bar items (like the one below) have rich UI with buttons for other operations. If this is possible, I wonder if just enhancing the current status bar hover by adding a `info` icon to the right of the job ID to open the new job log UI (while keeping the current copy command on the job ID label) would be good?

image

buzzia2001:
Ciao @SanjulaGanepola and @sebjulliand,
here the version updated, let me know if it's ok for you... The style and entries are the same as those in the database extension


image


database:
image

SanjulaGanepola:
Yep, looks great.

Can you re-sync with main? GitHub still says 33 file changes including stuff from other PRs

src/commands/connection.ts
```diff
@@ -30,5 +30,26 @@ export function registerConnectionCommands(context: ExtensionContext, instance:
...
+      } else {
+        window.showWarningMessage(`No job ID available`);
+      }
+    }),
+    commands.registerCommand(`code-for-ibmi.showIBMiJobInfo`, async () => {
+      const connection = instance.getConnection();
+      if (connection) {
+        const job = await connection.getSqlJobId();
+        if (job) {
+          await commands.executeCommand('vscode-ibmi-fs.wrkjob', job);
```
SanjulaGanepola: We are switching this to `code-for-ibmi.showJobLog` right?

src/instantiate.ts
```diff
@@ -31,6 +31,12 @@ connectedBarItem.command = {
   title: `Show connection settings`
 };
 
+const jobInfoBarItem = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Left, 10);
+jobInfoBarItem.command = {
+  command: `code-for-ibmi.copyJobId`,
+  title: `Copy job ID to clipboard`
+};
+
```
SanjulaGanepola: Are we removing this given the job info moved into the existing connected `connectedBarItem`?

src/extension.ts
```diff
@@ -131,6 +131,7 @@ export async function activate(context: ExtensionContext): Promise<CodeForIBMi>
       commands.executeCommand("code-for-ibmi.refreshLibraryListView");
       commands.executeCommand("code-for-ibmi.refreshIFSBrowser");
       commands.executeCommand("code-for-ibmi.environment.refresh");
+      commands.executeCommand("code-for-ibmi.updateJobInfoBar");
```
SanjulaGanepola: This doesn't look to be a registered command. My understanding is that updating of the status bar should be handled by `onConnected` in the `src\instantiate.ts` already

buzzia2001:
@SanjulaGanepola here the new version:

1. moved parseStatusBarColor inside tools file (updated imports)
2. fix mapepire config page 
3. deleted command code-for-ibmi.updateJobInfoBar, modified onConnected to update this bar
4. deleted refs to jobInfoBarItem

buzzia2001:
I've also exported onCodeForIBMiConfigurationChange on the public API

SanjulaGanepola (approved):
Nice work!

PR #3347, opened 2026-07-04
Fixed errors being stacked when reported on multiple files

### LINKED ISSUE #3345: RPGLE compile diagnostics are duplicated in Problems view for included copybooks
When compiling RPGLE programs, diagnostics are shown not only in the compiled source member, but also in every copybook that the program includes. This causes the same errors to be duplicated or multiplied in the Problems view and makes it much harder to identify the actual source of the problem. In the current scenario, a single compile issue is reported against the main program and also against all included copybooks, which creates a lot of noise.

To Reproduce

Open an RPGLE program that includes one or more copybooks.
Compile the program.
Open the Problems view.
Notice that diagnostics are shown for the compiled source member and also for the included copybooks, even though the problem originates from the compile of the main program.
Expected behavior

Compile diagnostics should be reported only at the relevant source member(s) that actually caused the error. Copybooks should not receive duplicate diagnostics just because they were included by the program being compiled.

Screenshots

Image

Environment (please complete the following information)

Extension version: 3.0.11
IBM i OS version: V7R5

sebjulliand:
### Changes
Fixes #3345 

This PR fixes a bug in the loop gathering the errors for each files from EVFEVENT. The `diagnostics` array was not cleared after each iteration, so the errors would be stacked and reported for each subsequent member being processed. 

Now the array is cleared and errors are only reported on the member they belong to.

### How to test this PR
1. Create the example described by @SJLennon here: https://github.com/codefori/vscode-ibmi/issues/3345#issuecomment-4878311362 (or compile a member, with an error, that includes another member)
2. Compile
3. Errors must be reported only in the main member

### Checklist

* [x] have tested my change

### DIFF
--- a/src/ui/diagnostics.ts
+++ b/src/ui/diagnostics.ts
@@ -107,10 +107,10 @@ export async function handleEvfeventLines(connection: IBMi, lines: string[], evf
   const config = connection.getConfig();
   const errorsByFiles = parseErrors(lines);
 
-  const diagnostics: vscode.Diagnostic[] = [];
   if (errorsByFiles.size) {
     for (const [file, errors] of errorsByFiles) {
       if (file !== '.') {
+        const diagnostics: vscode.Diagnostic[] = [];
         for (const error of errors) {
           error.column = Math.max(error.column - 1, 0);
           error.lineNum = Math.max(error.lineNum - 1, 0);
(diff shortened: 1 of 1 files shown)


buzzia2001:
It works fine, thank you @sebjulliand for the resolution and many thanks to @SJLennon that create an easy test case

PR #3358, opened 2026-07-12
Specific color for the connection in the status bar

### LINKED ISSUE #3078: Allow colour coding part of the interface based on the currently connected host
Simultaneous connection to multiple hosts requires multiple windows. All of those windows look the same, so it is very easy to use the wrong one.

See if you can tell the host name from this screen shot _at a glance_. It's buried amidst a lot of other text and icons and requires concentration to explicitly seek out each time.

Image

Bespoke colour configurations of ACS 5250 session profiles (most people change the font colour for 'green' to something else; I change the background of the OIA) have long been used to distinguish hosts.

ACS RSS window allows setting a coloured border which is stored per host for the same benefit.

I would like to see Code for i adopt a similar approach of allowing a connection-configured colour.

At a minimum, this could be surfaced on that small icon/name in the status bar. (I used to use a git enhancing extension that would give the branch name a bright red background for configured branches - by default, main).

Alternatively, allow 'skinning' some larger part of the interface in a colour. I have seen an extension called "Peacock" which sets a colour of the user's choice on title, status, and side bars.

I am not aware of any means to configure a noticeable indicator of the connected host, automatically.

**Additional context**

buzzia2001:
### Changes
This PR adds the ability to specify the color to apply to the session name in the status bar, so that you can distinguish between the various open connections on different systems.

### How to test
1. Going to system settings inside FEATURES TAB, you'll fin a new form that asks you hexadecimal color for your session, then save changes:
image

2. Connect to your system, once the connection complete you'll see name of your session with a different color:
image

### Checklist
* [x] have tested my change

Closes #3078


src/instantiate.ts
```diff
@@ -142,6 +146,8 @@ async function onConnected() {
     connectedBarItem,
     disconnectBarItem,
   ].forEach(barItem => barItem.show());
+
+  connectedBarItem.color = parseStatusBarColor(instance.getConnection()?.getConfig().statusBarColor);
```
SanjulaGanepola: This line looks unnecessary as we already set the color in the `updateConnectedBar` call above.

SanjulaGanepola (changes requested):

sebjulliand:
Thans for picking that up @buzzia2001 !

A few remarks:
- Could we use a color picker to select the color ? (VS Code element provides one as far as I know)
- Adding another field to set the background color could be nice
- Can the color(s) in the bar be changed automatically when the settings are updated? (using the onConfigChange listener?)

buzzia2001:
> Thans for picking that up @buzzia2001 !
> 
> A few remarks:
> 
> * Could we use a color picker to select the color ? (VS Code element provides one as far as I know)
> * Adding another field to set the background color could be nice
> * Can the color(s) in the bar be changed automatically when the settings are updated? (using the onConfigChange listener?)

Hi @SanjulaGanepola @sebjulliand, 

-> @sebjulliand 
I'll try to look into the color picker issue today... 
When you say “background color,” do you mean just the block or the entire bar? At first I tried the whole bar, but I ran into problems in standalone/workspace environments, so I switched to the block...
I agree with you about the onchange event, though… I’ll take a look at it later

-> @SanjulaGanepola 
yes, that bit actually remained from my various tests.

sebjulliand:
@buzzia2001 just the block background color. I don't think you can change the whole bar unless you create a Them which is not the point here 😉

buzzia2001:
> @buzzia2001 just the block background color. I don't think you can change the whole bar unless you create a Them which is not the point here 😉

Hi @sebjulliand,
bad and good news for you...

The color picker works even though its interface isn't very appealing. The color reload also works while the connection is active.
The bad news is that it seems VS Code doesn’t handle colors in individual blocks in the status bar. All the information I’ve found refers to highlighting (standard templates) for errors or warnings, which does indeed turn the cell red or yellow (but causes the text color to disappear).

image
image

buzzia2001:
@sebjulliand @SanjulaGanepola I've uploaded a new version with color picker only for text not for background because of vscode limitations

SanjulaGanepola:
> The color picker works even though its interface isn't very appealing

Yeah I was surprised by this as well especially cuz it doesn't look to respect the color theme. But this does make it much easier to customize.

image

One suggestion would be to not have the bar goes the full width:

image

image

image

buzzia2001:
@SanjulaGanepola to you 
image

SanjulaGanepola (approved):
Awesome!

PR #3362, opened 2026-07-14
updated processEvent to process dispatching of events in parallel

supravi96:
### Changes
Currently once event is fired then order in which notifications are dispatched is sequential so some subscribers face delay in receiving notifications . Hence this pr will provide the changes that will make the processing parallelly not just one or two subscribers facing delay. This will resolve issue https://github.ibm.com/bob-for-ibm-i/vscode-ibmi-bob/issues/679

### Checklist

* [x] have tested my change

### DIFF
--- a/src/Instance.ts
+++ b/src/Instance.ts
@@ -298,23 +298,25 @@ export default class Instance {
   }
 
   async processEvent(event: IBMiEvent) {
-    const eventSubscribers = this.getSubscribers(event)
+    const eventSubscribers = this.getSubscribers(event);
     console.time(event);
-    for (const [identity, callable] of eventSubscribers.entries()) {
-      try {
-        console.time(identity);
-        await callable.func();
-        console.timeEnd(identity);
-      }
-      catch (error) {
-        console.error(`${event} event function ${identity} failed`, error);
-      }
-      finally {
-        if (callable.transient) {
-          eventSubscribers.delete(identity);
+    await Promise.allSettled(
+      Array.from(eventSubscribers.entries()).map(async ([identity, callable]) => {
+        try {
+          console.time(identity);
+          await callable.func();
+          console.timeEnd(identity);
         }
-      }
-    }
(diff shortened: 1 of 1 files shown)


sebjulliand:
Having this loop to be sequential was made on purpose, originally because having too many parallel requests sent on the ssh channel would cause "ssh channel error" issues.

With v3 and the introduction of Mapepire for SQL and CL commands, I reckon this won't happen as often. But still, if the event processing becomes sequential, we need a queue system for the ssh commands to prevent some events from crashing.

SanjulaGanepola:
@sebjulliand For context, Supravi was discussing this one with me because the team noticed delays in other extensions that were subscribing to the `connected` event. With the current implementation, I guess this is because there is a growing number of extensions subscribing to this event and so extensions that get registered late, may find some slow down given we are doing this sequentially.

> queue system for the ssh commands

If we did this, I wonder if this would actually slow things down for certain users which would not be good.

As of right now, we don't make use of `Pools` in Mapepire. Have we considered doing this? Or does that introduce more issues given that then we will have multiple SQL jobs.

sebjulliand:
@SanjulaGanepola Mapepire is not the issue here; it has its own queue and there is no concurrency problem with it in Code for i as far as I can tell.

So as long as subscribers run CL commands/SQL queries, it won't be a problem running everything in parallel, the queue will take care of handling that.

The "Can't open SSH channel" issue seems to happen on some systems, maybe based on sshd settings. It happens for me when Mapepire is started, so...at a very early stage 🤣 

I think haveing this PR tested and merged would be fine, the risk is limited.

Then we should take a look at the channel thing separatly.

SanjulaGanepola:
@supravi96 I am currently looking into investigating the use of Mapepire Pools in Code4i which impact this. I will revisit this PR afterwards

SanjulaGanepola:
Closing this for now. We will re-visit this

PR #3329, opened 2026-06-24
Use a single method to get or prompt for connection's password

sebjulliand:
### Changes

Resolves #3327 

This PR provides a single internal method to retrieve the connection stored password or prompt and temporary store it for the duration of the connection if there's no stored password.

This avoids prompting multiple times for the password in case it's not stored or if the connection uses a private key or an SSH agent.

### How to test this PR

1. Update connection settings to connect to a Mapepire server
2. Connect using a private key
3. A prompt to provide the password for Mapepire should show up once
4. The connection must succeed
5. Db2 for i connection should not prompt for a password
6. Starting another Db2 for i DB job should not prompt for a password

### Checklist

* [x] have tested my change

PR #3366, opened 2026-07-16
Clear temporary password when actually disconnecting

sebjulliand:
### Changes

This PR follows up on #3329.
The temporary password used to be cleared in response to the `disconnected` event. This would clear it unconditionally, even if the connection dropped, as reported here https://github.com/codefori/vscode-ibmi/issues/3327#issuecomment-4986756891 by @SJLennon .

This PR changes this behavior and clears the temporary password only when an actual disconnection occurs. Reconnecting after a connection drop will not ask for the password again. 

### How to test this PR

1. Connect to a system with a private key or without saving the password
2. Fill in the prompt when asked for a password
3. Once connected, kill your connection (pull the plug or turn off WiFi)
4. Wait for the reconnection dialog to show up
5. Clicking on reconnect must not ask for the password again

### Checklist

* [x] have tested my change


buzzia2001:
It works fine, thank you @sebjulliand

PR #3374, opened 2026-07-21
Parameters for managing table rows and the auto-refresh threshold

buzzia2001:
### Changes
This PR introduces two parameters that are used to customize the number of pages displayed in a table, as well as the page refresh interval.

### How to test this PR
Open vscode and go to preferences, you'll find new configs like:
image
image

Change parameters and open joblog view to test the change.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/configuration/config/types.ts
+++ b/src/api/configuration/config/types.ts
@@ -48,6 +48,30 @@ export interface ConnectionConfig extends ConnectionProfile {
   [name: string]: any;
 }
 
+// Global `code-for-ibmi.*` settings not related to a single connection
+export interface GlobalConfiguration {
+  recentlyOpenedFilesLimit: number;
+  defaultOpenMode: DefaultOpenMode;
+  autoReconnect: ReconnectMode;
+  grepIgnoreDirs: string[];
+  createLibraryOnBranchChange: boolean;
+  clearDiagnosticOnEdit: boolean;
+  clearErrorsBeforeBuild: boolean;
+  'tables.itemsPerPage': number;
+  'views.autoRefreshInterval': number;
+  postActionView: 'task' | 'problems' | 'none';
+  clearOutputEveryTime: boolean;
+  sortActionsBy: 'usage' | 'name' | 'config';
+  'IfsBrowser.DragAndDropDefaultBehavior': 'ask' | 'copy' | 'move';
+  autoRefresh: boolean;
+  safeDeleteMode: boolean;
+  'ObjectBrowser.showNamesInLowercase': boolean;
+  'ObjectBrowser.sortObjectsByName': boolean;
+  autoOpenFile: boolean;
+  'terminals.5250.openInEditorArea': boolean;
+  'terminals.pase.openInEditorArea': boolean;
+}
+
 export interface RemoteConfigFile {
   codefori?: Partial<ConnectionConfig>;
 }
--- a/src/api/configuration/config/types.ts
+++ b/src/api/configuration/config/types.ts
@@ -1,5 +1,5 @@
 import { FilterType } from "../../Filter";
-import { ConnectionData, DeploymentMethod } from "../../types";
+import { Action, ConnectionData, DeploymentMethod } from "../../types";
 
 export type DefaultOpenMode = "browse" | "edit";
 export type ReconnectMode = "always" | "never" | "ask";
@@ -60,7 +60,9 @@ export interface GlobalConfiguration {
   'tables.itemsPerPage': number;
   'views.autoRefreshInterval': number;
   postActionView: 'task' | 'problems' | 'none';
+  logCompileOutput: boolean;
   clearOutputEveryTime: boolean;
+  persistOutputOnConnect: boolean;
   sortActionsBy: 'usage' | 'name' | 'config';
   'IfsBrowser.DragAndDropDefaultBehavior': 'ask' | 'copy' | 'move';
   autoRefresh: boolean;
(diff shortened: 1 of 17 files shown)


src/config/Configuration.ts
```diff
@@ -11,6 +11,46 @@ export function onCodeForIBMiConfigurationChange<T>(props: string | string[], to
...
+  /** Fallback interval, in seconds, when the setting is unset or unusable. */
+  const DEFAULT_AUTO_REFRESH_SECONDS = 30;
+
+  /**
+   * Page size for every paginated table, from `code-for-ibmi.tables.itemsPerPage`.
+   * Callers that paginate server-side must use this for their own LIMIT/OFFSET too,
+   * otherwise the page count shown by the table won't match the rows it receives.
+   */
+  export function getItemsPerPage(): number {
+    const configured = vscode.workspace.getConfiguration(`code-for-ibmi`).get<number>(`tables.itemsPerPage`);
```
SanjulaGanepola: In the rest of the codebase, we use `IBMi.connectionManager.get` and we should add it to `ConnectionConfig`

src/config/Configuration.ts
```diff
@@ -11,6 +11,46 @@ export function onCodeForIBMiConfigurationChange<T>(props: string | string[], to
...
+    }
+    return Math.max(MIN_ITEMS_PER_PAGE, Math.floor(configured));
+  }
+
+  /**
+   * Auto-refresh interval in milliseconds, from `code-for-ibmi.views.autoRefreshInterval`
+   * (which is expressed in seconds). Returns 0 when auto-refresh is disabled.
+   */
+  export function getAutoRefreshInterval(): number {
+    const configured = vscode.workspace.getConfiguration(`code-for-ibmi`).get<number>(`views.autoRefreshInterval`);
```
SanjulaGanepola: Same as above, can we use `IBMi.connectionManager.get`?

src/ui/frontendTables.ts
```diff
@@ -1,11 +1,19 @@
...
 export namespace FrontendTables {
 
   const EMPTY_VALUE_PLACEHOLDER = '<span style="color: var(--vscode-descriptionForeground); font-style: italic;">—</span>';
 
+  /**
+   * Page size used by every paginated table.
+   * @see ViewSettings.getItemsPerPage — re-exported here so callers already using this
+   * namespace don't need a second import.
+   */
+  export const getItemsPerPage = ViewSettings.getItemsPerPage;
```
SanjulaGanepola: I don't quite understand the need for this. If we did this, we would need to just export every func so I think just using `ViewSettings` directly is fine. What do you think?

src/webviews/wrkjob/index.ts
```diff
@@ -79,6 +78,23 @@ export namespace JobLogUI {
...
+      }),
+      onCodeForIBMiConfigurationChange(`tables.itemsPerPage`, async () => {
+        // The page size is baked into the webview's script, so the whole page has to
+        // be rebuilt. The current page number no longer means the same thing either.
+        for (const [jobName, state] of activePanels) {
+          state.currentPage = 1;
+          try {
+            await render(state.panel, jobName);
+          } catch (error) {
+            console.error(`Job log re-render error:`, error);
```
SanjulaGanepola: Do we instead want to surface this error to the user? `window.showErrorMessage` with a human readable message?

SanjulaGanepola (changes requested):

src/webviews/wrkjob/index.ts
buzzia2001: Replaced console.log with vscode error message

src/ui/frontendTables.ts
buzzia2001: removed useless wrapper

src/config/Configuration.ts
buzzia2001: replaced call

src/config/Configuration.ts
buzzia2001: replaced call and export of global config

buzzia2001:
Hi @SanjulaGanepola, I've aligned the call with the rest of the package as you requested. I've also exported the global configurations that aren't specific to a single connection (I took them from the JSON package; there might be some we won't use, but just in case, they're already set up). Finally, I've replaced the console log with a VS Code error message (including the corresponding translations).

src/api/configuration/config/types.ts
```diff
@@ -48,6 +48,30 @@ export interface ConnectionConfig extends ConnectionProfile {
...
+  sortActionsBy: 'usage' | 'name' | 'config';
+  'IfsBrowser.DragAndDropDefaultBehavior': 'ask' | 'copy' | 'move';
+  autoRefresh: boolean;
+  safeDeleteMode: boolean;
+  'ObjectBrowser.showNamesInLowercase': boolean;
+  'ObjectBrowser.sortObjectsByName': boolean;
+  autoOpenFile: boolean;
+  'terminals.5250.openInEditorArea': boolean;
+  'terminals.pase.openInEditorArea': boolean;
+}
```
SanjulaGanepola, line 76: > there might be some we won't use, but just in case, they're already set up

I just had a look and these 3 were missing. For completeness, lets add them 🙂

- `logCompileOutput`
- `persistOutputOnConnect`
- `actions`

SanjulaGanepola (approved):
Just one minor comment about the types and then we are good to merge

buzzia2001:
Thank you @SanjulaGanepola for your code review

PR #3380, opened 2026-07-27
Add member text description filter to member list

### LINKED ISSUE #2683: Filter member list by text

bobcozzi:
## Summary

Adds a `memberText` filter field to the Object Browser filter system, allowing users to filter source file members by their text description — the same field displayed as "Member Text" or "Description" in IBM i tools like PDM.

## How It Works

The implementation follows the exact same two-phase pattern already used by the `memberType` (source type) filter:

1. **SQL-level optimization** — When the filter is a single simple pattern (e.g. `PAY*` or `*RPG*`), a `WHERE` clause is pushed down into the `PARTITION_STATISTICS` query to reduce rows returned from IBM i:
   ```sql
   AND UPPER(RTRIM(VARCHAR(PART_STAT.TEXT))) like 'PAY%' escape '+'
   ```
   `UPPER()` is applied on both sides for case-insensitive matching (unlike `memberType` / `member` which are always uppercase system names).

2. **Client-side filtering** — The full `parseFilter()` logic (supporting comma-separated multi-generic values and regex) is applied to `member.text` after the SQL result is received, consistent with how `memberFilter` and `memberExtensionFilter` work.

## Changes

| File | What changed |
|---|---|
| `src/api/configuration/config/types.ts` | Added `memberText?: string` to `ObjectFilters` (optional for backward compat) |
| `src/api/IBMiContent.ts` | Added `memberText` param to `getMemberList`, SQL LIKE clause, and client-side filter chain |
| `src/webviews/filters/index.ts` | Added "Member text" input to the filter editor UI; default value `*` |
| `src/ui/views/objectBrowser.ts` | Passes `filter.memberText` through to `getMemberList` |
| `package.json` | Added `memberText` property to the `objectFilters` JSON schema (not required, for backward compat) |

## Similarity to Source Type Filter

`memberText` is the direct parallel of `memberType` (`PART_STAT.SOURCE_TYPE`):

| Field | Filters on | SQL column | Case handling |
|---|---|---|---|
| `memberType` | Source type (e.g. `RPGLE`, `CLLE`) | `PART_STAT.SOURCE_TYPE` | Always uppercase (system names) |
| `memberText` | Text description | `PART_STAT.TEXT` | `UPPER()` for case-insensitive match |

Existing saved filter configs without `memberText` are unaffected — `undefined` maps to `noFilter: true` via `parseFilter`, passing all members through.

Close #2683

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -776,7 +776,7 @@ export default class IBMiContent {
         WHERE TRIM(PART_STAT.SYSTEM_TABLE_MEMBER) <> ''
         ${singleMember ? `AND RTRIM(PART_STAT.SYSTEM_TABLE_MEMBER) like '${singleMember.replaceAll('_', '+_')}' escape '+'` : ``}
         ${singleMemberExtension && singleMemberExtension.trim() !== '%' ? `AND RTRIM(CAST(PART_STAT.SOURCE_TYPE AS VARCHAR(10))) like '${singleMemberExtension.replaceAll('_', '+_')}' escape '+'` : ``}
-        ${singleMemberText && singleMemberText.trim() !== '%' ? `AND UPPER(RTRIM(VARCHAR(PART_STAT.TEXT))) like '${singleMemberText.replaceAll('_', '+_')}' escape '+'` : ``}
+        ${singleMemberText && singleMemberText.trim() !== '%' ? `AND UPPER(RTRIM(VARCHAR(PART_STAT.TEXT))) like '${singleMemberText.replaceAll('+', '++').replaceAll('_', '+_').replaceAll(`'`, `''`)}' escape '+'` : ``}
         ORDER BY ${sort.order === 'name' ? 'NAME' : 'CHANGED'} ${!sort.ascending ? 'DESC' : 'ASC'}`;
 
     const results = await this.ibmi.runSQL(statement);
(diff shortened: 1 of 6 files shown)


src/api/IBMiContent.ts
```diff
@@ -759,6 +759,9 @@ export default class IBMiContent {
     const memberExtensionFilter = parseFilter(filter.extensions, filter.filterType);
     const singleMemberExtension = memberExtensionFilter.noFilter && filter.extensions && !filter.extensions.includes(",") ? this.ibmi.upperCaseName(filter.extensions).replace(/[*]/g, `%`) : undefined;
 
+    const memberTextFilter = parseFilter(filter.memberText, filter.filterType);
+    const singleMemberText = memberTextFilter.noFilter && filter.memberText && !filter.memberText.includes(",") ? filter.memberText.toUpperCase().replace(/[*]/g, `%`) : undefined;
```
buzzia2001, line 763: Hi @bobcozzi, maybe we need to improve our replace logic to include ' and + that will break the query, in comments could happen to see this chars

buzzia2001 (changes requested):

src/api/IBMiContent.ts
bobcozzi: All set. Thanks for the suggestion!

buzzia2001 (approved):
Thank you @bobcozzi

PR #3388, opened 2026-07-29
Fix CCSID 65535 source member CAST to use IBM i Default Job CCSID

bobcozzi:
## Problem

Users with `QCCSID = 65535` (`*HEX`) and source dates enabled were getting `cast(srcdta as varchar(n) CCSID 65535)` in the member download SQL. Casting to CCSID 65535 is a no-op and leaves Mapepire with raw EBCDIC bytes.

## Root cause

Two bugs in `IBMi.ts`:

**1. Quick-connect bypassed ACTIVE_JOB_INFO resolution**
The quick-connect condition allowed a cached `jobCcsid = 65535` to be restored directly, skipping the `ACTIVE_JOB_INFO` lookup on subsequent connections.

**2. ACTIVE_JOB_INFO fallback silently kept 65535**
The ternary left `userJobCcsid` at 65535 if `DEFAULT_CCSID` was falsy, rather than trusting the IBM i-guaranteed value.

## IBM i CCSID hierarchy

QCCSID (system value) and Job CCSID can both be 65535. Default Job CCSID (ACTIVE_JOB_INFO.DEFAULT_CCSID) is NEVER 65535 — it is the value IBM i uses when the job CCSID is *HEX.

## Changes

- IBMi.ts: quick-connect now excludes cached 65535 from the fast path; ACTIVE_JOB_INFO result is trusted directly
- extendedContent.ts: getCcsid() called cleanly, relying on the resolved value

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -638,7 +638,9 @@ export default class IBMi {
         // TODO: since we are using Mapepire, we only need the QCCSID and the job CCSID now
 
         // Fetch conversion values?
-        if (quickConnect() && cachedServerSettings?.jobCcsid !== null && cachedServerSettings?.qccsid) {
+        // Do not restore a cached CCSID_NOCONVERSION value via quick-connect: it must go through
+        // the full resolution path (ACTIVE_JOB_INFO) to obtain a usable CCSID.
+        if (quickConnect() && cachedServerSettings?.jobCcsid !== null && cachedServerSettings?.jobCcsid !== IBMi.CCSID_NOCONVERSION && cachedServerSettings?.qccsid) {
           this.qccsid = cachedServerSettings.qccsid;
           this.userJobCcsid = cachedServerSettings.jobCcsid;
         } else {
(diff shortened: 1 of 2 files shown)


bobcozzi:
I created a test file for you to confirm that the fix works. once you confirm it does or does not help, you can delete it and revert back to 3.0.12
It is a ZIP file, just unzip it and load it via "Install VSIX from file" via the `...` menu on the Extensions Panel.

[code-for-ibmi-3.0.13-dev.0.vsix.zip](https://github.com/user-attachments/files/30520751/code-for-ibmi-3.0.13-dev.0.vsix.zip)

src/api/IBMi.ts
```diff
@@ -671,15 +673,19 @@ export default class IBMi {
...
                 select DEFAULT_CCSID from table(
                   QSYS2.ACTIVE_JOB_INFO(
                     JOB_NAME_FILTER => '*',
                     DETAILED_INFO => 'WORK'
                   )
                 )`);
-              this.userJobCcsid = row?.DEFAULT_CCSID ? Number(row.DEFAULT_CCSID) : this.userJobCcsid;
+              if (row?.DEFAULT_CCSID) {
+                this.userJobCcsid = Number(row.DEFAULT_CCSID);
+              }
```
sebjulliand, line 688: This doesn't seem necessary since it does what was done before 😅

sebjulliand (changes requested):
Unless I'm wrong, one of these changes is not necessary. It's just replacing a ternary with an if, but the result is the same (but please correct me if I'm wrong).

Forcing a refresh if the cached value is 65535 is great!

src/api/IBMi.ts
bobcozzi: I guess it is a matter of taste. Yes, I did convert it to an 'if' because I had changed it another way that did not work as I intended, so I deleted that fragment. Leaving it in the shape it is. But if you want it back at a ternary assignment, that's okay but not necessary.

sebjulliand: Fine by me!

sebjulliand (approved):

PR #3392, opened 2026-07-31
Fix extraction of `sqlstate` when using `runSQL` API

SanjulaGanepola:
### Changes

Currently, `runSQL` has logic to extract the `sqlstate`, but this would only work if the message also contained a `,`. This is not always the case such as for `[SQL0206] Column or global variable COPYRIGHT_STRINGS not found., 42703, -206`.

### How to test this PR
1. Run the test cases

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -1486,9 +1486,16 @@ export default class IBMi {
             error = new Tools.SqlError(e.message);
             error.cause = statement
 
-            const parts: string[] = e.message.split(`,`);
-            if (parts.length >= 3) {
-              error.sqlstate = parts[parts.length - 2].trim();
+            // First attempt to extract sqlstate using a regex
+            const sqlstateMatch = /,\s*([0-9A-Z]{5}),\s*-?\d+\s*$/.exec(e.message);
+            if (sqlstateMatch) {
+              error.sqlstate = sqlstateMatch[1];
+            } else {
+              // Fallback to splitting the message by commas and taking the second to last part
+              const parts: string[] = e.message.split(`,`);
+              if (parts.length >= 3) {
+                error.sqlstate = parts[parts.length - 2].trim();
+              }
             }
 
             this.appendOutput(`${log}-> Failed: ${error.sqlstate ? `[${error.sqlstate}] ` : ''}${error.message}`);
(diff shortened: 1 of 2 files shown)


buzzia2001 (changes requested):
To be honest, I'd rather avoid splitting by commas and switch to a regex instead... Assuming errors are formatted like this: “, SQLSTATE, SQLCODE” (e.g., “Column X not found., 42703, -206”), I'll use /,\s*([0-9A-Z]{5}),\s*-?\d+\s*$/

Test case was failed in an italian partition:
 FAIL  suites/content.test.ts > Content Tests > Test runSQL (column not found)
AssertionError: expected false to be truthy

- Expected
+ Received

- true
+ false

 ❯ suites/content.test.ts:253:61
    251|       expect.fail('Should have thrown an error');
    252|     } catch (e: any) {
    253|       expect(e.message.endsWith('not found., 42703, -206')).toBeTruthy();
       |                                                             ^
    254|       expect(e.sqlstate).toBe('42703');
    255|     }

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯


 Test Files  1 failed (1)
      Tests  1 failed | 4 passed | 43 skipped (48)
   Start at  21:56:00
   Duration  14.99s (transform 113ms, setup 0ms, import 217ms, tests 14.45s, environment 0ms)

buzzia2001 (approved):
It works, thank you @SanjulaGanepola

PR #3402, opened 2026-08-05
Fix status bar color ownership and remove forbidden cursor in Connection Settings

bobcozzi:
## Summary
Fixes a shipped UX/configuration issue in Connection Settings where **Status bar color** could appear server-managed (read-only) and show a forbidden cursor, even though it is intended to be a client-side preference.

## Problem
The new Status bar color option had two user-facing problems:

1. The field could be treated as server-managed and become read-only in the settings webview.
2. The color control could show a forbidden cursor / awkward interaction behavior in the custom UI.

### 1) Ownership/override mismatch
`statusBarColor` is a local, per-connection client preference, but remote settings handling used generic logic:

- During connection bootstrap, every key under `remoteConfig.codefori` was applied to local config when names matched.
- In the settings webview, any field key present in `serverConfig.codefori` was marked read-only.

That meant if `statusBarColor` appeared in remote settings payloads, it was incorrectly treated like a server-managed setting.

### 2) Color input rendering path
The shared custom webview input renderer treated color as a generic `vscode-textfield` path, which produced poor cursor/input behavior for this control.

### Client-only ownership guard
Added a client-only exclusion for `statusBarColor` in both places where server-managed logic was applied:

- Exclude from remote config merge into local runtime config.
- Exclude from server-driven read-only field locking in settings UI.

### Color control rendering fix
Updated `CustomUI` input rendering for `inputType: "color"` to use a native `<input type="color">` control and improved update handling safety for inputs without `shadowRoot`.

## Files changed
- `src/api/IBMi.ts`
- `src/webviews/settings/index.ts`
- `src/webviews/CustomUI.ts`

## Behavioral impact
- **Status bar color remains editable** in Connection Settings as intended.
- The setting is no longer overridden by remote/system settings payloads.
- Color picker interaction is normal (no forbidden cursor behavior for editable field).

## Validation
- Webpack build succeeded after changes.
- TypeScript diagnostics for edited files report no errors.

## Notes
This keeps server-managed behavior intact for legitimate system-governed fields, while explicitly preserving `statusBarColor` as client-owned UX configuration.


bobcozzi:
Root-cause note: reproducibility depended on config shape.

The lock happened only when remote settings exposed `codefori.statusBarColor`, because generic server-managed logic marked matching keys readonly and merged them into runtime config.

Users without that remote key (or without a loaded remote settings file) would not see the lock.

A separate UI rendering issue in the custom color control caused misleading cursor behavior in some environments.

This PR fixes both by treating `statusBarColor` as client-only and using a native color input path.

buzzia2001:
Hi @bobcozzi
let me understand better... were you trying to connect to an host with color configured server side?

buzzia2001 (changes requested):
Hi @bobcozzi to be honest, I suppose there are other settings that in my opinion should be client only (e.g. libary descriptions ecc)
Please use this pr to put also them

bobcozzi:
> Hi @bobcozzi let me understand better... were you trying to connect to an host with color configured server side?

I was not relying on server-side settings in my own environment. The bug was that the generic server-settings ownership logic could incorrectly treat UI-only/client-local settings as server-managed if those keys appeared in the remote payload. I’ve now expanded the client-only exclusion beyond status bar color to cover other clearly local display and terminal preferences as well. Let me know if you think I missed any of them.

bobcozzi:
> Hi @bobcozzi to be honest, I suppose there are other settings that in my opinion should be client only (e.g. libary descriptions ecc) Please use this pr to put also them

Done!

buzzia2001 (approved):

buzzia2001:
Thank you @bobcozzi

src/api/IBMi.ts
```diff
@@ -74,6 +74,23 @@ interface ConnectionConfigFiles {
...
+  `statusBarColor`,
+  `showDescInLibList`,
+  `showHiddenFiles`,
+  `autoSortIFSShortcuts`,
+  `autoSaveBeforeAction`,
+  `sourceDateGutter`,
+  `encodingFor5250`,
+  `terminalFor5250`,
+  `setDeviceNameFor5250`,
+  `connectringStringFor5250`,
```
SanjulaGanepola, line 91: @bobcozzi @buzzia2001 I just noticed an inconsistency here when I was testing a different PR. We currently list `encodingFor5250`, `terminalFor5250`, and `connectringStringFor5250` here as client only settings. However, they are still listed in `vscode-ibmi\schemas\settings.json` meaning they are shown to users when they invoke content assist. 

Two questions/suggestions:
1. Curious why these are client only settings. 
2. If they are indeed client only, we should drop them from the schemas

buzzia2001: @SanjulaGanepola encoding depends on the client, so if you a partition multi-national you can have multiple clients with different ecnoding... Terminal is something that may vary depending on the user's preferences... Actually, perhaps the connection string and the use of the terminal can be specified on the server side, so I'll quickly create a PR to update the schema and the “client-only” entries.

PR #3409, opened 2026-08-11
Add support for member locking

SanjulaGanepola:
### Changes

This PR adds support for member locking using a shared-update lock (`*SHRUPD`) that is allocated on reads and deallocated when we close the editor. This is an opt in setting meaning it is disabled by default.

This has been requested in https://github.com/codefori/vscode-ibmi/issues/1688 and https://github.com/codefori/vscode-ibmi/issues/2409

### How to test this PR
1. Debug the extension using this branch.
1. Navigate to the connection settings and enable `Member Locking` under the `Source Code` tab.
2. Open a member.
3. Launch a separate VS Code window running the release version of Code4i.
4. Attempt to open the same member in this new window and observe the member is locked.
5. In the first window, disable the setting.
6. In the second window, reattempt to open the member and observe that it can be opened.

### Checklist
* [ ] have tested my change

### DIFF
--- a/src/api/configuration/config/ConnectionManager.ts
+++ b/src/api/configuration/config/ConnectionManager.ts
@@ -39,6 +39,7 @@ function initialize(parameters: Partial<ConnectionConfig>): ConnectionConfig {
     readOnlyMode: (parameters.readOnlyMode === true),
     quickConnect: (parameters.quickConnect === true || parameters.quickConnect === undefined),
     defaultDeploymentMethod: parameters.defaultDeploymentMethod || ``,
+    lockMembers: (parameters.lockMembers === true),
     protectedPaths: (parameters.protectedPaths || []),
     showHiddenFiles: (parameters.showHiddenFiles === true || parameters.showHiddenFiles === undefined),
     secureSQL: (parameters.secureSQL === true),
--- a/src/api/configuration/config/types.ts
+++ b/src/api/configuration/config/types.ts
@@ -34,6 +34,7 @@ export interface ConnectionConfig extends ConnectionProfile {
   readOnlyMode: boolean;
   quickConnect: boolean;
   defaultDeploymentMethod: DeploymentMethod | '';
+  lockMembers: boolean;
   protectedPaths: string[];
   showHiddenFiles: boolean;
   secureSQL: boolean;
--- a/src/api/memberLocks.ts
+++ b/src/api/memberLocks.ts
@@ -1,5 +1,13 @@
 import IBMi, { MemberParts } from "./IBMi";
 
+export enum LockState {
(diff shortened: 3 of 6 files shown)


SanjulaGanepola:
@codefori/core Before proceeding with more testing of this PR, any major issues with support member locking?

buzzia2001:
Hi @SanjulaGanepola,
There is something that is not ok for me...
With *SHRRD any other job can take a lock for that specific member and in that case I'm not able to change it from vscode:

In my example I've opened a member from vscode (job 682461/QUSER_NC/QZDASOINIT) and from a 5250 session with the same command:
image

Now, if I change a row and I try to save it, it doesn't work because the member is locked by my other job:

image
image

_My suggestion is to use EXCLUSIVE_ALLOW_READ instead of SHARED_UPDATE. In this way, we can be sure that no one can steal our lock._

<hr>

Another issue I had is that the object is locked even if I'm not connected anymore:
image
image
image
image

_Without Mapepire, my suggestion is to use *THREAD instead of *JOB in lock scope. In this way, when your connection end, the thread is closed and the lock released. But with Mapepire is not so easy, it seams that the lock survive also to the client disconnection. We need to analyze better this issue._


<hr>

Currently, there are no safeguards in place when opening members with defined locks; this means that a user can open a member, attempt to modify it, and then get stuck when trying to save... Ideally, you should check for locks before opening the member, so that it opens in view-only mode if locks already exist. You could manipulate the WRKOBJLCK spool or use the SQL service, but be careful not to make this operation take too long... 

<hr>

At the moment I can't see anything related to IFS, anyway I think that also IFS should be covered (IDK idf in this pr or in another one).

Balrocj:
Hi @SanjulaGanepola, I opened #3446 with another member-locking implementation. @sebjulliand suggested we compare notes and align on a unified approach.

The main differences in #3446 are:

- Uses `*EXCLRD` with `WAIT(0)`.
- Detects `CPF1002` before opening an editable editor and offers an **Open read only** fallback.
- Surfaces other IBM i or Mapepire lock-acquisition errors with their diagnostic details instead of treating them as lock conflicts.
- Tracks locks independently per member, including multiple members in one source physical file.
- Releases locks on document close and disconnect, and restores active locks after reconnecting.
- Uses the existing persistent Mapepire SQL job and serializes `IBMi.runSQL()` calls to avoid overlapping job-scoped work.
- Keeps Source Date Support saves in that same job when a member lock is active.
- Includes unit tests, focused IBM i integration coverage, and CI integration support.

IFS streamfiles are intentionally out of scope in #3446.

I would be glad to compare the approaches and help converge on the direction that best fits the project.

PR #3411, opened 2026-08-13
Validate if library in system portion before adding to user portion in libl

SanjulaGanepola:
### Changes

Recently Scott had noticed that if you add a system library to the user library list (which Code4i does not prevent), it ends up breaking filters listing source physical files:

image

This PR addresses this by:
1. Adding a check to validate any libraries being added to the user library list are not in the system portion.
2. Adding a red error decoration for any existing system libraries that have been added.

### How to test this PR
1. Attempt to add a system library to Code4i's user library list.
2. Manually add a system library (via settings.json), refresh the user library list view, and validate that it is highlighted in red.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -535,6 +535,21 @@ export default class IBMiContent {
   }
 
   /**
+   * Returns the names of libraries currently in the system portion of the library list.
+   * @returns Array of library names in the system portion
+   */
+  async getSystemLibraries(): Promise<string[]> {
+    try {
+      const result = await this.ibmi.runSQL(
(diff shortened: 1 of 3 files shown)


buzzia2001 (changes requested):
Hi @SanjulaGanepola we should protect profiles

buzzia2001:
@SanjulaGanepola can u review it?

SanjulaGanepola:
@buzzia2001 I made one minor change. All else looks good. Thanks for those changes.

buzzia2001 (approved):

buzzia2001:
@SanjulaGanepola feel free to merge

PR #3413, opened 2026-08-13
Read server configuration on every connection

buzzia2001:
### Changes
This PR adds the step of reading the server configuration file /etc/vscode/settings.json on every connection, rather than just running **CONNECT AND RELOAD SERVER SETTINGS**.
Additionally, it prevents mapepire files from being uploaded when used in server mode.

### How to test this PR
Connect to a system, try editing the configuration file /etc/vscode/settings.json, reconnect. The change should have been applied without having to reload.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -190,6 +190,27 @@ export default class IBMi {
     }
   }
 
+  /**
+   * loads the remote configuration files /etc/vscode/settings.json and applies
+   * runs on every connection
+   */
+  private async loadAndApplyRemoteConfigs() {
+    await this.loadRemoteConfigs();
+
+    const remoteConnectionConfig = this.getConfigFile<RemoteConfigFile>(`settings`);
+    if (this.config && remoteConnectionConfig.getState() === `ok`) {
+      const remoteConfig = await remoteConnectionConfig.get();
+
+      if (remoteConfig.codefori) {
+        for (const [key, value] of Object.entries(remoteConfig.codefori)) {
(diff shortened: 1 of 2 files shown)


src/api/components/mapepire/index.ts
```diff
@@ -66,6 +66,11 @@ export class Mapepire implements IBMiComponent {
 
 
   async update(connection: IBMi): Promise<SecureComponentState> {
+    if (connection.getConfig().mapepireUseServer) {
+      //no upload the remote JAR file in server mode
```
SanjulaGanepola:
```suggestion
      //No need to upload the remote JAR file in server mode
```

src/api/IBMi.ts
```diff
@@ -839,24 +863,6 @@ export default class IBMi {
 
       this.appendOutput(`\n`);
 
-      // Load the remote connection configuration and apply it to the connection
-
-      callbacks.progress({ message: `Loading remote configuration files.` });
```
SanjulaGanepola, line 844: In my testing using the main branch, this logic seemed to already load the remote configs every time.

SanjulaGanepola (changes requested):

buzzia2001 (commented):
Updated as requested

src/api/IBMi.ts
SanjulaGanepola: Confirmed with @buzzia2001 that this is indeed the case, but we will proceed with this change as it is better to call this logic earlier.

SanjulaGanepola (approved):

PR #3414, opened 2026-08-13
Create helper for check routine signature

buzzia2001:
### Changes
This PR add an helper to get signature from a FUNCTION or a PROCEDURE using test defined environment.

### How to test this PR
1. Ensure you have src\api\tests\.env properly filled
2. Run from the basepath this command: `npx tsx tools/sqlSignature CODE4I.VALIDATE_STATEMENT0002`, it will return you the correct signature to be hardcoded:
image

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/IBMi.ts
+++ b/src/api/IBMi.ts
@@ -1240,6 +1240,8 @@ export default class IBMi {
   }
 
   async disconnect() {
+    await this.sqlJob?.close();
+
     if (this.client?.connection) {
       await (await this.getComponent<Mapepire>(Mapepire.ID))?.endJobs();
       //Close the connection and triggers its 'end' event
--- a/tools/sqlSignature.ts
+++ b/tools/sqlSignature.ts
@@ -1,0 +1,109 @@
+import { existsSync, readFileSync } from "fs";
+import path from "path";
+
+type RoutineType = "PROCEDURE" | "FUNCTION";
+
+const ENV_FILE = path.join(__dirname, `..`, `src`, `api`, `tests`, `.env`);
+
+function usage() {
+  console.log([
+    ``,
+    `Usage: npx tsx tools/sqlSignature <library>/<name> [PROCEDURE|FUNCTION]`,
+    ``,
+    `  <library>/<name>  the routine to look up; a dot also works (MYLIB.MYPROC).`,
+    `                    Names are uppercased unless "quoted".`,
+    `  [type]            when omitted, PROCEDURE and FUNCTION are both tried.`,
+    ``,
+    `Connection settings are read from ${path.relative(process.cwd(), ENV_FILE)}`,
+    `(same VITE_* variables as the test suite) or from the environment.`,
+    ``,
+  ].join(`\n`));
+}
+
+/**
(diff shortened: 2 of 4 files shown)


src/api/IBMi.ts
```diff
@@ -1240,6 +1240,8 @@ export default class IBMi {
   }
 
   async disconnect() {
+    await this.sqlJob?.close();
```
SanjulaGanepola, line 1243: In the code below this line we already call `dispose` which results in `sshSqlJob#end` to be called which ultimately closes the SQL Job. Did you notice a case where this didn't happen causing us to add this?

tools/sqlSignature.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  for (const line of readFileSync(ENV_FILE, `utf-8`).split(/\r?\n/)) {
+    const variable = /^\s*([\w.-]+)\s*=\s*(.*?)\s*$/.exec(line);
+    if (variable) {
+      const [, key, value] = variable;
+      if (process.env[key] === undefined && value) {
+        process.env[key] = value.replace(/^["'](.*)["']$/, `$1`);
+      }
+    }
+  }
```
SanjulaGanepola: What is the need for this logic? Shouldn't the env vars already be accessible via `process.env`? And if not, they may not be getting picked up in which case do we need to use like `dotenv`?

tools/sqlSignature.ts
```diff
@@ -0,0 +1,109 @@
...
+import path from "path";
+
+type RoutineType = "PROCEDURE" | "FUNCTION";
+
+const ENV_FILE = path.join(__dirname, `..`, `src`, `api`, `tests`, `.env`);
+
+function usage() {
+  console.log([
+    ``,
+    `Usage: npx tsx tools/sqlSignature <library>/<name> [PROCEDURE|FUNCTION]`,
```
SanjulaGanepola, line 11: We should add something like `"sqlsignature": "npx tsx tools/sqlSignature"` to the scripts section of package.json so we can just call it as `npm run sqlsignature ...`

SanjulaGanepola (changes requested):

tools/sqlSignature.ts
buzzia2001: Added inside package json, thank you for the suggestion

src/api/IBMi.ts
buzzia2001: Yes whitout that close the job stuck and doesn't close properly

buzzia2001 (commented):
@SanjulaGanepola to you

src/api/IBMi.ts
SanjulaGanepola: I tried it on my end with that line removed, and it did exit fine 🤔

Image

SanjulaGanepola (changes requested):

src/api/IBMi.ts
SanjulaGanepola: Was able to reproduce the issue on server mode

SanjulaGanepola (approved):

PR #2241, opened 2024-08-31
#2105 Improved error when running an action in readonly mode

sathisnsk:
Changes

Changed the error message to make it more contextual when running an action while browsing a member (read-only mode). 

How to test this PR

Examples:
1. Open a member in browse mode.
2. Run an action using the keyboard shortcut.
3. New error message will be displayed in the VSCode message notification area.

Checklist

* [x] have tested my change
* [ ] have created one or more test cases
* [ ] updated relevant documentation
* [ ] Remove any/all `console.log`s I added
* [ ] have added myself to the contributors' list in [CONTRIBUTING.md](https://github.com/codefori/vscode-ibmi/blob/master/CONTRIBUTING.md)

PR #3415, opened 2026-08-14
Fix #2105: show specific error when running an Action on a read only target

### LINKED ISSUE #2105: Improved error when running an action in readonly mode

christianlarsen:
### Changes
Shows a specific error message when an Action can't run because the target (member/streamfile/object) is protected or opened in read-only mode, instead of the generic "No suitable actions found" message.

This was originally fixed in #2241, but the fix was lost during the CompileTools refactor in ce5ac12c (Jan 2025) when this logic moved from `CompileTools.ts` to `src/ui/actions.ts` — the `isProtected` branch wasn't carried over.

Fixes #2105

### How to test this PR
1. Connect with `readOnlyMode` enabled (or open a member/file explicitly as read-only)
2. Try to run an Action on it
3. Confirm the error now says the target is read only, instead of "No suitable actions found"

### Checklist
- [x] have tested my change
- [ ] have created one or more test cases
- [ ] updated relevant documentation
- [x] Remove any/all `console.log`s I added
- [ ] have added myself to the contributors' list in CONTRIBUTING.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,3 +77,4 @@ Thanks so much to everyone [who has contributed](https://github.com/codefori/vsc
 * [@phpdave](https://github.com/phpdave)
 * [@neerup](https://github.com/neerup)
 * [@Frank-Hildebrandt](https://github.com/Frank-Hildebrandt)
+* [@christianlarsen](https://github.com/christianlarsen)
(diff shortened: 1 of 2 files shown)


buzzia2001 (approved):
Thank you @christianlarsen 

Please update CONTRIBUTING.md with your name, then I'll merge this pr

PR #3422, opened 2026-08-21
Use `await` to check IFS shortcuts sequentially

sebjulliand:
### Changes
The decorator responsible for warning if an IFS shortcut doesn't exist anymore was not awaiting when running the `isDirectory` method.

That could lead to the SSH connection opening too many channels at once and disrupt other operations happening during the `connected` event.

This PR makes sure the call to `isDirectory` is `awaited` so it doesn't overwhelm the SSH connection.

### How to test this PR

Check the decorator logic still works
1. Create a temporary IFS folder
2. Create a shortcut in the IFS browser for it
3. Delete the folder from a terminal
4. Reconnect
5. The folder must be decorated in red

### Checklist

* [x] have tested my change


buzzia2001 (approved):
Thank you @sebjulliand

PR #3424, opened 2026-08-21
Shared snippets

buzzia2001:
### Changes
This PR adds the ability to create custom code snippets for use; these snippets are tied to the IBM i system you are connected to. The metadata is located at /etc/vscode/Code4iRepo.json, while the source files are located at /etc/vscode/Code4iRepo.

### How to test this PR
You can try the steps shown in this video:

https://github.com/user-attachments/assets/56a2108c-a2d0-4bfd-b90d-329b79178a9f

### Checklist
* [x] have tested my change


SanjulaGanepola:
After further thought about this, I am wondering if this should be a feature that is implemented slightly differently and is actually in the Db2 extension and integrated with the `SQL Examples` view.

As requested in https://github.com/codefori/vscode-db2i/issues/526, users wanted to have reusable SQL examples stored in the IFS and easily accessible in the `SQL Examples` view. Couldn't this be implemented as easy as having a connection specific configuration to store the IFS directories (which the user also gets to define). Essentially, when the user goes to add a `Custom Examples Directory`, we can prompt them if it is in the IFS or Local. Wouldn't this be a much simpler implementation of this feature that also makes the scripts viewable in the SQL Examples view?

In my opinion, I also prefer to have these reusable SQL example queries browsable in the view rather than in the content assist as I don't always start with an editor. It is nice to just find the example I need and it brings up an untitled editor with the query just ready to run.

@buzzia2001 @sebjulliand What are your thoughts?

buzzia2001:
> After further thought about this, I am wondering if this should be a feature that is implemented slightly differently and is actually in the Db2 extension and integrated with the `SQL Examples` view.
> 
> 
> 
> As requested in https://github.com/codefori/vscode-db2i/issues/526, users wanted to have reusable SQL examples stored in the IFS and easily accessible in the `SQL Examples` view. Couldn't this be implemented as easy as having a connection specific configuration to store the IFS directories (which the user also gets to define). Essentially, when the user goes to add a `Custom Examples Directory`, we can prompt them if it is in the IFS or Local. Wouldn't this be a much simpler implementation of this feature that also makes the scripts viewable in the SQL Examples view?
> 
> 
> 
> In my opinion, I also prefer to have these reusable SQL example queries browsable in the view rather than in the content assist as I don't always start with an editor. It is nice to just find the example I need and it brings up an untitled editor with the query just ready to run.
> 
> 
> 
> @buzzia2001 @sebjulliand What are your thoughts?

Hi
This PR isn't specific to SQL; maybe I misunderstood the request for the SQL code repository… 
I mean, I see it more as something for RPG or CL… In any case, users can actually view the list of snippets, and with a good description, they can paste them directly from the connection menu…

buzzia2001:
If we want to add custom examples in db2 extension, we can open another pr

SanjulaGanepola:
> I mean, I see it more as something for RPG or CL… In any case, users can actually view the list of snippets, and with a good description, they can paste them directly from the connection menu

Good point. I guess there can be different ways this same type of feature is used.

If we do plan to keep this shared templates feature, what do you think about better aligning the name of this feature with what VS Code calls it: `snippets`. Beyond just the name, I would say we should align the behavior as well for consistency. My thought is they should be stored in a single `/etc/vscode/snippets.json` file that follows the same structure that VS Code uses locally.

https://code.visualstudio.com/docs/editing/userdefinedsnippets#_create-your-own-snippets

You can check this out by using the `Configure Snippets` command and using `New Global Snippets File...` where you see something like this:

image

I am thinking we store everything in this single file instead of having another directory to store the snippet. This keeps things consistent and makes it easy to copy snippets between the IFS and local. It would be great if we could also have this placeholders feature in VS Code also for these IFS snippets.

Again, what do you think about this? 😁

> If we want to add custom examples in db2 extension, we can open another pr

Yes I think we can still add support for this in the Db2 extension separately as I described. I believe I started this change a while ago, I will check later if I have the code for it still and will PR it.

buzzia2001:
Hi @SanjulaGanepola,

I made the requested change. The configuration is now contained in a single file and no longer in directories. I also replaced the various entries in the frontend from “templates” to “snippets.”

image

sebjulliand (changes requested):
This is a nice idea @buzzia2001 - but it needs some refining, IMHO.

We should need to stick to the VS Code UX as much as possible. For instance, VS Code allows to create global snippets, or snippets per language.

Image

That feature should allow the same thing: a global set of snippets (which is `/etc/vscode/snippets.json` at the moment) and snippets files per language id.

There is no need to have a new dedicated `File System` either. It can all be managed using what's already available.

Besides, it is possible to reuse the VS Code snippets schema for validation and completion when editing a snippet file.

Here are my initial suggestions:
- The `Snippets` menu from the status bar is a good entry point.
- Delete `schemas/snippets.json`
- Delete the `SharedSnippetFileSystemProvider`
- Store all the shared snippets (global and language specific) under a unique `/etc/vscode/snippets` IFS folder
- The folder should be created when needed, initially with access rights similar to `/etc/vscode/settings.json`. Sys admins can take care of fine tuning the rights afterwards.
- The snippet menu should allow to:
  - Create a global snippets file if not exists (`/etc/vscode/snippets/global.json`)
  - Open an existing language specific snippets file (`/etc/vscode/snippets/${language_id}.json`)
  - List the languages for which a snippets file doesn't exist yet
- Change the package.json `jsonValidation` contribution so it uses the VS Code snippets schema to validate shared snippets files:
```json
{
	"fileMatch": [
		"/etc/vscode/snippets/*.json"
	],
	"url": "vscode://schemas/snippets"
}
```
- Update the `SharedSnippet` type so it matches the VS Code snippets' schema
- When opening a snippet file, use this method to link the remote file to the snippets language id and keep it consistent with how VS Code do it:
```typescript
const document = await workspace.openTextDocument(getUriFromPath(<snippet file path>));
if(document.languageId !== "snippets"){
  await languages.setTextDocumentLanguage(document, 'snippets');
}      
await window.showTextDocument(document);
```
- Snippet files should be created before being opened (with UTF-8 encoding and rights similar to settings.json, as above)
- The `SharedSnippetCompletionItemProvider` must honor the `include` and `exclude` snippets field (taken from VS Code schema) to check if it matches the active document.
- The cache is a good idea - but storing and checking the files' modification date to prevent unneeded reloads should be used.
- Clear the remaining, unused code.

To summzarize, I think it would be best to stick to VS Code UX and keep it simple 😉 
Thanks!

buzzia2001:
ciao @sebjulliand...

As you requested, I have modified how snippets are managed. Specifically, a folder named /etc/vscode/snippets with 755 permissions is now created, containing a global file for all extensions and a dedicated file for each individual extension. 
In the global file, you can specify the scope (i.e., the extension) to which the snippet refers, allowing you to create snippets that are shared across the various selected languages... Of course, the snippet files are created in UTF-8. 

As for the cache, I now save the timestamp of the last modification, and if there are any differences, I re-read the file. 

However, I’ve kept the snippet creation wizard—which I consider quite essential—so that anyone can follow the wizard and create a snippet by choosing between code written on the spot or the editor window. 

I’ve also cleaned up dead and unused files following this refactor (which I hope will be the last one 😂)

package.json
```diff
@@ -62,6 +62,19 @@
...
+					"!/etc/vscode/snippets/global.json"
+				],
+				"url": "vscode://schemas/snippets"
+			},
+			{
+				"fileMatch": [
+					"/etc/vscode/snippets/global.json"
+				],
+				"url": "vscode://schemas/global-snippets"
 			}
```
sebjulliand, line 71: There is no such schema `vscode://schemas/global-snippets`, as far as I know.
Covering all the JSON files as snippets under the snippets folder should be enough.
```suggestion
			{
				"fileMatch": [
					"/etc/vscode/snippets/*.json"
				],
				"url": "vscode://schemas/snippets"
			}
```

src/commands/snippets.ts
```diff
@@ -0,0 +1,473 @@
...
+    languages.registerCompletionItemProvider(
+      [{ scheme: `file` }, { scheme: `untitled` }, { scheme: `member` }, { scheme: `streamfile` }],
+      new SharedSnippetCompletionItemProvider(instance)
+    ),
+
+    // Snippets files are edited like any other streamfile, and the editor knows nothing about our cache
+    workspace.onDidSaveTextDocument(document => {
+      const connection = instance.getConnection();
+      if (connection && document.uri.scheme === `streamfile` && SharedSnippets.isSnippetsFile(document.uri.path)) {
+        SharedSnippets.invalidate(connection);
```
sebjulliand: Since you know which file has been modified, try to invalidate only this one in the cache instead of clearing it all.
```suggestion
        SharedSnippets.invalidate(connection, document.uri.path);
```

src/commands/snippets.ts
```diff
@@ -0,0 +1,473 @@
...
+
+function toCompletionItems(snippet: NamedSharedSnippet) {
+  const body = toArray(snippet.body).join(`\n`);
+  const description = toArray(snippet.description).join(`\n`);
+
+  //Like VS Code does, a snippet with no prefix is offered under its name
+  const prefixes = toArray(snippet.prefix).filter(Boolean);
+  return (prefixes.length ? prefixes : [snippet.name]).map(prefix => {
+    const item = new CompletionItem(prefix, CompletionItemKind.Snippet);
+    item.detail = snippet.name;
```
sebjulliand: Just an idea: show that this snippet comes from an IBM i shared snippet. VS Code's user snippets are shown this way to differenciate them from others. 
```suggestion
    item.detail = l10n.t("{0} (IBM i shared)", snippet.name);
```

sebjulliand (changes requested):
Good job @buzzia2001 , we're getting there.
See my comments/suggestions in the code.

Also:
- There should be more feedback when a user is not allowed to create the snippet folder, file or if he's not allowed to write it. Right now, it fails silently. A notification about what went wrong and why would be nice.
- The snippets menu is confusing at first. When clicking on `Create new shared snippet... `, you get presented with the same items as before. Maybe only offer to Open a shared snippet file from there and take the `Add snippet` wizard elsewhere. 
- The wizard is nice, but goes a bit against [VS Code UX guidelines](https://code.visualstudio.com/api/ux-guidelines/quick-picks#multiple-steps), as there are many steps. That's understandable, but debatable as well - especially for creating a snippet out of the blue. The content assist does a fairly good job at guiding users when a snippet file is opened. However, the creation of a snippet out of the selection or the whole active editor is great. So maybe keep the wizard for these two options and let the user invoke them from the active aditor context menu ("Create shared IBM i snippet" submenu -> "from document" and "from selection", for example). And make the snippets status bar action open/create a shared snippet file.
- When a snippet gets added to an existing file, it's put at the top. So it may be put above the generated documentation comment. Try to insert it at the end instead.

buzzia2001:
Hi @sebjulliand , 
I've made the various changes you requested, so the cache is now invalidated only for the modified file, and the entries are now placed at the top. Errors related to accessing snippets are also displayed.

I modified the snippets menu to remove the wizard and allow only the viewing and creation of snippets. I also created a new option to create a snippet directly from a source.

image

package.json
```diff
@@ -2161,9 +2190,31 @@
...
+				{
+					"command": "code-for-ibmi.createSharedSnippetFromDocument",
+					"group": "1_create@2"
+				}
+			],
+			"editor/context": [
+				{
+					"submenu": "code-for-ibmi.createSharedSnippet",
+					"when": "code-for-ibmi:connected && !isInDiffEditor",
+					"group": "z_commands@1"
```
sebjulliand: Put the seubmenu in a dedicated group. It will be separated from the rest and it will be clearer to the user.
```suggestion
					"group": "1_codefori@1"
```

sebjulliand (changes requested):
@buzzia2001 apart from a small change to the new submenu, I'd suggest to remove the `scope` field and everything that relates to it. It overlaps with the `include` and `exclude` fields which are supported by standard snippets.

Finally, some small UX enhancements:
- If the snippet file doesn't exist yet and you don't have the rights to create it, you have to reach the end of the wizard to figure that out. It would be better to check if the directory and/or target file can be created beforehand. It does the check if the file exists though, so that check just need to cover more cases. Always fail as early as possible 😉 
- A nice to have change: let's says I'm creating a snippet out of the active editor or its selection. It could be nice to directly show the snippet file related to the editor's language (or offer to create it), since most of the time you'll want to create a snippet for that language in particular.
- Finally, with regards to the `UX guidelines`, the wizard should show the current and total steps in the process. The title may remain "New IBM i Shared Snippet (X/Y)" during the whole process, and the placeholder or prompt should show what kind of input is expected. (checkout the guidelines screenshots) 

Keep it up, you're almost there!

buzzia2001:
@sebjulliand to you :-D

sebjulliand (approved):
We're good to go now 😉 
Good job @buzzia2001 and thanks for your patience.
The UX is nice and this will prove to be useful feature, hopefully.

Now, just make a note to update the user documentation to describe this new feature.
Thanks!

PR #3425, opened 2026-08-21
Fix source members not saving when clearing all contents

SanjulaGanepola:
### Changes

Previously if you clear the entire contents of a source member and save, it would appear like the contents was removed. However, if you click the `Refresh File` action on the top right of the editor, you would see the contents come back. Part of the fix for this was in https://github.com/codefori/vscode-ibmi/pull/3156, but now this PR completes the fix.

### How to test this PR
1. Open a source member with some content.
2. Remove all the content.
3. Click the `Refresh File` button in the top right of the editor.
4. Observe that the content is not restored and the file is still empty.
5. Close and re-open the member and again verify the content is not restored.

### Checklist
* [x] have tested my change


buzzia2001 (commented):
Hi @SanjulaGanepola,
it works, thank you

PR #3432, opened 2026-08-29
Protected check on object

buzzia2001:
### Changes
This PR adds a check on “protected” filters for objects as well

### How to test this PR
Create a “protected” filter; when you click on an object, there should be no options to rename, move, or delete objects.
The actions to be performed must also be flagged to run in a protected environment.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/ui/actions.ts
+++ b/src/ui/actions.ts
@@ -47,7 +47,11 @@ export function uriToActionTarget(uri: vscode.Uri, workspaceFolder?: WorkspaceFo
     uri,
     extension: uri.path.substring(uri.path.lastIndexOf(`.`) + 1).toUpperCase(),
     fragment: uri.fragment.toUpperCase(),
-    protected: parseFSOptions(uri).readonly || ibmi?.getConfig().readOnlyMode || ibmi?.getContent().isProtectedPath(uri.path) || false,
+    protected: parseFSOptions(uri).readonly || ibmi?.getConfig().readOnlyMode
+      || (uri.scheme === `object`
+        ? ibmi?.getContent().isProtectedPath(Tools.parseQSysPath(uri.path).library)
+        : ibmi?.getContent().isProtectedPath(uri.path))
+      || false,
     workspaceFolder: workspaceFolder || vscode.workspace.getWorkspaceFolder(uri),
     executionOK: false,
     hasRun: false,
@@ -73,6 +77,17 @@ export async function runAction(instance: Instance, uris: vscode.Uri | vscode.Ur
     const content = connection.getContent();
 
     const targets = uris.map(uri => uriToActionTarget(uri, workspaceFolder, connection));
+
+    // To perform an action on Objects/members protected filter you need "runOnProtected" action
+    const protectedUris = new Set((browserItems || [])
+      .filter(item => (item as { isProtected?: () => boolean }).isProtected?.())
+      .map(item => item.resourceUri?.toString())
+      .filter((value): value is string => Boolean(value)));
+    targets.forEach(target => {
+      if (protectedUris.has(target.uri.toString())) {
+        target.protected = true;
+      }
+    });
 
     workspaceFolder = targets[0].workspaceFolder;
     if (!targets.every(target => target.workspaceFolder === workspaceFolder)) {
--- a/src/ui/actions.ts
+++ b/src/ui/actions.ts
@@ -43,15 +43,20 @@ export function registerActionTools(context: vscode.ExtensionContext) {
 }
 
 export function uriToActionTarget(uri: vscode.Uri, workspaceFolder?: WorkspaceFolder, ibmi?: IBMi): ActionTarget {
+  const extension = uri.path.substring(uri.path.lastIndexOf(`.`) + 1).toUpperCase();
+  let path = uri.path;
+  if (uri.scheme === `object`) {
+    const qsysPath = Tools.parseQSysPath(path);
+    path = (extension === `LIB` ? qsysPath.name.split(`.`)[0] : qsysPath.library);
+  }
   return {
     uri,
-    extension: uri.path.substring(uri.path.lastIndexOf(`.`) + 1).toUpperCase(),
+    extension,
     fragment: uri.fragment.toUpperCase(),
-    protected: parseFSOptions(uri).readonly || ibmi?.getConfig().readOnlyMode
-      || (uri.scheme === `object`
-        ? ibmi?.getContent().isProtectedPath(Tools.parseQSysPath(uri.path).library)
-        : ibmi?.getContent().isProtectedPath(uri.path))
-      || false,
+    protected: ibmi?.getConfig().readOnlyMode ||
+      parseFSOptions(uri).readonly ||
+      ibmi?.getContent().isProtectedPath(path) ||
+      false,
     workspaceFolder: workspaceFolder || vscode.workspace.getWorkspaceFolder(uri),
     executionOK: false,
     hasRun: false,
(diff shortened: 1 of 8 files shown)


src/webviews/filters/index.ts
```diff
@@ -63,7 +63,7 @@ export async function editFilter(filter?: ObjectFilters, copy = false) {
       .addInput(`memberText`, `Member text`, `Member text description filter.`, { default: filter.memberText || `*` })
       .addInput(`memberCreated`, `Member created from`, `Only list members created on or after this date. Leave blank to list members regardless of what their creation date is.`, { default: filter.memberCreated || ``, inputType: `date` })
       .addInput(`memberChanged`, `Member changed from`, `Only list members last changed on or after this date. Leave blank to list members regardless of what their last changed date is.`, { default: filter.memberChanged || ``, inputType: `date` })
-      .addCheckbox(`protected`, `Protected`, `Make this filter protected, preventing modifications and source members from being saved.`, filter.protected)
+      .addCheckbox(`protected`, `Protected`, `Make this filter protected: objects can't be renamed, moved, deleted or have their description changed, source members are read-only and can't be saved, and Actions run only when flagged to run on protected targets.`, filter.protected)
```
sebjulliand: Can be simplified a bit.
```suggestion
      .addCheckbox(`protected`, `Protected`, `Make this filter protected: objects and members can't be altered and Actions can only run if flagged to run on protected targets.`, filter.protected)
```

src/ui/actions.ts
```diff
@@ -47,7 +47,11 @@ export function uriToActionTarget(uri: vscode.Uri, workspaceFolder?: WorkspaceFo
     uri,
     extension: uri.path.substring(uri.path.lastIndexOf(`.`) + 1).toUpperCase(),
     fragment: uri.fragment.toUpperCase(),
-    protected: parseFSOptions(uri).readonly || ibmi?.getConfig().readOnlyMode || ibmi?.getContent().isProtectedPath(uri.path) || false,
+    protected: parseFSOptions(uri).readonly || ibmi?.getConfig().readOnlyMode
+      || (uri.scheme === `object`
+        ? ibmi?.getContent().isProtectedPath(Tools.parseQSysPath(uri.path).library)
+        : ibmi?.getContent().isProtectedPath(uri.path))
+      || false,
```
sebjulliand: The whole function should be like this to take into account library objects:

```ts
const extension = uri.path.substring(uri.path.lastIndexOf(`.`) + 1).toUpperCase();
  let path = uri.path;
  if (uri.scheme === `object`) {
    const qsysPath = Tools.parseQSysPath(path);
    path = (extension === 'LIB' ? qsysPath.name.split('.')[0] : qsysPath.library);
  }
  return {
    uri,
    extension,
    fragment: uri.fragment.toUpperCase(),
    protected: ibmi?.getConfig().readOnlyMode ||
      parseFSOptions(uri).readonly || 
      ibmi?.getContent().isProtectedPath(path) ||
      false,
    workspaceFolder: workspaceFolder || vscode.workspace.getWorkspaceFolder(uri),
    executionOK: false,
    hasRun: false,
    processed: false,
    output: []
  };
```

src/ui/views/objectBrowser.ts
```diff
@@ -1534,6 +1547,9 @@ Do you want to replace it?`, item.name), { modal: true }, skipAllLabel, overwrit
     }),
     vscode.commands.registerCommand(`code-for-ibmi.renameQSYS`, async (node?: (ObjectBrowserMemberItem | ObjectBrowserObjectItem)) => {
       node = getSelectedItems(node).at(0);
+      if (node instanceof ObjectBrowserItem && checkProtected(node)) {
```
sebjulliand: This is enough; all nodes from the object browser extend `ObjectBrowserItem` that defines `isProtected()`.
```suggestion
      if (node && checkProtected(node)) {
```

src/ui/views/objectBrowser.ts
```diff
@@ -1545,6 +1561,14 @@ Do you want to replace it?`, item.name), { modal: true }, skipAllLabel, overwrit
   );
 }
 
+function checkProtected(node: { isProtected?: () => boolean, toString(): string }): boolean {
+  if (node?.isProtected?.()) {
+    vscode.window.showWarningMessage(vscode.l10n.t(`{0} is in a protected filter or library and cannot be modified.`, node.toString()));
+    return true;
+  }
+  return false;
+}
```
sebjulliand: `node` is always defined and its type will always inherit `ObjectBrowserItem` that always defines `isProtected()`.
```suggestion
function checkProtected(node: ObjectBrowserItem): boolean {
  if (node.isProtected()) {
    vscode.window.showWarningMessage(vscode.l10n.t(`{0} is in a protected filter or library and cannot be modified.`, node.toString()));
    return true;
  }
  return false;
}
```

sebjulliand (changes requested):
@buzzia2001 overall it's OK, but there are a couple of things to change.

And I guess think you can push the logic a bit further:
- the `code-for-ibmi.createSourceFile` command should also not be allowed to run on protected targets (filters and library objects)
- the `isProtected()` abstract method from `ObjectBrowserItem` should be pulled up into its super class `ObjectItem`. Then, `DebugItem` and `EnvironmentItem` will define it to always return `false`, and `IFSItem` can benefit from it and return an actual value, based on the actual protection status of the IFS item. Thi seems to be a cleaner approach and avoid type checks like `{ isProtected()?: boolean }`.

buzzia2001:
@sebjulliand to you

src/ui/views/ifsBrowser.ts
```diff
@@ -136,6 +144,10 @@ class IFSItem extends BrowserItem implements WithPath {
     this.tooltip = VscodeTools.ifsFileToToolTip(this.path, file);
   }
 
+  isProtected(): boolean {
+    return isProtected(this.path);
```
sebjulliand: Since the path protection applies to directories, a file needs to check its parent node protection state.
```suggestion
    return this.file.type === "directory" ? isProtected(this.path) : this.parent?.isProtected() || false;
```

sebjulliand (changes requested):
We're almost there @buzzia2001!

Now that we have the protected state accross the IFS browser, actions that cannot run on protected IFS files should be hidden.

These four specifically:

Image

Keep up the good work! 😄

buzzia2001:
@sebjulliand I hope we've reached the end of this PR hahaha

sebjulliand (changes requested):
@buzzia2001 it's all good!

Just one thing I noticed, not directly related to this PR but worth fixing while we're at it: the IFS paths `isProtected` check is case sensitive, which is only valid for /QOpensys paths. The rest of the IFS FS is case insensitive.

So here is an enhancement for the `isProtectedPath()` function in `IBMIContent.ts` on line 1058: 
```ts
  isProtectedPath(path: string) {
    if (path[0] === '/') { //IFS path
      //QOpensys FS is case sensistive
      const flag = /^\/QOpenSys/i.test(path) ? '' : 'i';
      return this.config.protectedPaths.some(p => new RegExp(escapeStringRegexp(p), flag).test(path));
    }
    else { //QSYS path
      const qsysObject = Tools.parseQSysPath(path);
      return this.config.protectedPaths.includes(this.ibmi.upperCaseName(qsysObject.library));
    }
  }
```

With this, we make the check case insensitive unless the protected path is in QOpenSys.

buzzia2001:
@sebjulliand to you

sebjulliand (approved):
Good job @buzzia2001 , it's fully working!

PR #3438, opened 2026-09-06
Fixed Mapepire component SQL jobs tracking

sebjulliand:
### Changes

This PR fixes the tracking of SQL jobs opened by the Mapepire component:
- All jobs are now correctly added to the job Map; only SSH Jobs were added, jobs opened in Websocket mode were not.
- Closing a job will remove it from the job Map; job's `close()` has been overriden to do so

### How to test this PR

1. Run the extension in debug mode
2. Put a breakpoint in `src/api/components/mapepire/index.ts` on line `168`
3. Open multiple jobs with the Db2 extension and then close them
4. Check when the breakpoint is reached that the jobs Map doesn't contain that job
5. Check the job is correctly closed on the IBM i

### Checklist

* [x] have tested my change

### DIFF
--- a/src/api/components/mapepire/index.ts
+++ b/src/api/components/mapepire/index.ts
@@ -20,7 +20,7 @@ export class Mapepire implements IBMiComponent {
   private installPath = "";
   private readonly version: SemanticVersion;
 
-  readonly jobs: Map<string, SQLJob> = new Map;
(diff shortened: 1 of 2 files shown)


buzzia2001 (approved):
Work as expected, thank you @sebjulliand!

PR #3442, opened 2026-09-10
Add PGMINFO option to command strings in package.jso

anandsha20:
### Changes
Why Needed:
When adding a service program or RPG program to an IWS  server, a PCML file is required. Currently, generating the PCML file is not included by default in the compile commands, which makes it harder to add newly compiled or updated programs to IWS.

Additonal Info:
According to IBM documentation, specifying PGMINFO(*PCML *MODULE) causes the compiler to generate an XML (PCML) description of the program's parameter interface and embed it inside the compiled module object (*MODULE). When that module is later bound into a *PGM or *SRVPGM, the PCML travels with it into that object too.

This change will not cause any issue with non-IWS programs
### Checklist

* [x] have tested my change

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -90,3 +90,5 @@ Thanks so much to everyone [who has contributed](https://github.com/codefori/vsc
 * [@Frank-Hildebrandt](https://github.com/Frank-Hildebrandt)
 * [@christianlarsen](https://github.com/christianlarsen)
 * [@Balrocj](https://github.com/Balrocj)
+* [@anandsha20](https://github.com/anandsha20)
+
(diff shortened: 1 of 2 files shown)


SanjulaGanepola:
This is not a parameter which I have used before, but given the description this sounds like it is applicable for specific use cases as you mention for IWS. Thus, this to me looks like it should be configured by the user if needed in their actions rather than in the defaults shipped with the extension.

sebjulliand:
Sorry @anandsha20 but @SanjulaGanepola is right. Actions are meant to be customized by the users to fit their needs.
PGMINFO is only used when needing the PCML to deploy in IWS, which is not the default use case for a majority of users.
This is typically something users will add if needed in their actions. It doesn't belong in the actions shipped by default with the extension.

PR #3445, opened 2026-09-11
Fix `writeFile` file system API when used on members by other extensions

SanjulaGanepola:
### Changes

In one of our other extensions, we make use of VS Code's file system APIs to read / write to members using Code4i's custom file system provider for QSYS:

```ts
await workspace.fs.writeFile(memberUri, Buffer.from(content, 'utf8'));
```

Previously, when the member does not exist and this API is used, it would fail in 2 scenarios:
1. **Source Dates Enabled**: The member would fail to be created.
2. **Source Dates Disabled**: The member would be created successfully with the content, but it would not have the source type.

This PR now fixes both scenarios.

### How to test this PR

This is not exactly testable using any functionality in Code4i. You can reproduce the original issue and verify the fix using an extension that uses the `writeFile` API on a `uri` with the `member` scheme. You can also use the unit tests to validate that nothing has been broken and the new test properly sets the source type now.

### Checklist
* [x] have tested my change

### DIFF
--- a/src/api/IBMiContent.ts
+++ b/src/api/IBMiContent.ts
@@ -301,7 +301,7 @@ export default class IBMiContent {
    * @param member
    * @param content
    */
-  async uploadMemberContent(library: string, sourceFile: string, member: string, content: string | Uint8Array): Promise<boolean> {
+  async uploadMemberContent(library: string, sourceFile: string, member: string, content: string | Uint8Array, extension?: string): Promise<boolean> {
     library = this.ibmi.upperCaseName(library);
     sourceFile = this.ibmi.upperCaseName(sourceFile);
     member = this.ibmi.upperCaseName(member);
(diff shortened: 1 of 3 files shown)


sebjulliand (approved):
Tested by invoking `writeFile` with various settings. It works fine!
One very minor issue: the browser gets refreshed if the member gets created and the source date support is enabled; but when source date support if off, it doesn't refresh (because of the way the member is created on the fly if it doesn't exist).
It's very minor so I'll approve the PR and let you decide if you want to fix that before merging @SanjulaGanepola .

SanjulaGanepola:
Good catch. However, since the `IBMiContent#uploadMemberContent` is part of the API (where we want to avoid `vscode` dependencies), fixing this would require further changes to make the `uploadMemberContent` API more complicated. For now this is an acceptable limitation so I'll leave it as is. Thanks for the review

Issue #2409, opened 2024-12-10
Member Lock


pabloto:
Hi,

I need a suggestion, some colleagues still using SEU 😭, I want to give their VSCode. 

We have a validity check program in STRSEU command. This program simple access to a table and write a row with the developer who are editing that source. If another developer  try to open that source it receive a message that prevent open it.

Could I simulate this with vscode? Is there the possibility to simulate a validity check program with a codeforI api?

Many thanks

<hr />

⚠️ **REMOVE THIS LINE AND ANY SENSITIVE INFORMATION BELOW!** ⚠️

|Context|Version|
|-|-|
|Code for IBM i version|2.14.5|
|Visual Studio Code version|1.95.3|
|Operating System|darwin_arm64|

<details>
<summary>Active extensions</summary>

```
COBOL (cobol): 24.11.28
Code Spell Checker (code-spell-checker): 4.0.21
Code for IBM i Walkthroughs (vscode-ibmi-walkthroughs): 0.5.0
Configuration Editing (configuration-editing): 1.0.0
Continue - Codestral, Claude, and more (continue): 0.8.61
Db2 for IBM i (vscode-db2i): 1.7.0
Dev Containers (remote-containers): 0.388.0
ESLint (vscode-eslint): 3.0.10
Emmet (emmet): 1.0.0
Error Lens (errorlens): 3.20.0
Extension Authoring (extension-editing): 1.0.0
Git (git): 1.0.0
Git Base (git-base): 1.0.0
Git Graph (git-graph): 1.30.0
GitHub (github): 0.0.1
GitHub Authentication (github-authentication): 0.0.2
GitHub Pull Requests (vscode-pull-request-github): 0.100.3
GitLab Workflow (gitlab-workflow): 5.21.0
GitLens — Git supercharged (gitlens): 16.0.5
IBM i Debug (ibmidebug): 2.0.1
IBM i Project Explorer (vscode-ibmi-projectexplorer): 2.12.0
JSON Language Features (json-language-features): 1.0.0
Live Share (vsliveshare): 1.0.5941
Markdown All in One (markdown-all-in-one): 3.6.2
Markdown Math (markdown-math): 1.0.0
Merge Conflict (merge-conflict): 1.0.0
Microsoft Account (microsoft-authentication): 0.0.1
NPM support for VS Code (npm): 1.0.1
Node Debug Auto-attach (debug-auto-launch): 1.0.0
Postman (postman-for-vscode): 1.5.0
[...]

Issue #1688, opened 2023-12-02
Locking the source on the IBMi while editing in Code for IBM i


christianlarsen:
I am trying to introduce "Code for IBMi" in my company, and one of the problems that I am reporting is that, while someone is using a source code with another IDE (RDI, SEU...), other people can open and modify the same source with VSCode. And that is not good..., because it can cause errors, code deletion, etc...

I don't know if there is a way to do this, but it would be great to be able to lock the source when editing it using Code for IBMi, so that no one could touch it in the meantime...

Thank you!

worksofliam:
We have this discussion every once in a while, but it is likely impossible for us to implement locking due to the way jobs are managed and the way source is opened.

Cc @codefori/core any creative ideas?

christianlarsen:
Well, when you are using RDi, it is copying the source you are editing into your own computer... and it is locking the source, so nobody can open it for editing...  and when you are saving the source, it is copying it from your computer to the IBMi ... 

It can be made with an ALCOBJ for the file QRPGLESRC (for example) and the member....

The problem is when you are trying to save your changes, it has to deallocate the file, save it, and allocate it again... so between those operations, someone can try to open it, and allocate it ... so you will get an error....
But I know that in my team, some people has been having problems with this in RDi, too ... and I think that is because of lost connections... and so on....

I dont know if this is an idea for you, or a problem. :-D

Wright4i:
> We have this discussion every once in a while, but it is likely impossible for us to implement locking due to the way jobs are managed and the way source is opened.
> 
> Cc @codefori/core any creative ideas?

Sure here's a crazy idea: Would it be possible to detect the source has changed and auto open a diff? 

I seriously doubt we could get the magical git 3way merge diff, but even just being able to communicate the changes to the user when they save and offering to commit(replace) or rename would probably be a big QOL upgrade.

IMHO I **don't** think it's worth trying to get locking working in the same way RDi does as the headache of locks lingering unintentionally due to network drops/closed vscode sessions, etc. would introduce even more problems.

PR #3446, opened 2026-09-13
feat: add IBM i source member locking

Balrocj:
## Related discussions

- #1688 and #2409: original member-locking requests.
- https://github.com/orgs/codefori/discussions/2468
- https://github.com/orgs/codefori/discussions/3132

## Summary

Adds new `codefori.memberLocking` setting to prevent concurrent edits to IBM i source members in QSYS source physical files.

When enabled, Code for IBM i acquires an IBM i `*EXCLRD` member allocation through the persistent Mapepire SQL job before opening a QSYS member for editing. IFS streamfiles are not affected. If another job already holds the member, the user can open it read-only instead.

https://github.com/user-attachments/assets/450e3d10-0550-48de-a426-f078673bb69a

## Implementation

- Adds a `MemberLockManager` to acquire, track, restore, and release member allocations.
- Acquires the lock before opening an editable member.
- Releases locks when the document closes or the connection disconnects.
- Restores active member locks after reconnecting.
- Uses the existing persistent Mapepire SQL job; no new dedicated job or public API is introduced.
- Serializes `IBMi.runSQL()` calls per connection so job-scoped commands and SQL work cannot overlap.
- Runs Source Date Support save work through the same SQL job when a member lock is active.
- Adds remote schema validation for `codefori.memberLocking`.

## User Experience

When an IBM i job already locks a member, Code for IBM i shows a modal message and offers **Open read only**.

## Validation

- `npm run test:unit` passes.
- `npm run webpack` passes.
- Focused IBM i integration test on PUB400 passes:
  - conflicting allocation returns `CPF1002`
  - independent member locks in the same source file can be released independently
  - cleanup releases allocations

## Scope

This setting applies only to QSYS source members (`member:` URIs). IFS streamfiles continue to use the existing writable-file permission checks and are not IBM i member-locked.


sebjulliand:
@SanjulaGanepola this seems similar to what's done in #3409 . You and @Balrocj should compare notes and come up with a unified solution. What do you guys think?

@Balrocj be careful: there are some commits coming from a previous PR (the RPGLE folding thing). These commits must be removed from here. Thanks.

Balrocj:
Thanks  @sebjulliand, I have updated the branch so the unrelated RPGLE folding commits have been removed. The PR now contains only the member-locking changes.

I will also compare notes with @SanjulaGanepola on #3409 to align on a unified solution.

sebjulliand:
Great! Thank you @Balrocj

SanjulaGanepola:
@Balrocj Thanks for also working on this feature. I will take a look at this PR tomorrow and compare with what I had started. Will comment back here soon.

buzzia2001 (changes requested):
So, I started reviewing this PR... 
To be honest, I don't like that this parameter is under the “mapepire” tab; I'd move it under ‘FEATURES’ or “SOURCE CODE.” The entry should also be included in the server's JSON schema.
Another issue to fix: if I’m connected to Mapepire in server mode and I abruptly disconnect by closing the client, the lock remains… this can cause problems… 
Another point to review is reconnection: if, in the meantime, someone locks the object (which can be simulated with STRSEU option 2) and I reopen the client and reconnect, there’s no check on the object’s lock status. So, if you try to edit it, when you save, it gets stuck until the timeout.
I've run some tests and I've also noticed that if I'm disconnected in server mode, it maintains the locks. In my case, this is the configuration used by a profile with iASP enabled:

image

But @SanjulaGanepola I think this is an issue with server mode, since when I'm not using the server, I close the connection and the lock is released instantly.

SanjulaGanepola:
@Balrocj I think the implementation here is slightly more complicated than required. Also, this PR includes several other changes which definitely need to be broken up into other PRs (listed below). We try to keep PRs focused on a single change so it is easier to review and for tracking over time.

1. For `runSQL`, I see that a queue was added. Curious as to what exactly wasn't working which resulted in having to add this?
2. There appears to also be a new style of tests introduced which makes use of mocking of the VS Code APIs. I actually like this idea, but I think we definitely need to do this in a different PR. If we do want to go ahead with this style of tests, I think we should consider first migrating the [tests in our old test framework](https://github.com/codefori/vscode-ibmi/tree/master/src/testing).
3. Other changes to certain APIs in `extendedContent`, etc.

Have you tried the implementation I proposed here: https://github.com/codefori/vscode-ibmi/pull/3409? There are some things I noticed in your PR which I can migrate over which were good ideas:
- Error handling to open in read-only mode

Shall I proceed with making the necessary changes mentioned above in PR https://github.com/codefori/vscode-ibmi/pull/3409 for member locking support and then you can help add the other changes in subsequent PRs?

In regards to the type of lock being used, I want to understand better the reason for using a `*EXCLRD` lock? In the original PR I opened, I had used a `*SHRUPD` with the idea that anyone editing the member needs to acquire this type of lock. When testing in RDi, I noticed this is also the type of lock which they use:

image

I know @buzzia2001 this is something you also mentioned and suggested to use `*EXCLRD` so curious to hear your thoughts on this as well.

> I think this is an issue with server mode, since when I'm not using the server, I close the connection and the lock is released instantly.

@buzzia2001 Yes, sounds like a Mapepire bug. Will investigate.

SanjulaGanepola:
@Balrocj Right as I posted the comment above, I just noticed your comment in my PR going over the [main differences](https://github.com/codefori/vscode-ibmi/pull/3409#issuecomment-5654420901) 😅 (thanks for taking a look). Couple things based on that:

- Surfaces other IBM i or Mapepire lock-acquisition errors with their diagnostic details instead of treating them as lock conflicts.

> Can you clarify this one?

- Tracks locks independently per member, including multiple members in one source physical file.

> Not sure I quite follow this one. Isn't that tracked [here](https://github.com/codefori/vscode-ibmi/pull/3409/changes#diff-89a0df2debbf3914722387c7fd669424da380c02bf3912f74120baec39a4e6b0R12)

- Releases locks on document close and disconnect, and restores active locks after reconnecting.

> Disconnect is handled [here](https://github.com/codefori/vscode-ibmi/pull/3409/changes#diff-b86b2b7a0fff2110ee6f6aaec94ecd51897b1ce4106a285dae902b07e89e6668R76). Reconnect is handled [here](https://github.com/codefori/vscode-ibmi/pull/3409/changes#diff-b86b2b7a0fff2110ee6f6aaec94ecd51897b1ce4106a285dae902b07e89e6668R67).

IFS streamfiles are intentionally out of scope

> Yes, I think we can add this separately afterwards if needed as it requires more investigation.

Balrocj:
Thanks everyone for the detailed review.

I agree that member locking should be consolidated into #3409 rather than maintained as two parallel implementations.

@SanjulaGanepola, please proceed with the member-locking implementation in #3409. I am happy to help test it and contribute smaller follow-up changes separately.

The main improvement from this PR that I suggest carrying over is the conflict UX: when the allocation fails with `CPF1002`, shoud offer the user an **Open read only** option. Other allocation failures should continue to show their diagnostic details rather than being treated as lock conflicts.

I also agree that the per-connection `runSQL` queue should be split out. I added it because the lock lifecycle and Source Date Support can both issue work through the persistent SQL job, and job-scoped operations must not overlap.
However, this is a general `IBMi.runSQL` behavior change and deserves its own focused PR with a clear reproduction case and tests.

@buzzia2001, I agree that the setting should not live under Mapepire. When the locking feature is integrated, it should be placed under a more appropriate section such as Source Code or Features, and included in the server JSON schema.

Regarding Mapepire Server mode: the retained allocation after an abrupt client close appears related to the server/job lifecycle, since it is released immediately in non-server mode. This needs investigation before considering server mode fully supported.

The reconnect path should also revalidate the allocation before allowing editing or saving, so a lock taken by another job while disconnected is detected early instead of causing a save timeout.

Finally, on the lock type: I used `*EXCLRD` to make the allocation conflict visible immediately and prevent concurrent access while editing. Since RDi uses `*SHRUPD`, I agree we should validate the expected interoperability with SEU, RDi, and other IBM i tools, then choose one lock type intentionally before the implementation is merged.

I will close this PR to avoid maintaining two parallel solutions. The conflict UX and the observations documented here remain available as reference. I am happy to help test #3409 and open focused follow-up PRs for independent improvements once the core approach is merged.

PR #3451, opened 2026-09-15
Add support to reset current library to default

SanjulaGanepola:
### Changes
A user reported that Code4i does not support resetting the current library to what is set on their user profile so this PR addresses that. Resetting was possible on the user libraries by entering `*reset`, but we didn't have anything to reset the current library.

image

I also moved the logic from https://github.com/codefori/vscode-ibmi/pull/3411 where we tracked the system libraries from `IBMiContent` to the `IBMi` class.

### How to test this PR
1. Connect with a user profile that has `*CRTDFT` as their current library.
2. Observe that you have the `Reset to Default` option available if you set the current library in Code4i.

Repeat the above with a user profile that does have a current library set on their profile.

### Checklist

* [x] have tested my change


sebjulliand (approved):
Works like a charm. Good one @SanjulaGanepola

PR #3454, opened 2026-09-17
Enforce protection GOTO

buzzia2001:
### Changes
This PR fixes a bug related to the use of the GOTO function. If a file is opened directly via GOTO without ever having been opened before with a filter, it is always opened in edit mode, regardless of whether it is in a protected path or not. Below is a video showing how to reproduce the issue:

https://github.com/user-attachments/assets/9838033f-db08-4156-b330-80d01d3d2346

### How to test this PR
Use the `goto` function to open a member in a protected library. A padlock icon will now appear next to the member if it is protected, and opening it will make the file read-only.

image

### Checklist
* [x] have tested my change


sebjulliand (changes requested):
>If a file is opened directly via GOTO without ever having been opened before with a filter, it is always opened in edit mode, regardless of whether it is in a protected path or not. 

I tried exactly that and it works as expected, without the PR.
I connect to PUB400 and use `Go To` to open a member in my library which is in the protected paths: it opens as read-only.

However, de-protecting it from the settings and opening it again using `Go To` still opens it as read-only, because of that block of code:
image

Just removing that fixes the issue you mentioned. What's missing in that test is ensuring that the opened document is actually visible (there is a difference in VS Code between an opened document and a visible document).

The additional test in the QSysFS doesn't seem necessary in this case, as the read only flag from the URI should prevail.

buzzia2001:
Hi @sebjulliand ,
Sorry, but I don't agree... I changed the controls as you requested (I'm attaching the files below), but I'm still getting the same error I had before applying the PR... 
Did you happen to follow all the steps I showed in the video? Did you establish a new connection, clear the recent items from GOTO, and open a source in a protected path?

[open.ts](https://github.com/user-attachments/files/32419061/open.ts)
[QSysFs.ts](https://github.com/user-attachments/files/32419062/QSysFs.ts)

PR #3455, opened 2026-09-18
Fix deployment startup subscription push

bobcozzi:
This PR fixes the startup crash caused by an unbound `push` callback in the deployment initializer.

Root cause
- `src/filesystems/local/deployment.ts` called `workspaceWatcher().then(context.subscriptions.push);`
- When the promise resolved, `push` was invoked without the `Subscription` array as its receiver.
- That produced the startup failure observed during activation: `TypeError: Cannot convert undefined or null to object` at `push (<anonymous>)`.

Fix
- Replace the unbound callback with an explicit arrow function: `workspaceWatcher().then(watcher => context.subscriptions.push(watcher));`
- This preserves the `this` binding and registers the watcher correctly.

Verification
- The branch was created from `upstream/master` so the PR is isolated to this one fix.
- `deployment.ts` validates cleanly.
- Earlier `package.json` and `.vscode/launch.json` edits used during investigation were intentionally left out of this branch.

Scope
- One file
- One behavioral change


src/filesystems/local/deployment.ts
```diff
@@ -39,7 +39,7 @@ export namespace Deployment {
       vscode.commands.registerCommand(`code-for-ibmi.launchActionsSetup`, DeployTools.launchActionsSetup),
       vscode.commands.registerCommand(`code-for-ibmi.launchDeploy`, DeployTools.launchDeploy),
       vscode.commands.registerCommand(`code-for-ibmi.deploySelected`, async (item?: Uri, items?: Uri[]) => {
-        if(!(item instanceof Uri)) {
+        if (!(item instanceof Uri)) {
```
bobcozzi, line 42: Sorry about the format change being in here. The F'ing linter kept changing it on me.

sebjulliand (approved):
Good catch @bobcozzi !

PR #3458, opened 2026-09-21
Refactor connection setup in vitest and fix connection cache

SanjulaGanepola:
### Changes

This PR refactors the connection setup code used in vitest so it is more structured and includes more setup logs. It also addresses an issue where the cache in `.storage.json` and `.config.json` was being re-used even though the connection in the `.env` file was different than what was cached.

### How to test this PR
1. Run `npm run test`

### Checklist
* [x] have tested my change


SanjulaGanepola:
I am still in the process of finding a machine we can use for our CI. The machine I had switched to originally is too slow.

PR #3461, opened 2026-09-23
Quality of life improvements after connect process and during deploy/compile

SanjulaGanepola:
### Changes

- [x] Removed the below 2 info notifications after connecting successfully
    image
- [x] Updated the `No suitable actions found for file` notification to have buttons to `Setup Workspace Actions` (this is shown for local only) and `Manage All Actions` (which navigates to the `Actions` node in the `Environment` view)
    image
- [x] Fixed a bug where if you do a `Launch Workspace Action Setup` in a workspace and then delete the `actions.json` file, several child nodes under the `Action` node would be duplicated (I believe it is a concurrency issue where multiple refreshes were invoked)
- [x] Improve the `Set Deployment Directory` action to have the current deploy directory or the suggested one already prefilled
    image

### How to test this PR
Simply just run the extension and verify each of the above actions

### Checklist
* [x] have tested my change


buzzia2001 (approved):
Works as expected

PR #3462, opened 2026-09-23
Modernize IBM i source member save with Mapepire SQL and add archive support

bobcozzi:
### Main Feature: SQL-only source member save (Mapepire)
- New Mapepire-based save member implementation, conditionally enabled and off by default
- Replace legacy CL/RUNSQLSTM host-save path with pure SQL-based staging flow
- Stage member data in QTEMP tables with 500-row batches to handle large members without size limits
- Use SQL alias to target the source member and rebuild from staged data
- Backup current member before write and restore on validation failure
- Support decimal sequence numbering (0.01, 0.02, ...) for large members (>= 10k lines)
- Controlled by the new  setting, shipped disabled (false) by default so the original save routine continues to run unless explicitly enabled

### Secondary Feature: Local source save archive
- Archive previous member content to local filesystem before each source-date save
- Enable browsing and restoring archived snapshots from a quick-pick with partial-name matching
- Archive search supports any part of library/file/member without requiring full 3-part path
- Archive is independent from the save method (works with both old and new paths)
- Configuration defaults: archive enabled, path ~/.code-for-ibmi/source-archive, retention 25 snapshots per member

### Mapepire SQL save flow
1. Alias the target source member to QTEMP.C4I_MBR_ALIAS
2. Create staging table QTEMP.C4I_SRCSAVE from member file structure
3. Insert new source data in batches of 500 rows to the staging table
4. Backup current member to QTEMP.C4I_BACKUP before replacing
5. Delete and reload the member from staged data using SQL INSERT
6. Verify row count matches staged rows; restore backup if mismatch
7. Drop temporary alias and tables in cleanup phase

### Archive snapshot structure
- Preserves member text and source dates, not sequence numbers
- Snapshot file is JSON stored in local archive directory
- Restore regenerates sequence numbers from line order on reinsertion
- Retention limit enforced per member (configurable, default 25)

## Testing
- npm run test:unit: 1 suite, 7 tests passed

## Notes
- Fully SQL-based, no CL scripts or temp file I/O
- Backward compatible: archive works with any save path
- Large members (100k+ lines) handled via batching and decimal sequence mode
- Safe rollout: new save path is off by default and only used when enabled


sebjulliand:
@bobcozzi you'll need to clean up that PR before it gets reviewed.
- Some unrelated changes, already merged, are in there (i.e. workspace watcher)
- MD files presumably used with/by your AI agent
- Test build zip files

The `DebugConfiguration` was also modified for no reason (I guess the AI ran loose and applied some rules along the way...).

Please clean it up so only what's related to the PR can be reviewed. Thanks!

bobcozzi:
✅ PR cleaned up - all unrelated files and changes have been removed:

- ✅ Removed AI markdown files (COPILOT_HANDOFF.md, cnnTree.md, mapepire_saveFile.md, plan-crtdftPrFeedback.md)
- ✅ Removed build zip files (*.vsix.zip)
- ✅ Reverted unrelated DebugConfiguration.ts changes
- ✅ Reverted unrelated l10n changes  
- ✅ Reverted workspace watcher fix (already merged)
- ✅ Reverted environment view changes (separate feature)

Branch now contains only the two main features:
1. **Mapepire SQL-based source member save** (opt-in, default off)
2. **Local source save archive** with restore capability

✅ All unit tests pass (7/7)

bobcozzi:
Okay, left it in draft mode, but it is really ready to be test. Just don't have time today to do more testing... will get to it later this weekend, likely.

